PR #2, opened 2018-01-05
Various API changes

fancycode:
- All functions that can potentially fail now return a `heif_error`.
- Decoding plugins can also return errors in case something fails.
- The libde265 decoder only uses the public API.

### DIFF
--- a/examples/convert.cc
+++ b/examples/convert.cc
@@ -147,8 +147,7 @@ int main(int argc, char** argv)
     }
 
     struct heif_image* image;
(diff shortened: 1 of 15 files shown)


fancycode:
I'm thinking we should even change all functions to return a `heif_error`, even if they just return a number like `heif_context_get_number_of_images`. This allows returning an error if the passed pointers are not valid.

fancycode:
Manually rebase and merged by @farindk

PR #7, opened 2018-03-29
Use generic API to set encoder options.

fancycode:
This is what we were talking about - still work in progress, but should go in the right direction.


fancycode:
Implemented differently in 1.1.0 - this PR is no longer required.

PR #13, opened 2018-04-27
Mark operator() on comparator function as const

alex:
Without this, on a sufficiently new clang, with enough warnings turned on, you get a compilation error:

https://gist.github.com/reaperhulk/17f35e9410177970677bb9a30958f884


fancycode:
Thanks, I squashed before landing your changes.

PR #19, opened 2018-05-13
Fix VPATH (out-of-tree) builds.

Jehan:
I always do VPATH builds, and libheif was not building correctly because some files were in the src dir and other in the build dir, and the current automake code was not properly looking in both.

This fixes it.


farindk:
Thanks, Jehan !

PR #24, opened 2018-05-24
Add generic reader API.

fancycode:
This can be used to load images from arbitrary sources by providing the necessary callbacks for reading, seeking, etc.

### DIFF
--- a/src/heif.h
+++ b/src/heif.h
@@ -273,6 +273,36 @@ void heif_context_free(struct heif_context*);
 
 struct heif_reading_options;
 
+enum heif_reader_offset {
+  heif_seek_start,
+  heif_seek_current,
+  heif_seek_end
+};
+
+struct heif_reader {
+  // API version supported by this reader
+  int reader_api_version;
+
+  // --- version 1 functions ---
+  uint64_t (*get_length)(struct heif_context* ctx, void* userdata);
+  uint64_t (*get_position)(struct heif_context* ctx, void* userdata);
+  int (*read)(struct heif_context* ctx,
+              void* data,
+              size_t size,
+              void* userdata);
+  int (*seek)(struct heif_context* ctx,
+              int64_t position,
+              enum heif_reader_offset offset,
+              void* userdata);
+};
+
+// The heif_reading_options should currently be set to NULL.
+LIBHEIF_API
+struct heif_error heif_context_read(struct heif_context* ctx,
+                                    struct heif_reader* reader,
+                                    const struct heif_reading_options* options,
+                                    void* userdata);
+
 // Read a HEIF file from a named disk file.
 // The heif_reading_options should currently be set to NULL.
 LIBHEIF_API
--- a/src/heif.h
+++ b/src/heif.h
@@ -284,8 +284,8 @@ struct heif_reader {
   int reader_api_version;
 
   // --- version 1 functions ---
-  uint64_t (*get_length)(struct heif_context* ctx, void* userdata);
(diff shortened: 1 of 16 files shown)


src/heif.h
```diff
@@ -273,6 +273,36 @@ void heif_context_free(struct heif_context*);
 
 struct heif_reading_options;
 
+enum heif_reader_offset {
+  heif_seek_start,
+  heif_seek_current,
+  heif_seek_end
```
farindk: We probably can remove heif_seek_end, because I do not know any case where we would need this.
Omitting this could make the reader easier to write.

src/heif.h
```diff
@@ -273,6 +273,36 @@ void heif_context_free(struct heif_context*);
...
+};
+
+struct heif_reader {
+  // API version supported by this reader
+  int reader_api_version;
+
+  // --- version 1 functions ---
+  uint64_t (*get_length)(struct heif_context* ctx, void* userdata);
```
farindk: There might be inputs with unknown length or even inputs that grow while reading (e.g. displaying of partial images while downloading).
Do we really need the length explicitly? I think the current code does not use this.

src/heif.h
```diff
@@ -273,6 +273,36 @@ void heif_context_free(struct heif_context*);
...
+  uint64_t (*get_position)(struct heif_context* ctx, void* userdata);
```
farindk: As seek() takes an int64, we probably also should return an int64 here.

src/heif_context.h
```diff
@@ -202,6 +218,20 @@ namespace heif {
 
     void write(StreamWriter& writer);
 
+    class InternalReader {
```
farindk: How about "ReaderInterface" instead of "InternalReader" and "Reader_Memory" / "Reader_File" instead of InternalMemoryReader?

src/heif.h
fancycode: This is used in `BitstreamRange::skip_to_end_of_file()`.

src/heif.h
fancycode: The `BitStreamRange` uses it to check if data can be read.

src/heif.h
fancycode: As the position of a file can never be negative, I would rather keep the `uint64_t`. `seek()` allows negative numbers to seek backwards.

src/heif_context.h
fancycode: Renamed the classes.

src/heif.h
farindk: Ok, I see. But skip_to_end_of_file() is only used if a box cannot be parsed because its type is unknown and the box length is "until end of file". Hence, we will find another way to handle this.

src/heif.h
farindk: Currently, it is checking the EOF flags.

src/heif.h
farindk: I'm afraid of all the compiler warnings that will show up: "passing unsigned value to signed parameter may lose data". 63 bit for file length should be enough.

src/heif.h
fancycode: It also checks the remaining bytes:
https://github.com/strukturag/libheif/blob/a53a44e779e57b578324d465cf51dc4b40b0275a/src/bitstream.h#L66

src/heif.h
fancycode: Ok, changed.

farindk:
Thanks. Integrated in v1.3.0 with changes.

PR #25, opened 2018-05-24
Win32 build

Jehan:
For GIMP, we need libheif to build for Windows. Here are the commits which resulted after failing to build (now it builds fine).

### DIFF
--- a/src/file-fuzzer.cc
+++ b/src/file-fuzzer.cc
@@ -19,7 +19,7 @@
  */
 
 #include <assert.h>
-#include <alloca.h>
+#include <stdlib.h>
 
 #include "heif.h"
 
@@ -77,7 +77,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     goto quit;
   }
 
-  image_IDs = (heif_item_id*)alloca(images_count * sizeof(heif_item_id));
+  image_IDs = (heif_item_id*)malloc(images_count * sizeof(heif_item_id));
 
   for (int i = 0; i < images_count; ++i) {
     err = heif_context_get_image_handle(ctx, image_IDs[i], &handle);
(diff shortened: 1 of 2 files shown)


src/file-fuzzer.cc
```diff
@@ -19,7 +19,6 @@
  */
 
 #include <assert.h>
-#include <alloca.h>
```
fancycode: You have to include `stdlib.h` now for `malloc`/`free` (see error in Travis)..

src/file-fuzzer.cc
```diff
@@ -103,5 +102,6 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
 
 quit:
   heif_context_free(ctx);
+  free (image_IDs);
```
fancycode: No whitespace between function name and parameters.

src/heif_image.cc
```diff
@@ -896,7 +896,8 @@ Error HeifPixelImage::fill_RGB_16bit(uint16_t r, uint16_t g, uint16_t b, uint16_
     int stride = plane.stride;
     uint8_t* data = plane.mem.data();
 
-    uint16_t val16;
+    // initialization only to avoid a warning of uninitalized variable.
+    uint16_t val16 = r;
```
fancycode: Please set to `0` in the `default` case below before the `assert` instead.

Jehan:
Sorry for the style bugs. We are the first to ask of contributors to follow our style guidelines. I should have been more careful. ;-)
Anyway I fixed and rebased the commits.

> Btw, which compiler are you using? We are testing with AppVeyor where building with Visual Studio 14 2015 succeeds.

I cross-build from Linux with mingw-w64.

fancycode:
> I cross-build from Linux with mingw-w64.

Hmm, we should cover that with Travis then :smile:

PR looks good now, thanks! I'll merge when CI is green.

Issue #28, opened 2018-05-24
Linking fails without -no-undefined


Jehan:
By the way, as a followup of PR #25, there is a last problem whose solution, I haven't found yet.
A simple make (crossbuild with mingw-w64) fails during linking with:

```
make  all-recursive
make[1]: Entering directory '/opt/user-cache/jehan/dev/crossbuild/w64/libheif'
Making all in src
make[2]: Entering directory '/opt/user-cache/jehan/dev/crossbuild/w64/libheif/src'
/bin/sh ../libtool  --tag=CXX   --mode=link x86_64-w64-mingw32-g++  -I/home/jehan/.local/share/crossroad/roads/w64/libheif/include   -DLIBHEIF_EXPORTS  -g -O2 -Wall -Werror -Wsign-compare -Wconversion -Wno-sign-conversion -Wno-error=conversion -version-info 2:0:1  -o libheif.la -rpath /home/jehan/.local/share/crossroad/roads/w64/libheif/lib libheif_la-bitstream.lo libheif_la-box.lo libheif_la-error.lo libheif_la-heif_file.lo libheif_la-heif_image.lo libheif_la-heif.lo libheif_la-heif_context.lo libheif_la-heif_hevc.lo libheif_la-heif_plugin_registry.lo libheif_la-heif_plugin.lo libheif_la-heif_decoder_libde265.lo  -lde265 -L/home/jehan/.local/share/crossroad/roads/w64/libheif/lib   -lpthread 
libtool:   error: can't build x86_64-w64-mingw32 shared library unless -no-undefined is specified
make[2]: *** [Makefile:647: libheif.la] Error 1
make[2]: Leaving directory '/opt/user-cache/jehan/dev/crossbuild/w64/libheif/src'
make[1]: *** [Makefile:502: all-recursive] Error 1
make[1]: Leaving directory '/opt/user-cache/jehan/dev/crossbuild/w64/libheif'
make: *** [Makefile:411: all] Error 2
```

A workaround exists, which is to build with `make LDFLAGS=-no-undefined` instead (as proposed by the error message). Then it goes to the end.
I'm not sure of the cause yet. Maybe some library is missing on the linking flags.

fancycode:
Fixed by #29

PR #29, opened 2018-05-25
Pass "-no-undefined" when linking with MinGW.

fancycode:
This should fix the error reported in #28.

@Jehan: could you please check if this fixes the error you were getting?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -25,6 +25,7 @@ AC_PROG_CC
 AC_PROG_INSTALL
 AC_PROG_LN_S
 AC_PROG_GREP
+AM_CONDITIONAL(MINGW, expr $host : '.*-mingw' >/dev/null 2>&1)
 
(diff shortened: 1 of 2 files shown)


Jehan:
@fancycode Yes it does "fix" the build, since it does the same thing as adding the link flag on command line. But is it really a "fix"?
I felt like it is more a workaround of a problem we didn't get (yet). Otherwise I would just included this fix in my previous PR. :-)

Or am I wrong? Do you know exactly what this error was?

fancycode:
For reference, here is some background on the flag:
http://article.gmane.org/gmane.comp.gnu.mingw.user/18727

> Common practice in packages is to either use `-no-undefined`
unconditionally, or, when that turns out not to work everywhere,
to use `-no-undefined` on $host_os systems matching `cygwin*` or
`mingw*`.[3]  In case of doubt, the latter is pretty safe.

Issue #31, opened 2018-06-06
wrong padding bits before min_spatial_segmentation_idc


yoya:
# patch

```
-  writer.write16(c.min_spatial_segmentation_idc & 0x0FFF);
+  writer.write16((c.min_spatial_segmentation_idc & 0x0FFF) | 0xF000);
```

# reference

- https://mpeg.chiariglione.org/standards/mpeg-h/image-file-format/text-isoiec-cd-23008-12-image-file-format
  - w14148.zip => W14148-HEVC-still-WD.doc

```
aligned(8) class HEVCDecoderConfigurationRecord {
(omit)
	bit(4) reserved = ‘1111’b;
	unsigned int(12) min_spatial_segmentation_idc;
(omit)

farindk:
Yes, thanks for spotting this. Fixed in accaddc.

PR #32, opened 2018-06-06
fixed to padding bits before min_spatial_segmentation_idc.

yoya:
issue #31 : wrong padding bits before min_spatial_segmentation_idc
https://github.com/strukturag/libheif/issues/31


farindk:
already fixed in accaddc.

PR #59, opened 2018-08-14
Format .go files

leonklingele-work:
Done by running the following command
    find . -name *.go -exec go fmt {} \;


go/src/heif-test.go
```diff
@@ -20,7 +20,6 @@
 
 package main
 
-
 /*
 #cgo CPPFLAGS: -I/home/domain/farindk/sys/include/libheif
```
fancycode: While you're at it, this should be the same as in https://github.com/strukturag/libheif/blob/8a3020dc8d26b18f2f5bc202af5024086281b7e4/go/src/heif/heif.go#L24
e.g. without the hardcoded include path and updated `#include <libheif/heif.h>`

leonklingele-work: #61

scripts/gofmt.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -eufo pipefail
+
+GIT_ROOT="$(git rev-parse --show-toplevel)"
```
fancycode: This won't work if not running from a git checkout.
Please determine folder the script is located at using
```bash
DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
```

Also the folder should not be changed, you could pass it to `find` instead:
```
ERR=$(find $DIR/.. -name *.go -exec go fmt "{}" \;)
```

scripts/run-ci.sh
```diff
@@ -32,6 +32,12 @@ if [ ! -z "$CPPLINT" ]; then
     exit 0
 fi
 
+if [ ! -z "$GOFMT" ]; then
```
fancycode: Please reuse the existing variable `CPPLINT` to determine if `gofmt.sh` should be run.
Also `ABORT_ON_ERR` should set before calling so the CI job fails on formatting errors.

scripts/pre-commit.hook
```diff
@@ -45,3 +45,11 @@ if [ "$check_enums" = "1" ]; then
     fi
     "$CHECK_EMSCRIPTEN"
 fi
+
+# Assume running from ".git/hooks" folder.
```
fancycode: A better way would be to only check modified .go files instead of running gofmt on all commits:
https://github.com/strukturag/spreed-webrtc/blob/master/src/hooks/pre-commit.hook#L6

scripts/gofmt.sh
leonklingele-work: Updated.

> Also the folder should not be changed

It should be fine to change the folder, the script is launched in its own bash process including  a separate env.

scripts/pre-commit.hook
leonklingele-work: I decided to go with a single `gofmt.sh` file.
Using your approach would be slightly more efficient but carries additional code which will make things too complex imo.

fancycode: Yes, but running `go fmt` on all .go files for any commit is quite a lot overhead. Especially given that most changes in the repository will only affect C++ code.

leonklingele-work: Agreed, updated again.

scripts/run-ci.sh
```diff
@@ -32,6 +32,12 @@ if [ ! -z "$CPPLINT" ]; then
     exit 0
 fi
 
+if [ ! -z "$CPPLINT" ]; then
+    echo "Running gofmt ..."
```
fancycode: This must be included in the `if [ ! -z "$CPPLINT" ]; then` block above (before the `exit 0`), otherwise it will not be reached.

scripts/pre-commit.hook
fancycode: Please use the whole `for file in ` block here instead of calling `gofmt.sh`.

scripts/gofmt.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
fancycode: Please rename to `check-gofmt.sh` for consistency with other check scripts and remove the conditions `ONLY_GIT_STAGED` (see below) and `ABORT_ON_ERR` (should always fail on errors).

If we need a separate way to format Go code besides calling `go fmt <filename>`, there should be a make target like `make format` or similar (not needed for now).

fancycode:
Make sure to include all new files in `EXTRA_DIST` of the corresponding `Makefile.am` files.

scripts/check-gofmt.sh
```diff
@@ -0,0 +1,28 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with libheif.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
+
+ERR=$(find "$DIR"/.. -name *.go -exec go fmt "{}" \;)
+if [[ (-v ABORT_ON_ERR) && (-n "$ERR") ]]; then
```
fancycode: I think this should be `if [ -n "$ERR" ]; then`? You are not setting `ABORT_ON_ERR` anywhere and the script should fail on errors.

scripts/run-ci.sh
```diff
@@ -29,6 +29,9 @@ if [ ! -z "$CPPLINT" ]; then
     echo "Running cpplint ..."
     find -name "*.cc" -o -name "*.h" | sort | xargs ./scripts/cpplint.py
     ./scripts/check-emscripten-enums.sh
+
+    echo "Running gofmt ..."
+    ./scripts/check-gofmt.sh || (echo "Go files are not properly formatted." ; exit 1)
```
fancycode: Running `./scripts/check-gofmt.sh` should be sufficient. This script is `set -e`, so it will fail if any of the called scripts fail (which `check-gofmt.sh` will do on errors).

scripts/check-gofmt.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -eufo pipefail
+#
+# HEIF codec.
+# Copyright (c) 2018 struktur AG, Joachim Bauch <bauch@struktur.de>
```
fancycode: Feel free to use your name here ;-)

scripts/check-gofmt.sh
leonklingele-work: Right, fixed.

scripts/run-ci.sh
leonklingele-work: But it won't show a nice error message :)
Anyway, removed the `echo` call.

scripts/check-gofmt.sh
leonklingele-work: Updated.

fancycode:
Feel free to merge when CI is green.

PR #67, opened 2018-08-21
go: Call 'heif_context_read_from_memory_without_copy' instead of deprecated function

leonklingele-work:
heif.c:319 says:
    // DEPRECATED: use heif_context_read_from_memory_without_copy() instead.

### DIFF
--- a/go/heif/heif.go
+++ b/go/heif/heif.go
@@ -321,7 +321,12 @@ func (c *Context) ReadFromFile(filename string) error {
 
 func (c *Context) ReadFromMemory(data []byte) error {
 	// TODO: Use reader API internally.
-	err := C.heif_context_read_from_memory(c.context, unsafe.Pointer(&data[0]), C.size_t(len(data)), nil)
+	cData := C.CBytes(data)
+	runtime.SetFinalizer(&cData, func(p *unsafe.Pointer) {
+		C.free(*p)
+		runtime.SetFinalizer(cData, nil)
+	})
+	err := C.heif_context_read_from_memory_without_copy(c.context, cData, C.size_t(len(data)), nil)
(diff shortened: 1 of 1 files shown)


farindk:
Please make sure that this does not introduce hard to find conflicts with the garbage collector.
If you use `*from_memory_without_copy`, it means that libheif assumes that the memory stays valid while the `heif_context` is still alive. If the Go GC decides to release the `byte[]` before that, you get into problems.

leonklingele-work:
@farindk ptal. This now allocates C memory and frees it again before exiting.

farindk:
Doesn't that just set the finalizer for the copied array? I.e. it can still garbage-collect the array before the heif_context? I think the finalizer should be attached to the context, such that the array lifetime is at least as long as the context.

Then again, I have no idea about Go... So I might be wrong.

go/heif/heif.go
```diff
@@ -321,7 +321,12 @@ func (c *Context) ReadFromFile(filename string) error {
 
 func (c *Context) ReadFromMemory(data []byte) error {
 	// TODO: Use reader API internally.
-	err := C.heif_context_read_from_memory(c.context, unsafe.Pointer(&data[0]), C.size_t(len(data)), nil)
+	cData := C.CBytes(data)
```
fancycode: Why not store a reference to `data` in the `Context`? This would make sure it is kept alive until the `Context` is released.

If you do that, make sure to clear the referene manually if it is no longer needed (e.g. when loading from a file in the same context).

fancycode (changes requested):

farindk:
@fancycode Do you want to take over this issue? Otherwise, let's close it.

PR #68, opened 2018-08-21
gofmt

leonklingele-work:


leonklingele-work:
@fancycode any idea why the precommit hook no longer complained? Seems like the issue was introduced in 30251779113b7e61e26c62b3f08606f298ff8d70 or so.

farindk:
Nothing special during the commit. Yes, I think there was an error shown, but it didn't prevent it from committing.
Which is nice, BTW, because I do not want to dig out Go formatting documentation just because I had to do a quick fix.

If Go formatting is that important, why not simply run the formatting command automatically on check-in instead of printing an error and leaving the committer alone with the problem. :-)

leonklingele-work:
> Go formatting documentation just because I had to do a quick fix.

Just run `find . -name *.go -exec go fmt "{}" \;`

> If Go formatting is that important, why not simply run the formatting command automatically on check-in instead of printing an error and leaving the committer alone with the problem. :-)

That was my initial approach in #59 which got rejected. Maybe we want to nag at users to learn proper formatting? :)

fancycode:
The pre-commit hook should show an error message that also contains the necessary commands to fix the error:
https://github.com/strukturag/libheif/blob/master/scripts/pre-commit.hook#L63

fancycode:
> That was my initial approach in #59 which got rejected.

You approach was formatting the files on Travis, which wouldn't fix the formatting errors introduced by commits.

PR #64, opened 2018-08-20
Various improvements to the updated Go library and example app

leonklingele-work:
- Register another heif image format header
- Remove debug code
- Don't pre- _and_ suffix error variables
- Better output formatting of example
- Goify names (remove underscores, uppercase `ID`, etc.)

PR #69, opened 2018-08-22
Add API to encode image to heif format

leonklingele-work:


leonklingele-work:
This PR is based on #64, please merge that one before.
To review the patch, check 36a1d6fcf7705b1e6dba603406ebebdb347edc5d.

go/heif/heif.go
```diff
@@ -602,43 +694,259 @@ func (img *Image) GetImage() (image.Image, error) {
...
+}
+
+func EncodeFromImage(img image.Image, compression Compression, quality int, lossless losslessMode, logging loggingLevel) (*Context, error) {
+	var out *Image
+
+	switch cm := img.ColorModel(); cm {
+	default:
+		return nil, fmt.Errorf("color model %v not supported yet", cm)
+	case color.NRGBAModel:
+		i, ok := img.(*image.NRGBA)
```
fancycode: Checking the `ColorModel()` is not sufficient to determine the type. There could be implementations of `image.Image` with a color model `color.NRGBAModel` that are not subclasses of `image.NRGBA`.

Better would be to use a type switch (https://tour.golang.org/methods/16).

go/heif/heif.go
```diff
@@ -602,43 +694,259 @@ func (img *Image) GetImage() (image.Image, error) {
...
+	pY.SetData([]byte(i.Pix), i.Stride)
+
+	return out, nil
+}
+
+func imageFromYCbCr(i *image.YCbCr) (*Image, error) {
+	max := i.Bounds().Max
+	w, h := max.X, max.Y
+
+	out, err := NewImage(w, h, ColorspaceYCbCr, Chroma420)
```
fancycode: You need to check the `SubsampleRatio` of the `image.YCbCr`, it could also be `4:2:2` or `4:4:4`.

go/heif/heif.go
```diff
@@ -602,43 +694,259 @@ func (img *Image) GetImage() (image.Image, error) {
...
+	}
+
+	const depth = 8
+	pY, err := out.NewPlane(ChannelY, w, h, depth)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add Y plane: %v", err)
+	}
+	pY.SetData([]byte(i.Y), i.YStride)
+
+	halfW, halfH := (w+1)/2, (h+1)/2
```
fancycode: Might need to be updated for other `SubsampleRatio` values.

fancycode:
#64 has landed, please rebase.

go/heif/heif.go
```diff
@@ -323,7 +339,73 @@ func (c *Context) ReadFromMemory(data []byte) error {
...
+}
+
+func (c *Context) NewEncoder(compression Compression) (*Encoder, error) {
+	const max = 1
+	descriptors := make([]*C.struct_heif_encoder_descriptor, max)
+	num := int(C.heif_context_get_encoder_descriptors(c.context, uint32(compression), nil, &descriptors[0], C.int(max)))
+	if num == 0 {
+		return nil, fmt.Errorf("no encoder for compression %v", compression)
+	}
+	return c.convertEncoderDescriptor(descriptors[0])
```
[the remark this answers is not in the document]
leonklingele: It just cleans up the code a bit and will probably be inlined by the Go compiler, so not even a performance issue. Shall I still inline it manually?

go/heif/heif.go
```diff
@@ -602,43 +694,259 @@ func (img *Image) GetImage() (image.Image, error) {
 }
 
 type ImageAccess struct {
-	Plane  []byte
-	Stride int
+	Plane   []byte
+	Stride  int
+	Height  int
+	Pointer unsafe.Pointer
```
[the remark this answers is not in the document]
leonklingele-work: Good spot, fixed.

go/heif/heif.go
```diff
@@ -602,43 +694,259 @@ func (img *Image) GetImage() (image.Image, error) {
...
+}
+
+func freeHeifEncodingOptions(options *EncodingOptions) {
+	C.heif_encoding_options_free(options.options)
+	options.options = nil
+	runtime.SetFinalizer(options, nil)
+}
+
+func hasAlpha(i image.Image) bool {
+	rect := i.Bounds()
```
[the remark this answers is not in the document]
leonklingele-work: > implementations that are known to not have an alpha channel

`*image.CMYK, *image.Gray, *image.Gray16, *image.YCbCr`
Does this cover them all?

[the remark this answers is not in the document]
leonklingele-work: Will remove the func altogether as we apparently only need it for `NRGBA` images.
I have no idea what I'm doing, let's hope this'll work out.

go/heif/heif.go
leonklingele-work: Well, that's why I do the type assertion (you have even commented on this exact line).
Shall I remove the ColorModel() switch and go for a type switch? The current implementation uses both which seems redundant to me, but I have no clue about how images work _at all_ and what needs to be checked.

go/heif/heif.go
leonklingele-work: Like that?

```go
var chroma Chroma
switch i.SubsampleRatio {
case image.YCbCrSubsampleRatio444:
	chroma = Chroma444
case image.YCbCrSubsampleRatio422:
	chroma = Chroma422
case image.YCbCrSubsampleRatio420:
	fallthrough
default:
	chroma = Chroma420
}
```

leonklingele-work:
@fancycode rebased & addressed some of your comments. PTAL and please also help me with the remaining issues :)

go/heif/heif.go
```diff
@@ -637,6 +754,217 @@ func (img *Image) ScaleImage(width int, height int) (*Image, error) {
...
+	var cm Chroma
+	switch i.SubsampleRatio {
+	case image.YCbCrSubsampleRatio444:
+		cm = Chroma444
+	case image.YCbCrSubsampleRatio422:
+		cm = Chroma422
+	case image.YCbCrSubsampleRatio420:
+		fallthrough
+	default:
+		cm = Chroma420
```
leonklingele-work: Default could return an error as well, what do you think @fancycode?

fancycode: Yes, please return an error for `default` as the subsample ratio will not be supported in this case.

go/heif/heif.go
```diff
@@ -600,32 +694,55 @@ func (img *Image) GetImage() (image.Image, error) {
...
-	Stride int
+	Plane    []byte
+	planePtr unsafe.Pointer
+	Stride   int
+	Height   int
 
 	image *Image // need this reference to make sure the image is not GC'ed while we access it
 }
 
+func (i *ImageAccess) SetData(data []byte, stride int) {
```
fancycode: Please also require `width` and `height` to be passed and check their values (`width` and `height` must match values from `ImageAccess` object), otherwise this could lead to segfaults (if `data` is too small or `width`, `height` or `stride` are too large). Also check if the size of `data` is `stride * height`.

You can then handle the common case directly:
```go
if stride == i.Stride {
  i.Plane = data
  i.planePtr = unsafe.Pointer(&i.Plane[0])
  return
}
```

Maybe it's easier to not expose `SetData` for now if not required to be called by apps using the library.

go/heif/heif.go
```diff
@@ -600,32 +694,55 @@ func (img *Image) GetImage() (image.Image, error) {
 }
 
 type ImageAccess struct {
-	Plane  []byte
-	Stride int
+	Plane    []byte
+	planePtr unsafe.Pointer
+	Stride   int
+	Height   int
```
fancycode: Why are you not exposing `Width`?

go/heif/heif.go
```diff
@@ -637,6 +754,217 @@ func (img *Image) ScaleImage(width int, height int) (*Image, error) {
...
+	}
+	pCr.SetData([]byte(i.Cr), i.CStride)
+
+	return out, nil
+}
+
+func EncodeFromImage(img image.Image, compression Compression, quality int, lossless losslessMode, logging loggingLevel) (*Context, error) {
+	var out *Image
+
+	switch cm := img.ColorModel(); cm {
```
fancycode: Yes, please change to use a type switch, this removes your redundant checks.

go/heif/heif.go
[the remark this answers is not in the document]
fancycode: Feel free to keep it as-is if you want.

go/heif/heif.go
```diff
@@ -637,6 +754,217 @@ func (img *Image) ScaleImage(width int, height int) (*Image, error) {
...
+	}
+
+	const depth = 8
+	pY, err := out.NewPlane(ChannelY, w, h, depth)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add Y plane: %v", err)
+	}
+	pY.SetData([]byte(i.Y), i.YStride)
+
+	// TODO: Might need to be updated for other SubsampleRatio values.
```
fancycode: This needs to be implemented if you support 4:2:2 and 4:4:4 above.

leonklingele-work: As discussed, I will remove support for subsample ratios other than 4:2:0.

go/heif/heif.go
leonklingele-work: As discussed, I will unexport the method for now and only handle the common case (`stride == i.Stride `) without checking `width` or `height`.

go/heif/heif.go
leonklingele-work: Apparently only the height is required (for internal purposes). I will unexport `Height` for now.

leonklingele-work:
I've address your comments in a separate commit so you don't have to recheck the whole code of this PR. See https://github.com/strukturag/libheif/pull/69/commits/73afba34313a48de2d5a06550f14fea7df2f1e0a. I will squash once accepted.

Changes:
- Only support YCbCr subsample ration 4:2:0
- Export types `LosslessMode` and `LoggingLevel`
- Unexport `ImageAccess.height`
- Unexport `ImageAccess.setData` and handle common case (`stride == i.Stride`) directly
- Use type switch instead of checking the color model
- Rename a variable

go/heif/heif.go
```diff
@@ -871,56 +872,56 @@ func imageFromYCbCr(i *image.YCbCr) (*Image, error) {
...
 
 	return out, nil
 }
 
-func EncodeFromImage(img image.Image, compression Compression, quality int, lossless losslessMode, logging loggingLevel) (*Context, error) {
+func EncodeFromImage(img image.Image, compression Compression, quality int, lossless LosslessMode, logging LoggingLevel) (*Context, error) {
 	var out *Image
 
-	switch cm := img.ColorModel(); cm {
+	switch t := img.(type) {
```
fancycode: This could even be simplified more:
```go
switch i := img.(type) {
case *image.NRGBA:
  // i has now type *image.NRGBA
case *image.Gray:
  // i has now type *image.NRGBA
...
}
```

You don't need to cast inside the `case`, see https://tour.golang.org/methods/16 for details.

leonklingele-work: TIL. Thanks, much better now :)

fancycode:
lgtm :+1: feel free to merge when CI is all-green

PR #79, opened 2018-09-12
go: Delegate responsibility to free C-allocated memory to caller

leonklingele-work:
This fixes an issue where struct members were passed to external C
functions after which the struct was no longer referenced.
As described in the docs of runtime.SetFinalizer, this would allow
the Go garbage collector to run every finalizer associated with the
struct (as it will no longer be used). Unfortunately, as we bound
a finalizer to free every C-allocated struct member, those
finalizers were run, freeing the struct members too early.

This patch currently requires the caller to manually call .Free()
on the following structs once they are no longer required:
- Context
- ImageHandle
- DecodingOptions
- Image

Such structs are currently returned by the following functions:
- NewContext
- NewDecodingOptions
- NewImage
- Context.GetPrimaryImageHandle
- Context.GetImageHandle
- ImageHandle.GetDepthImageHandle
- ImageHandle.GetThumbnail
- ImageHandle.DecodeImage
- Image.ScaleImage

### DIFF
--- a/examples/heif-test.go
+++ b/examples/heif-test.go
@@ -78,6 +78,7 @@ func testHeifLowlevel(filename string) {
 		fmt.Printf("Could not create context: %s\n", err)
 		return
 	}
+	defer c.Free()
 
 	if err := c.ReadFromFile(filename); err != nil {
 		fmt.Printf("Could not read file %s: %s\n", filename, err)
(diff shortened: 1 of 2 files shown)


fancycode (changes requested):
You should rather use `runtime.KeepAlive` to prevent the finalizers from running too early (https://golang.org/pkg/runtime/#KeepAlive).

With your change, you will introduce hard to track memory leaks if users forget to call `Free()` on their garbage-collected objects before they run out of scope.

leonklingele-work:
If users forget to call `runtime.KeepAlive` it will simply crash their app in very rare circumstances (depends on when the GC is run). This is so much harder and more frustrating to debug.
Even if this is a Go API, users using it should have the mindset of dealing with C code where heap data needs to be freed manually.
Also, do we really want callers to import `runtime` just to use this library?

leonklingele-work:
As none of the struct members is exposed in any way (not exported, not returned by any function), the issue can be solved much easier. See #81.

PR #86, opened 2018-09-24
Allow 10-bit and 12-bit encoding

jonsneyers:
As it is now, libheif encoding (with x265) would force an 8-bit encoding (which would not give the expected result if the input image has a higher bit depth). This pull request makes it use the 10-bit or 12-bit variant of x265 if the input image is 10-bit or 12-bit.

I also added or changed some of the x265 parameters; in general, these are to get better compression and better perceptual quality, at the expense of encode speed (since for still images I assume encode speed is not that important but compression and perceptual quality are).

(also libpng introduced png_const_bytep a bit later than what the ifdef said, so the code didn't compile on some libpng versions)

### DIFF
--- a/examples/encoder_png.cc
+++ b/examples/encoder_png.cc
@@ -85,7 +85,7 @@ bool PngEncoder::Encode(const struct heif_image_handle* handle,
     heif_image_handle_get_raw_color_profile(handle, profile_data);
     char profile_name[] = "unknown";
     png_set_iCCP(png_ptr, info_ptr, profile_name, PNG_COMPRESSION_TYPE_BASE,
-#if PNG_LIBPNG_VER <= 10250
+#if PNG_LIBPNG_VER < 10500
                  (png_charp)profile_data,
 #else
                  (png_const_bytep)profile_data,
--- a/libheif/heif_encoder_x265.cc
+++ b/libheif/heif_encoder_x265.cc
@@ -52,6 +52,7 @@ struct x265_encoder_struct
   std::string tune;
   int tu_intra_depth;
   int logLevel = X265_LOG_NONE;
+  int bit_depth;
 };
 
 
@@ -128,7 +129,7 @@ static void x265_init_parameters()
   p->version = 1;
   p->name = kParam_preset;
   p->type = heif_encoder_parameter_type_string;
-  p->string.default_value = "slow";
+  p->string.default_value = "veryslow";
   p->string.valid_values = kParam_preset_valid_values;
   d[i++] = p++;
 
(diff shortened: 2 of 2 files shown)


farindk:
I've cherry-picked the libpng version check into 'master' and added the 10-bit encoding to the new 'hdr' branch.

farindk:
Yes, of course. The 'hdr' branch is only for development until we are sure that everything works. Then we will merge it into `master`.

I noticed that encoding speed with the new x265 settings is >10x slower on my (quite fast) computer (>7 secs for a 1440x960 image), which might be way too much. Hence, we'll first need some user-controllable option to set the compression mode.

jonsneyers:
I think there is already a way to do that, it's just that I changed the default setting from "slow" to "veryslow" (to get better compression results). Maybe some of the other options also have an impact, but I think it's probably just that. Feel free to change the default back to "slow".

farindk:
slow vs. veryslow actually is not that much of a difference. When you look into the current 'hdr' branch, you can see that I added a "complexity" parameter to tweak the computation time / compression ratio.

These parameters seem to affect computation time:
- wpp = 0   this is the killer as it effectively turns off multi-threading
- cu-lossless = 1   probably many more modes to search
- rd-refind, rd   seem to increase computation time a little bit

I'm not that happy yet with the 'complexity' parameter, as it is not clear which of the settings has precedence. Have to think about that a bit more...

Jamaika1:
Hi team
I applied the jonsneyers patch by changing 8bit to 10bit.
Is it possible to add a jpeg 10 / 12bit codec for heif 10 / 12bit 420p?
jpeg(10bit)->heif(10bit) or jpeg(12bit)->heif(12bit)
It seems so simple, but I can't do that.

farindk:
There is no full HDR (>8 bit) image path in libheif yet. Color-space conversions, scaling functions, and more has to be adapted first. And yes, this is something we have on the roadmap.

jonsneyers:
The >8 bit image path indeed still requires some work, but now the complexity issue has been dealt with, can the hdr branch be merged to master? The functionality as it is now, is useful already – we use libheif with YCbCr 4:2:0 buffers, and for that use case, >8 bit already works. We also want to make use of the slower/better compression. We can work with the hdr branch, but I think it would be better to merge it with master before it starts diverging too much...

farindk:
I have reworked the parameter handling today. In the old parameter handling, it was not defined, in which order the parameters were set in x265. This left it unclear how a parameter like 'complexity' interacts with other parameters.
The new parameter handling now explicitly sets the parameters in the order they are specified. An earlier parameter may be overwritten by later parameters.

As an additional feature, you can now pass arbitrary x265 parameters by prefixing them with 'x265:'.
Hence, you can write `./heif_enc -p x265:ctb=32 -p x265:wpp=0`

The `complexity` parameter is still there, but I would not rely on a specific parameter mapping. If you want to make sure that exactly your parameters are used, you better use the 'x265:' prefix syntax.

The `hdr` branch is not merged into master, but it will stay active to implement real 10/12 bit support.

farindk:
Patches have already been integrated for a long time... Closing this.

PR #92, opened 2018-10-30
WebAssembly

1div0:
I would like to ask you for review of my attempt to compile WebAssembly version of the library. Thank you.


farindk:
A better way instead of writing `success=success;` is to write `(void)success`.
This makes it explicit to the compiler that the value is not used.

1div0:
Thank you for your advise. I will rewrite that accordingly.

fancycode:
Please take a look at the failing tests and update your changes accordingly.

PR #103, opened 2019-01-18
add color profile embedding example if input jpeg has an embedded color profile

asuwendi:
add color profile embedding example if input jpeg has an embedded color profile. This will use heif_image_set_raw_color_profile .

### DIFF
--- a/examples/heif_enc.cc
+++ b/examples/heif_enc.cc
@@ -58,6 +58,9 @@ extern "C" {
 
 #include <assert.h>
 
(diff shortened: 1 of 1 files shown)


farindk:
Thank you for this patch. This is great.
Please see my comment above and tell me whether we can modify the patch to only save the APP2 markers.

asuwendi:
Hi farindk, not a problem at all if what you want is to modify the patch to only save the APP2 markers.

Issue #147, opened 2019-08-28
segfault: heif.ColorspaceUndefined, heif.ChromaUndefined


duedal:
When decoding images in a multi-threaded environment, with Colorspace and/or Chroma set to undefined, it will segfault with `malloc: pointer being freed was not allocated`

Stable reproduce "script": https://gist.github.com/duedal/fd846c0d24d0d90db60fcb1862946040

Test image; https://gist.github.com/duedal/fd846c0d24d0d90db60fcb1862946040/raw/e3d68d623d0444c573f5a83fc4fb77197ccf54d3/world-political.heic

Reproduce
`go run race.go  -input world-political.heic`

if you run with `-rgb` it will not crash
`go run race.go  -input world-political.heic -rgb`

farindk:
Thanks.
Is this with the current master branch or with v1.5.0 ?
I'm asking because this sounds very much like a bug I fixed yesterday (dfb0eb972392e68).

duedal:
This was with v1.5.0 - I'll try with master

duedal:
Still happens with master
```
_debug_bin(76027,0x70000b9b4000) malloc: *** error for object 0x9126370: pointer being freed was not allocated
__debug_bin(76027,0x70000b9b4000) malloc: *** set a breakpoint in malloc_error_break to debug
...repeat
__debug_bin(76027,0x700019f77000) malloc: *** error for object 0x8c5e0a0: pointer being freed was not allocated
__debug_bin(76027,0x700019f77000) malloc: *** set a breakpoint in malloc_error_break to debug
__debug_bin(76027,0x70000e327000) malloc: Incorrect checksum for freed object 0x90623f8: probably modified after being freed.
Corrupt value: 0x400000000090377f
__debug_bin(76027,0x70000e327000) malloc: *** set a breakpoint in malloc_error_break to debug
...repeat
__debug_bin(76027,0x70000bcc6000) malloc: *** error for object 0x9126630: pointer being freed was not allocated
__debug_bin(76027,0x70000bcc6000) malloc: *** set a breakpoint in malloc_error_break to debug
(dlv) bt
 0  0x00007fff7c1a086a in ???
    at ?:-1
 1  0x00007fff7929aa0a in ???
    at ?:-1
 2  0x00007fff792a396b in ???
    at ?:-1
 3  0x00000000042c6220 in ???
    at ?:-1
 4  0x00000000042c12c6 in ???
    at ?:-1
 5  0x00000000042be713 in ???
[...]

PR #148, opened 2019-08-29
CI: Run race tester as part of Travis GO job.

fancycode:
To prevent regressions of #147.


fancycode:
@farindk I have updated this PR to only include the test now that branch `issue-147` has landed.

PR #154, opened 2019-09-03
fix shared install on cross environments

tgurr:
Do not install arch independent files into prefix, which can be e.g.
/usr/x86_64-pc-linux-gnu for cross environments so instead of the files
landing in the arch independent /usr/share they wrongly get installed into
the arch dependent /usr/x86_64-pc-linux-gnu/share location.


fancycode:
Thanks!

PR #156, opened 2019-09-11
Add man pages for example apps.

fancycode:
This is a first draft open for discussion.

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -9,11 +9,14 @@ examples_noinst = \
   heif-test \
   test-c-api
 
+dist_man_MANS =
+
 heif_convert_DEPENDENCIES = ../libheif/libheif.la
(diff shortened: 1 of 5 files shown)


farindk:
This is good, but what I always wanted to do is to clean-up these command line tools. The way they take arguments and options is not consistent. It even confuses myself. Maybe heif-convert and heif-enc could be combined into a single program. Thus, I am not sure we should publish the man pages before this is cleaned up.

fancycode:
Oh ok, I wasn't aware of that. Depending on when the refactoring of the command line tools will happen, it probably still makes sense to add man pages for what we have right now and change them later together with the tools.

farindk:
We can clean up the tools in the next release. I'm just a bit hesitant as I do not want to break any existing stuff.

PR #166, opened 2019-10-14
Fixed libde265_{v1,v2}_push_data to handle integer overflows on 32-bit

alex:


alex:
Travis failures appear to be an unrelated macOS brew failure.

farindk:
Thank you!

alex:
Thanks for the quick merge! This came out of an OSS-Fuzz finding -- let me CC you on it so you hae context.

fancycode:
> This came out of an OSS-Fuzz finding.

Interesting, we are also using OSS-fuzz for libheif but didn't get this finding (yet). Could you please add your testcase in a PR to our fuzzing corpus in https://github.com/strukturag/libheif/tree/master/fuzzing/corpus (or attach it here so I can do this)? Thanks!

alex:
Do you have 32-bit fuzzing enabled? That's the only way to trigger it.


[clusterfuzz-testcase-minimized-encoder_heic_fuzzer-5698815994101760.zip](https://github.com/strukturag/libheif/files/3725182/clusterfuzz-testcase-minimized-encoder_heic_fuzzer-5698815994101760.zip)

fancycode:
> Do you have 32-bit fuzzing enabled? That's the only way to trigger it.

Oh, I don't think so. Thanks for the heads up, we should probably look into enabling this.

PR #175, opened 2019-11-07
Add `-module` flag to get correct extension on macOS

samm-git:
Without `-module` flag libtool generats `.dylib` extension and pixbuf expects `.so`


farindk:
Thank you.

PR #194, opened 2020-01-08
Fix for the gdk pixbuf `gdk_pixbuf_new_from_file_at_size` call

samm-git:
1. If new size was requested and image was scaled - redefine width/height to the new values
2. Do not duplicate `heif_image_get_width` for no reason
3. Send image width/height to the size_func call as other pixbuf loaders doing.

This fixing `gdk_pixbuf_new_from_file_at_size` for me.

### DIFF
--- a/gdk-pixbuf/pixbufloader-heif.c
+++ b/gdk-pixbuf/pixbufloader-heif.c
@@ -99,19 +99,19 @@ static gboolean stop_load(gpointer context, GError** error)
 
   width = heif_image_get_width(img, heif_channel_interleaved);
   height = heif_image_get_height(img, heif_channel_interleaved);
-  requested_width = 0;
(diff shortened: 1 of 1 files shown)


samm-git:
P.S. testcase for the issue:

```
#include <gdk-pixbuf/gdk-pixbuf.h>

int main(void) {
  GdkPixbuf * test = gdk_pixbuf_new_from_file_at_size("/Users/asamorukov/tmp/autumn_1440x960.heic", 100, 100 ,NULL);

  if(test) {
    printf("width:%d, height: %d\n", gdk_pixbuf_get_width(test), gdk_pixbuf_get_height(test));
  }
  return 0;
}
```

will return glib warning and wrong size if not patched

farindk:
Thank you for this fix.

PR #195, opened 2020-01-09
Several fixes to build libheif

cryptomilk:
This fixes building libheif with gcc 9.2.1.

`cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..`

### DIFF
--- a/libheif/heif_encoder_x265.cc
+++ b/libheif/heif_encoder_x265.cc
@@ -598,7 +598,7 @@ static struct heif_error x265_encode_image(void* encoder_raw, const struct heif_
   // x265 cannot encode images smaller than one CTU size
   // https://bitbucket.org/multicoreware/x265/issues/475/x265-does-not-allow-image-sizes-smaller
   // -> use smaller CTU sizes for very small images
-  char ctu_buf[4];
+  const char *ctu = NULL;
   int ctuSize = 64;
 #if 0
   while  (heif_image_get_width(image, heif_channel_Y) < ctuSize ||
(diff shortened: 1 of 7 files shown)


libheif/heif_encoder_x265.cc
```diff
@@ -633,7 +633,14 @@ static struct heif_error x265_encode_image(void* encoder_raw, const struct heif_
 #endif
 
   int s = snprintf(ctu_buf,4,"%d",ctuSize);
-  assert(s<4);
```
farindk: Should probably be `s >= 4`.
I also think that instead of testing the output of `snprintf` here, `ctuSize` should be tested directly to be a power of two in the range [16;64].

cryptomilk: As -DNDEBUG is set, the assert is not compiled in if you build a release version (-O2) and you would still get the unused variable error ...

cryptomilk: Ah, sorry. You mean the if check, yes you're right. Fixed.

cryptomilk:
I implemented your suggestions.

cryptomilk:
I've added more fixes to cmake.

cryptomilk:
I hope that's it now. Those issues have been found while packaging libheif for openSUSE and using it in another project. Will do more tests later today.

farindk:
Thanks, Andreas, I'll have a look at it asap.

cryptomilk:
https://pmbs.links2linux.de/package/show/home:malcolmlewis:branches:Essentials/libheif

The package looks correct now.

farindk:
Great. Thanks again for the changes!

PR #210, opened 2020-03-12
Allow libheif to be compiled in Windows using MinGW

amyspark:
This involves the following changes:

- Use Triton's LibFindMacros to drop pkg-config to an indirect
dependency
- Use LibFindMacros to find libde265 and x265
- Relative-ize gdk-pixbuf install path so as to not break down when
installing as non-admin user
- Fix inclusion of libjpeg-turbo and feature testing
- Enable examples by default in CMake
- Fix missing bits for Appveyor CI: scripts and inclusion variables

The change involving gdk-pixbuf is necessary in Linux and macOS when the library isn't directly installed to `/usr/`. I slipped it here because it's part of my recent [merge request in Krita](https://invent.kde.org/kde/krita/-/merge_requests/258).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,11 +51,9 @@ if (has_potentially_evaluated_expression)
   add_definitions(-Wno-error=potentially-evaluated-expression)
 endif()
 
-if (UNIX)
-  find_package(PkgConfig)
-  pkg_check_modules(LIBDE265 libde265)
-  pkg_check_modules(X265 x265)
-endif()
+LIST (APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules")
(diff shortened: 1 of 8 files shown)


farindk:
Hi @amyspark. I have finally merged your changes (thanks for the contribution).
I had to change `X265_INCLUDE_DIRS` to `X265_INCLUDE_DIR` in FindX265.cmake to make it work for me. Please check that this is correct, as I do not really understand what goes on there.

I have also added a similar detection code for libaom. Would be great if you could cross-check that as I am new to LibFindMacros.

06needhamt:
I know this has already been merged but would it be possible to extend this to include aom for AVIF. I would do it myself but I'm not familiar with cmake.

farindk:
I have already added the checks for aom. Would be good if someone could cross-test this on his system, though.

PR #213, opened 2020-04-09
Fix heif_image_get_plane_readonly in Readme

lou-k:


farindk:
Thank you for spotting this.

PR #221, opened 2020-05-11
Ensure 64-bit int bitstream ranges, prevents possible overflow

lovell:
Thank you for creating and maintaining this very useful library.

I've been reviewing the recently-added AVIF logic and ran into this problem with a possible integer overflow. It looks like all the existing bitstream logic was expecting/able to handle 64-bit integers apart from one method signature, which this PR should hopefully fix.


farindk:
Thanks for cleaning this up.

PR #222, opened 2020-05-11
Expose have_aom status via pkg-config variables

lovell:
This allows for compile-time feature detection of AOM, which will be very useful for consumers of this library, especially where h265 support must be disabled e.g. due to patent issues.

### DIFF
--- a/libheif.pc.in
+++ b/libheif.pc.in
@@ -4,6 +4,8 @@ libdir=@libdir@
 includedir=@includedir@
 builtin_h265_decoder=@have_libde265@
 builtin_h265_encoder=@have_x265@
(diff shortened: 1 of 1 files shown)


farindk:
I would like to rename this to 'builtin_avif_decoder' instead of 'builtin_aom_decoder' (same for encoder), as aom will not be the only available codec plugin. We'll also add support for at least dav1d and rav1e codecs.

Could you please modify the pull-request accordingly? Alternatively, I can also just add this myself.

lovell:
Thanks, now updated.

farindk:
Great. Thank you !

PR #223, opened 2020-05-11
Remove debugging statements from AOM encoder

lovell:
Commented out for now; happy to delete these should that be more preferable.


farindk:
Fine for now. Will delete them eventually.

PR #246, opened 2020-06-21
Fixed adding the aom decoder and encoder files.

dlemstra:
This PR fixes the CMake build when AOM_FOUND is set to 1.


farindk:
Thanks. These changes are already in `master` due to this commit: 3357f013062a57d5c0e9a09ba094f8df9cf1c780.

PR #247, opened 2020-06-22
remove rtti and exception

zljj0818:
For some project will not use libheif.so, but compile libheif to their own libraries with -fno-rtti, and -fno-exception compile flags, so refactor code and remove riit and exception codes


zljj0818:
@farindk please review

zljj0818:
What about this change?

farindk:
The `new (std::nothrow)` has recently been merged to eliminate any use of exceptions. We keep the use of RTTI because it gives us much cleaner code.

PR #248, opened 2020-06-22
Add some more functions to the cxx header

hjmallon:
Some functions that I needed were missing from the C++ header.

### DIFF
--- a/libheif/heif_cxx.h
+++ b/libheif/heif_cxx.h
@@ -190,6 +190,14 @@ namespace heif {
 
     bool has_alpha_channel() const noexcept;
 
+    int get_luma_bits_per_pixel() const noexcept;
+
+    int get_chroma_bits_per_pixel() const noexcept;
+
+    int get_ispe_width() const noexcept;
+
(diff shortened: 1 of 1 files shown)


libheif/heif_cxx.h
```diff
@@ -549,6 +559,22 @@ namespace heif {
     return heif_image_handle_has_alpha_channel(m_image_handle.get()) != 0;
   }
 
+  inline int ImageHandle::get_luma_bits_per_pixel() const noexcept {
+    return heif_image_handle_get_luma_bits_per_pixel(m_image_handle.get()) != 0;
```
farindk, line 563: Why the "!= 0" ? That does not seem correct. Same for the other functions in this block.

farindk:
Thanks. I had a look and think that the "!= 0" in four of the functions is not correct. Maybe a copy-paster error from a function returning boolean. Please correct this, then we can merge it.

hjmallon:
Sorry yes it is absolutely a copy paste from the bool fn.

PR #281, opened 2020-08-11
Update README to latest api

jamiepinheiro:
The sample codes seems to be out of date with the current api for `heif_context_encode_image`


farindk:
Right. Thanks.

PR #292, opened 2020-08-23
Make heif_context mandatory again.

dlemstra:
This PR will make the `heif_context` for the `heif_context_get_encoder` and `heif_context_get_encoder_for_format` mandatory again. There is (now was) a comment in the `heif.h` header file with a question why the context was necessary. The context should be passed to allow storing the error message. We could use that to report an error message in our application when we use the C API. The current code now prints the following error message in the @ImageMagick code when the encoder could not be found:

`magick: Success `test.avif' @ error/heic.c/IsHeifSuccess/134.`

This is happening because the error has `kSuccess` as the error message instead of the correct error message.

This PR is split into multiple commits and also removes the unused `context` field from `heif_encoder`. If you would prefer to preserve this I can revert the commit but it looks like this is not used in the public API so it should be possible to remove this now.

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -1315,7 +1315,7 @@ struct heif_error heif_context_get_encoder(struct heif_context* context,
     return err.error_struct(context ? context->context.get() : nullptr);
   }
 
-  *encoder = new struct heif_encoder(nullptr, descriptor->plugin);
+  *encoder = new struct heif_encoder(descriptor->plugin);
   (*encoder)->alloc();
 
   struct heif_error err = {heif_error_Ok, heif_suberror_Unspecified, kSuccess};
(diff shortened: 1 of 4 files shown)


farindk:
Thanks. Generally agree. However, we should still allow the `context` parameter to be NULL because we explicitly allowed that in the old documentation.

I have changed this, but for some annoying authentication reason, I cannot push to this PR branch. I've attached the patch here. Maybe you can cross-check it and add it to the PR branch.

[0001-As-we-explicitly-allowed-context-to-be-NULL-previous.patch.txt](https://github.com/strukturag/libheif/files/5117734/0001-As-we-explicitly-allowed-context-to-be-NULL-previous.patch.txt)

dlemstra:
Fair point. I have added your patch and pushed it.

farindk:
Thank you!

PR #294, opened 2020-08-24
CI: Migrate to Github Actions

fancycode:


fancycode:
Should cover the same scenarios as the old Travis configuration.

Still using the generic scripts to install / prepare / run CI jobs, this will be simplified in a later step.

@farindk Please take a look and feel free to merge once everything is green. I noticed that the fuzzer jobs sometimes fail with an undefined behaviour error (e.g. https://github.com/strukturag/libheif/runs/1021765408?check_suite_focus=true#step:5:1974).

PR #302, opened 2020-08-30
Added mingw.mutex.h

Jamaika1:


fancycode:
Why is this needed? With this PR, the mingw CI tests are failing.

Jamaika1:
This is needed when creating libheif with Windows. MINGW initiation can be created in third_party. Of course, you can only add // NOLINT to potentially changed files
```
# std::thread support when cross-compiling with mingw.
# libheif uses <mutex> and therefore needs mingw-std-threads when compiling
# with mingw.
if(MINGW)
option(MINGW_STDTHREADS_GENERATE_STDHEADERS "" ON)
if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/mingw-std-threads/mingw.thread.h")
  message(SEND_ERROR "Please run git submodule update --init")
endif ()
add_subdirectory(mingw-std-threads) 
```

PR #303, opened 2020-08-30
New defined _MSC_VER for GCC

Jamaika1:


farindk:
Could you give a short description why this is needed? Our CI also tests with mingw and that seems to compile also without this patch.

Jamaika1:
I have error with gcc 11.0. _MSC_VER should be less than 1900
```
heif.cc: In function 'void heif_context_debug_dump_boxes_to_file(heif_context*, int)':
heif.cc:330:18: error: 'write' was not declared in this scope; did you mean 'fwrite'?
  330 |   auto written = write(fd, dump.c_str(), dump.size());
      |                  ^~~~~
      |                  fwrite
```

farindk:
superseded by #320

Issue #288, opened 2020-08-14
nclx profile is not passed through the color conversion


farindk:
We have to pass the NCLX profile through the color conversion so that the encoder plugins can embed the information in the image bitstreams.
See the comments from @wantehchang in #279.

novomesk:
Dirk, do you have some news how this could be solved?

If it is too complicated to change library structure to pass the information automatically, I think there could be an alternative solution: What about adding new parameters to AV1 encoders: full_range, primaries, transfer_characteristics, matrix_coefficients ? (with default values 1-full_range,2-unspecified primaries,2-unspecified trc, 6-BT_601)

Setting those values manually introduce small redundancy on library user's side but only if someone will need to use non-default values. It will solve also problem with saving AVIF with ICC profile, and matrix coefficients will be saved in AV1 stream too. Output file will be color compatible with other AVIF implementations.

farindk:
The information probably can be passed automatically. That's something for the next release, as I just didn't want to delay v1.8.0 any further.
I think the library has to handle this automatically. At least there is also the decoding case where the nclx comes from the AV1 bitstreams and the ICC from the container.

wantehchang:
I just proposed pull request #308 as an interim solution to this issue in the aom encoder. Please take a look. Thank you!

novomesk:
Hard coded full range is better than limited range.

farindk:
The nclx is now preserved through color conversion and passed to the aom/rav1e encoders.
If there is an ICC in the input, it will be stored in the `colr` box of the output. Otherwise the nclx is stored in the `colr` box.

Please check whether it is ok now in current master branch (5221af2).

novomesk:
I think there is a new bug now. Tried to save AVIF with ICC profile but avifdec did not find ICC. heif-info reports `color profile: nclx`

farindk:
@novomesk Are you sure that's the current master version? Sounds like the broken behavior in v1.8.0.
This is what I get now:
````
./heif-enc ~/.../jpeg/with-color-profile/YellowFlower-P3.jpg -o out.avif -A

./heif-info out.avif 
MIME type: image/avif
image: 600x400 (id=1), primary
  color profile: prof
  alpha channel: no
  depth channel: no
````

farindk:
With v1.8.0 it was as you describe:
````
./heif-enc ~/.../jpeg/with-color-profile/YellowFlower-P3.jpg -o out.avif -A

./heif-info out.avif 
[...]

PR #308, opened 2020-09-05
Set color_range to full range (1) in aom

wantehchang:
Until issue #288 is fixed, we have to live with using a hardcoded value
of color_range in the aom encoder. It is more common for images to be
full range than limited range. So set color_range to full range (1) in
aom.

### DIFF
--- a/libheif/heif_encoder_aom.cc
+++ b/libheif/heif_encoder_aom.cc
@@ -665,6 +665,10 @@ struct heif_error aom_encode_image(void* encoder_raw, const struct heif_image* i
     aom_codec_control(&encoder->codec, AV1E_SET_ROW_MT, 1);
 #endif
   }
+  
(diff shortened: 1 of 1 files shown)


farindk:
Thanks. I will add the proper forwarding of the nclx profile to aom this week.

wantehchang:
Dirk: Thank you very much! I look forward to your proper patch.

Please do not apply the nclx profile to the alpha auxiliary image. The alpha auxiliary image should always be full range. The colour primaries, transfer characteristics, and matrix coefficients of the alpha auxiliary image can be left unspecified because they are ignored by the decoder. The colr box can be omitted for the alpha auxiliary image. See the recent proposed change to the AVIF spec regarding this: https://github.com/AOMediaCodec/av1-avif/commit/a2180f8889960afb20dc37975eb258b4835686b3

farindk:
Thanks to the pointer to the revised spec. I'll take care of this.

PR #312, opened 2020-09-08
fix the include directory for the pkg-config file

Adsun701:
This corrects the include directory listed in the pkg-config file from @includedir@ to @includedir@/libheif. This allows the libheif headers to be included properly.

### DIFF
--- a/libheif.pc.in
+++ b/libheif.pc.in
@@ -1,7 +1,7 @@
 prefix=@prefix@
 exec_prefix=@exec_prefix@
 libdir=@libdir@
-includedir=@includedir@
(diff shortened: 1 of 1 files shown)


farindk:
I'm not sure we want it this way. Currently people are used to include the files as `<libheif/heif.h>`. If we would change this, all include statements in third-party code need to be changed.

fancycode:
We should not do this, this will break all existing users of libheif (see the failed Go tests for an example here).

farindk:
Sorry, Andrew. Just use `<libheif/heif.h>` in your program.

Adsun701:
@farindk Thanks for informing me.

PR #318, opened 2020-09-11
Leave color description in AV1 alpha unspecified

wantehchang:
For AV1, do not call alpha_image->set_color_profile_nclx() so that color
description in alpha images is left unspecified. This saves three bytes
in the sequence header OBU of alpha images. See Section 5.5.2 (Color
config syntax) in the AV1 spec:

    color_config( ) {
        ...
        color_description_present_flag                  f(1)
        if ( color_description_present_flag ) {
            color_primaries                             f(8)
            transfer_characteristics                    f(8)
            matrix_coefficients                         f(8)
        } else {
            color_primaries = CP_UNSPECIFIED
            transfer_characteristics = TC_UNSPECIFIED
            matrix_coefficients = MC_UNSPECIFIED
        }
        ...
    }

This PR is related to https://github.com/strukturag/libheif/issues/288.

### DIFF
--- a/libheif/heif_context.cc
+++ b/libheif/heif_context.cc
@@ -1625,11 +1625,6 @@ create_alpha_image_from_image_alpha_channel(const std::shared_ptr<HeifPixelImage
   alpha_image->fill_new_plane(heif_channel_Cb, half_range, chroma_width, chroma_height, bpp);
   alpha_image->fill_new_plane(heif_channel_Cr, half_range, chroma_width, chroma_height, bpp);
 
-  auto nclx = std::make_shared<color_profile_nclx>();
-  nclx->set_default();
-  nclx->set_full_range_flag(true); // in default, but just to be sure in case defaults change
-  alpha_image->set_color_profile_nclx(nclx);
-
   return alpha_image;
 }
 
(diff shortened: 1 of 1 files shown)


libheif/heif_context.cc
```diff
@@ -1847,6 +1842,10 @@ Error HeifContext::Image::encode_image_as_hevc(std::shared_ptr<HeifPixelImage> i
 
     std::shared_ptr<HeifPixelImage> alpha_image;
     alpha_image = create_alpha_image_from_image_alpha_channel(image);
+    auto nclx = std::make_shared<color_profile_nclx>();
+    nclx->set_default();
+    nclx->set_full_range_flag(true); // in default, but just to be sure in case defaults change
+    alpha_image->set_color_profile_nclx(nclx);
```
wantehchang, line 1848: I am not familiar with HEIC, so I preserved the current behavior in the encode_image_as_hevc() method and only changed the behavior in the encode_image_as_av1() method.

Please feel free to edit this pull request or replace it with your own as you see fit. Thank you!

farindk:
Thanks. I have implemented this a bit differently by explicitly saving the color information only when it is a normal image (not alpha or depth). That still keeps the nclx with the `full_range` flag in the image.

42c7e08

wantehchang:
Dirk: Your solution also works. Thank you for fixing this! I guess you prefer that alpha images have a color_profile_nclx object even though it doesn't seem to be used?

It seems that the color_profile_nclx objects of alpha images could be set to 2/2/2/true (as opposed to the default 2/2/6/true) to make it clear that the matrix coefficients of alpha images are ignored.

farindk:
Yes, I'd like to leave the nclx in the alpha image, just for the full_range flag. I understand that this should be always `true` for AVIF, but the alpha image might also go through the color conversion pipeline and one never knows what happens there. Also not sure if `full_range=false` might be used in HEIF.

You are right with the 2/2/2 defaults. I'll change that.

PR #321, opened 2020-09-11
Merge nested if statements in x265_encode_image()

wantehchang:
This is equivalent to the original code and matches the corresponding if
statements in aom_encode_image() and rav1e_encode_image().


farindk:
Yes, looks simpler :-)

PR #322, opened 2020-09-11
Omit set_default call on new color_profile_nclx

wantehchang:
No need to call set_default() on newly constructed color_profile_nclx
objects because the color_profile_nclx constructor calls set_default().

### DIFF
--- a/libheif/heif_context.cc
+++ b/libheif/heif_context.cc
@@ -1626,7 +1626,6 @@ create_alpha_image_from_image_alpha_channel(const std::shared_ptr<HeifPixelImage
   alpha_image->fill_new_plane(heif_channel_Cr, half_range, chroma_width, chroma_height, bpp);
 
   auto nclx = std::make_shared<color_profile_nclx>();
-  nclx->set_default();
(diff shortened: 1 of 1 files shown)


libheif/heif_context.cc
```diff
@@ -1626,7 +1626,6 @@ create_alpha_image_from_image_alpha_channel(const std::shared_ptr<HeifPixelImage
   alpha_image->fill_new_plane(heif_channel_Cr, half_range, chroma_width, chroma_height, bpp);
 
   auto nclx = std::make_shared<color_profile_nclx>();
-  nclx->set_default();
```
farindk, line 1629: Could you please rebase this PR to the current master?
I changed the set_default() at this line to set_undefined() because of your comment in (#318).

wantehchang: OK. I will need to create a new PR because I am creating PRs using the Web interface.

wantehchang: By the way, I think "set_unspecified" would seem more accurate than "set_undefined". Also, there are four white spaces in the line after set_undefined() in libheif/box.h.

PR #323, opened 2020-09-11
Fix -Wpotentially-evaluated-expression warnings

wantehchang:
Fix clang -Wpotentially-evaluated-expression warnings on typeid()
arguments that contain function calls. Here is an example of the warning
message:

  heif_colorconversion.cc:2952:33: warning: expression with side effects
  will be evaluated despite being used as an operand to
        'typeid' [-Wpotentially-evaluated-expression]
      std::cerr << "> " << typeid(*op).name() << "\n";
                                  ^

The actual warning fix is to pass to typeid() an expression that does
not call any function. Also modify related code to maintain a consistent
style.

### DIFF
--- a/libheif/heif_colorconversion.cc
+++ b/libheif/heif_colorconversion.cc
@@ -2885,13 +2885,14 @@ bool ColorConversionPipeline::construct_pipeline(ColorState input_state,
 
     // expand the node with minimum cost
 
-    for (size_t i = 0; i < ops.size(); i++) {
+    for (const auto& op_ptr : ops) {
 
 #if DEBUG_PIPELINE_CREATION
-      std::cerr << "-- apply op: " << typeid(*(ops[i])).name() << "\n";
+      auto& op = *op_ptr;
+      std::cerr << "-- apply op: " << typeid(op).name() << "\n";
 #endif
 
-      auto out_states = ops[i]->state_after_conversion(processed_states.back().color_state.color_state,
+      auto out_states = op_ptr->state_after_conversion(processed_states.back().color_state.color_state,
                                                        target_state,
                                                        options);
       for (const auto& out_state : out_states) {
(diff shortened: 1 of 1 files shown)


libheif/heif_colorconversion.cc
```diff
@@ -2885,13 +2885,14 @@ bool ColorConversionPipeline::construct_pipeline(ColorState input_state,
 
     // expand the node with minimum cost
 
-    for (size_t i = 0; i < ops.size(); i++) {
+    for (const auto& op_ptr : ops) {
 
 #if DEBUG_PIPELINE_CREATION
-      std::cerr << "-- apply op: " << typeid(*(ops[i])).name() << "\n";
+      auto& op = *op_ptr;
+      std::cerr << "-- apply op: " << typeid(op).name() << "\n";
```
wantehchang, line 2892: This is the first of the two warning fixes in this CL. This warning is emitted only when DEBUG_PIPELINE_CREATION is defined as 1.

libheif/heif_colorconversion.cc
```diff
@@ -2948,8 +2949,9 @@ bool ColorConversionPipeline::construct_pipeline(ColorState input_state,
 
 void ColorConversionPipeline::debug_dump_pipeline() const
 {
-  for (const auto& op : m_operations) {
-    std::cerr << "> " << typeid(*op).name() << "\n";
+  for (const auto& op_ptr : m_operations) {
+    auto& op = *op_ptr;
+    std::cerr << "> " << typeid(op).name() << "\n";
```
wantehchang, line 2954: This is the second of the two warning fixes in this PR.

All the other changes in this PR are intended to maintain a consistent style in the code that operates on 'ops' or m_operations. Please feel free to implement an alternative fix to suit your preference. Thank you!

farindk:
Great. Thanks.

wantehchang:
Dirk: It would be good if you could test the new code. I wrote the PR carefully by code transformation, so the new code should be equivalent to the original code. But one never knows.

There is one thing I don't understand. The argument we pass to typeid() is an expression of the ColorConversionOperation& type. Does typeid() return the type info of ColorConversionOperation or the actual subclass of ColorConversionOperation dynamically?

farindk:
Yes, I have tested it and it works fine.

typeid() returns the type info of the actual class, not that of the common base class. It gets this information from the virtual function table of that object. In libheif, this is only used for debugging output, showing the list of color transformations that are performed and the planning algorithm that determines the order of transformations.

PR #327, opened 2020-09-14
pkgconfig: add builtin_rav1e_encoder variable

lovell:
Sets this by exposing `have_rav1e` from both autotools and CMake, plus also exposes the missing `have_aom` variable from CMake.

As previously discussed in #222


farindk:
I am wondering whether we really should expose the details which encoders are used internally. Isn't that unnecessary and just complicating things for the users of the library?
How about setting `builtin_avif_encoder` to `true` if either `have_aom` or `have_rav1e` is `true`?

lovell:
Thanks, I agree, will update this PR accordingly.

lovell:
Now updated to ensure both aom and rav1e set relevant values for the existing `builtin_avif_decoder` and `builtin_avif_encoder` variables.

farindk:
Great, thanks !

PR #334, opened 2020-09-18
rav1e: ensure linker knows of dl dependency

lovell:
The next version of rav1e (v0.4.0) will require this linker flag. Adding it prevents the following linker errors:
```
/usr/bin/ld: ../libheif/.libs/libheif.so: undefined reference to `dladdr'
/usr/bin/ld: ../libheif/.libs/libheif.so: undefined reference to `dlsym'
```
Linking against the current v0.3.4 of rav1e works with and without this change so it's both backwards and forwards compatible.


farindk:
Thank you.

PR #337, opened 2020-09-24
build: ensure HEVC-related plugins are optional

lovell:
This allows the optional, explicit removal of the patent encumbered libde265 and/or x265 dependencies from the build.

Adds two new optional autotools flags, default "no":
* `--disable-libde265`
* `--disable-x265`

Adds two new optional CMake options, default ON:
* `-DWITH_LIBDE265=OFF`
* `-DWITH_X265=OFF`


farindk:
Yes, makes sense, now that there is AVIF as a HEIF alternative.

PR #341, opened 2020-09-25
Use "pkg-config" to detect if dav1d is available.

fancycode:
This will help in adding support for dav1d when packaging.

Updated the script to build custom version in "third-party" folder so it can be used through "pkg-config" instead of relying on hardcoded folder names.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         env:
           - { NAME: "nothing" }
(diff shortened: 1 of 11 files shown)


fancycode:
@farindk this is now ready for review. It would be great to have an `.avif` file in the `examples` folder so CI can check if decoding works (similar to what we are already doing with the `example.heic`).

farindk:
> @farindk this is now ready for review. It would be great to have an `.avif` file in the `examples` folder so CI can check if decoding works (similar to what we are already doing with the `example.heic`).

I have added the 'example.avif' image.

PR #342, opened 2020-09-27
Fix building on anything but Linux

brad0:
The addition of the alloca.h header to heif_convert.cc in commit 20bb92009fc95b95ba598b6485cd35670d79eb32#diff-4fc948cb4190594d731047b3335eed6d broke building on anything but Linux.


brad0:
and macOS too from the looks of it, since it supports both headers. Anyway, alloca.h is very much not portable. I also noticed after this that heif_info.cc uses alloca() but does not have the header, for Linux.

farindk:
I have now removed all usages of alloca() since it was causing endless trouble...

Issue #335, opened 2020-09-21
online JavaScript decoder shows corrupted image when image width is odd


farindk:
The decoder https://strukturag.github.io/libheif/ shows a corrupted image when the image width is odd. This is not due to libheif itself, as the image decodes fine even with v1.0.0. I guess it is a problem in the JS canvas drawing code.
See also #331 for a screenshot and the demo image.

catdad:
Starting the investigation, the image looks like this:

![image](https://user-images.githubusercontent.com/2205537/94495287-59afb200-01bf-11eb-9daf-f99b263b9249.png)

We are both off-by-one for loading pixels into the canvas data, and we also have the color all wrong.

For the color, I had a guess. There is some division in the `display` code here: https://github.com/strukturag/libheif/blob/a1051cf9be37889a0684ba8eafed215596ef70c0/post.js#L83-L84

When the width is odd, we get a decimal value. We would need to do `Math.floor(w / 2)`. Doing this gets us the correct colors:

![image](https://user-images.githubusercontent.com/2205537/94495243-38e75c80-01bf-11eb-85b5-e10c3c36949f.png)

For the off-by-one pixel while loading data into the canvas, I have not yet figured out the problem. I can tell, however, that if I create the canvas image data like this `const imageData = context.createImageData(width - 1, height);` the final image is drawn correctly. That at least means the off-by-one issue causes us to load 1 less pixel than expected.

If I find the pixel issue as well, I will submit a PR

catdad:
I apologize if this is too much noise.

Looking further into the algorithm for filling the canvas data with pixels, it looks like it fills 2 adjacent pixels at a time. When it gets to the end of the line, it has filled `width - 1` pixels. Then on the next iteration of the loop, the 2 pixels of data that it reads from the original image are the first 2 pixels on the following line. It seems that not only is it filling 1 less pixel per line, but it actually does not have data for that pixel in the first place? Does this sound right? Maybe I am missing something.

Either way, I submitted a PR with just the color fix, as a user can work around the missing pixel issue by passing in smaller image data and then cropping (but a user cannot work around the color issue).

I am unsure of how to test it but am happy to add tests if someone can point me in the right direction.

PR #345, opened 2020-09-28
fix issue with poor color rendering for images that have an odd width for javascript version

catdad:
This addresses half of #335. See my comment here: https://github.com/strukturag/libheif/issues/335#issuecomment-700333672


fancycode:
Continued in #369, thanks for the first part of the fix.

PR #349, opened 2020-09-29
Detect encoding/decoding support in AOM + option to disable AOM

DarthSim:
Hi everyone!

I have libaom built with decoding support but without encoding support. With this setup, libheif can't be compiled since it can't find `aom_encoder.h`. It would be useful to check aom's encoder and decoder separately or even disable aom if I decide to switch to dav1d. This PR adds both features.


fancycode:
Thanks!

PR #355, opened 2020-10-04
fix alpha encoding when using grid workaround

itaibenari:
when using the grid workaround the alpha `auxl` item refers to the image before it was wrapped in a grid (`image_id`) and causes alpha to be lost. so I changed the order and build the alpha layer after the main image.

additionally, for small images (64px), there's an assertion that breaks at `heif_encoder_x265.cc:769`, hence changed the alpha layer to be monochrome and only use Y.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -2413,8 +2413,11 @@ void Box_clap::set(uint32_t clap_width, uint32_t clap_height,
   m_clean_aperture_width = Fraction(clap_width, 1);
   m_clean_aperture_height = Fraction(clap_height, 1);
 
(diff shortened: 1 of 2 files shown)


farindk:
@itaibenari I think your first patch 8e930c2 is fine (if the alpha is encoded as monochrome and hence can have odd size). The second patch (5866f1a) is problematic because it appears to disable the whole 'clap' processing. This would break any correct codec just to match the error is macOS.

Do we have any example case why we need to modify the 'clap' processing?

farindk:
Seems to be obsolet now.

Issue #350, opened 2020-09-29
realtime parameter in AOM encoder produced incomplete .avif file


novomesk:
Hello,
I was trying to encode AVIF with realtime speed:
`heif-enc -A -p speed=6 -p realtime=1 -o z.avif z.png`

the produced z.avif was only 283 bytes, it is obviously incomplete.

heif-info indicate:
```
heif-info z.avif
MIME type: image/avif
Could not read HEIF/AVIF file: Invalid input: Item has no data: Item with ID 1 has no compressed data
```

novomesk:
@wantehchang If you have time, can you please take a look why AOM_USAGE_REALTIME in libheif/AOM encoder doesn't work?

Could it be potentially related to https://bugs.chromium.org/p/aomedia/issues/detail?id=2723 ?

wantehchang:
Daniel: I can't reproduce this bug. I use the Debian package libaom-dev:amd64 (1.0.0.errata1-3). I tried both a JPEG and a PNG file as input.

https://bugs.chromium.org/p/aomedia/issues/detail?id=2723 is a crash due to segmentation fault. Did your heif-enc command crash with a segmentation fault?

novomesk:
I used libaom 2.0.0 and 2.0.0-918-g75a47cc18
There was no crash, no error printed, but produced output file is incomplete.

wantehchang:
Daniel: Thank you. I reproduced the bug and wrote a fix:
https://github.com/strukturag/libheif/pull/361

PR #361, opened 2020-10-15
Call aom_codec_get_cx_data() after encode_frame()

wantehchang:
We need to call aom_codec_get_cx_data() after each aom_codec_encode()
call. In real-time mode, the output packet is generated after the first
aom_codec_encode() call. Apparently that output packet is deleted when
the second aom_codec_encode() call (with the NULL second argument) is
invoked.

Fix https://github.com/strukturag/libheif/issues/350.


farindk:
Thank you for this.

PR #362, opened 2020-10-16
Set still_picture to 1 in AV1 sequence header OBU

wantehchang:
Set cfg.g_limit to 1 to make the libaom encoder set still_picture and
reduced_still_picture_header to 1 in the AV1 sequence header OBU.

Empirically this saves six bytes in image size.


farindk:
Thanks. I did not find the information how to switch on the still_picture_header...

wantehchang:
> Thanks. I did not find the information how to switch on the still_picture_header...

I also had to analyze the libaom source code to figure this out. I have an open bug to document this officially:
https://crbug.com/aomedia/2733

PR #363, opened 2020-10-18
Fix heif_image_get_primary_height for interleaved RGB images

0xC0000054:
The image width was being returned instead of the height.


fancycode:
Thanks for spotting this.

PR #367, opened 2020-10-31
Ensure box clap values are sanitised when parsing

lovell:
Helps prevent possible integer overflows and floating point exceptions that will crash the process.

Discovered via fuzz testing - I can (privately) provide test cases to trigger these, if required.


fancycode:
Thanks, feel free to add a testcase to the `fuzzing/corpus` folder in the repository so it gets picked up by our and thethe oss-fuzz runners.

lovell:
@fancycode Thank you, I've added a couple of test cases to this PR.

fancycode:
Perfekt, thank you!

PR #369, opened 2020-11-02
Fix JavaScript rendering for images with odd widths

### LINKED ISSUE #335: online JavaScript decoder shows corrupted image when image width is odd

fancycode:
Fixes #335 and obsoletes #345.

@catdad can you confirm this fixes the issue?


catdad:
Overall, the images look correct now, and this is a great improvement.

However, I still see something curious. If I take the resulting image and inspect the bottom closely, I see this (zoomed in in Photoshop):

image
image

While some parts of the bottom line of pixels on the left side are not completely black, they are very very dark. The right side has two rows of missing pixels, and they do appear to be completely black.

I also see this along the right edge:

image

It shows discoloration on the entire right column of pixels.

For reference, here is the entire resulting jpeg:

download

With other converters I have available, like the built-in Windows Photo converter, I do not see these discrepancies.

fancycode:
Thanks for the additional feedback, the missing pixels at the bottom are fixed in https://github.com/strukturag/libheif/pull/369/commits/3a023f9dd1aea629a12bcfe4e518bf53cb88e38f

I already have an idea about the wrong colors on the right side but that needs more changes.

fancycode:
Right side fixed with https://github.com/strukturag/libheif/pull/369/commits/d97b79f3c1657ba1497e9a549a0f0e6bb9547d11

catdad:
Hmm... something about the version at https://github.com/strukturag/libheif/commit/d97b79f3c1657ba1497e9a549a0f0e6bb9547d11 didn't work quite right. Here is the same image:

image

fancycode:
Thanks for testing and providing feedback!

PR #370, opened 2020-11-06
Use PROJECT_SOURCE_DIR for CMAKE_MODULE_PATH

theirix:
This patch allows using libheif included in another CMake project.


farindk:
Thank you.

PR #375, opened 2020-11-12
Fix clap parse

FrankGalligan:
The reading of the clap values were reversed in the Fraction constructor.


FrankGalligan:
I'm not sure the errors are from this PR.

farindk:
Yes, right. Having two read32() calls in the function parameters has undefined ordering.

PR #385, opened 2020-11-24
Improve aom encode error reporting

kleisauke:
This PR improves the error reporting of encoding with aom and thereby resolves assertion failures when encoding large images on a Raspberry Pi 4B (with only 4GB ram available).

To summarize, ~the `aom_log_error` function (similar to the [`die_codec`](https://aomedia.googlesource.com/aom/+/6514dccc195aa95d45fe156fac77ff57627975ff/common/tools_common.c#71) example) was introduced for logging aom encoding errors and~ (potentially) reachable `assert(false)` calls are removed in favor of returning appropriate errors.

### DIFF
--- a/libheif/heif_encoder_aom.cc
+++ b/libheif/heif_encoder_aom.cc
@@ -28,15 +28,11 @@
 #endif
 
 #include <cstring>
-#include <cstdio>
 #include <cassert>
 #include <vector>
 
 #include <aom/aom_encoder.h>
 #include <aom/aomcx.h>
-
-
-#include <iostream>  // TODO: remove me
 
 
 struct encoder_struct_aom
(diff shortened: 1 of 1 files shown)


fancycode:
Thanks. Imho we should not output anything if the errors are handled / returned correctly now (instead of being `assert`ed). @farindk what do you think?

farindk:
@kleisauke Thank you. Yes, the `printf` should be removed. Moreover, while we are at this, I think it would be even better when we would forward the AOM errors. We can use the encoder struct for storing a copy of the AOM error string. I can have a look at this.

kleisauke:
OK, I've removed the `printf` calls (and the related `aom_log_error` function). This PR now only focuses on returning the appropriate errors instead of `assert`-ing.

Feel free to implement logic to forward aom encoding errors. Although, I'm not sure whether it contains any useful details, usually when there is a encoding error in an aom, it is mainly due to OOM-issues.

farindk:
Thank you again.

PR #392, opened 2020-11-30
Added libjpeg turbo version

Jamaika1:


farindk:
Thanks, but I don't think this is correct. libjpeg-turbo's version number is defined like this:

````
/* libjpeg-turbo version in integer form */
#define LIBJPEG_TURBO_VERSION_NUMBER  2000091
````

Hence, we get the major version number by dividing by 1000000, not >>16, for example.
Moreover, the code only checks for major/minor parts, but not the patch number, even though the bug is present in 2.0 and was fixed in 2.0.1.

Jamaika1:
I am indeed a poor beginner programmer, but I don't give up. Shake what author came up with a new one.
For example for WebPX codecs I have WP2GetVersion () as 0x020507 equivalent to "2.5.7"
Here they came up with 002005007. I don't know what for?
The int value is displayed as shown below. I will correct it.

PR #393, opened 2020-11-30
Show version library

Jamaika1:


fancycode:
The various defines are only available if the corresponding library is included (see the various CI build errors).

PR #395, opened 2020-12-01
Added libjpeg turbo version

Jamaika1:


farindk:
@Jamaika1 . Thanks.
I have squashed your commits into two. First the check for libjpeg-turbo v2.0 (de6b8d2f3aa8a3acda7d0d48eb084af9a2aaaf30) and the ICC check based on the libjpeg version number (517c549cb8f8d99a1a456f9cd0e73144a18af037).

I have cherry-picked the first commit into master (because it tests for a specific broken version in libjpeg-turbo).

I did not merge the ICC check, because I think that the current method of explicitly testing for existence of the ICC-write functions is better. There are more JPEG encoder implementations (e.g. MozJPEG) and we might get into problems when we check the version numbers explicitly.

PR #396, opened 2020-12-02
Show version library

Jamaika1:


farindk:
This only works when libjpeg is libjpeg-turbo.

farindk:
Thanks @Jamaika1 , but the problem is that it would only output the version used during compilation as you are showing the #defines. The currently installed version on the user PC might be different when it is linked from a shared library.
That's why libheif has a function `heif_get_version()` to query the actual version.

Issue #398, opened 2020-12-09
Cannot load from non-ANSI paths on Windows


tomalakgeretkal:
`HeifFile::read_from_file` is going to need to pass UTF-16 in a `wchar_t*` into Microsoft's non-standard `ifstream` constructor, otherwise all you can use is ANSI paths.

I propose we treat the input `const char*` as UTF-8 and add a conversion.

The whole thing can be wrapped in a macro so as not to affect other platforms at all.

farindk:
Fixed in #399.

PR #399, opened 2020-12-09
Fixed issue #398 (Cannot load from non-ANSI paths on Windows)

tomalakgeretkal:


silverbacknet:
Only nit: Instead of `_MSC_VER`, I think it should be `_WIN32 || _WIN64`, so that it also works on gcc and clang.

tomalakgeretkal:
It _shouldn't_ be used on GCC or Clang, even on Windows; this is a Visual Studio specific thing. :)

silverbacknet:
Gotcha, I didn't realize it was purely an stdlib implementation difference.

tomalakgeretkal:
Fixed a couple of formatting niggles.

tomalakgeretkal:
Swapped out `#undef min` for `#define NOMINMAX`, which is better.

libheif/heif_file.cc
```diff
@@ -27,6 +27,11 @@
 #include <cstring>
 #include <cassert>
 
+#ifdef _MSC_VER
+#define NOMINMAX
```
Jamaika1, line 31: heif_file.cc:31: warning: "NOMINMAX" redefined
   31 | #define NOMINMAX
      |
In file included from c:\msys1021\include\c++\10.2.1\x86_64-w64-mingw32\bits\c++config.h:522,
                 from c:\msys1021\include\c++\10.2.1\cstdint:38,
                 from c:\msys1021\include\c++\10.2.1\cinttypes:38,
                 from box.h:28,
                 from heif_file.h:28,
                 from heif_file.cc:21:
c:\msys1021\include\c++\10.2.1\x86_64-w64-mingw32\bits\os_defines.h:45: note: this is the location of the previous definition
   45 | #define NOMINMAX 1

tomalakgeretkal: Thanks! Easy enough to fix, but I'm concerned that MinGW is defining `_MSC_VER`. It shouldn't be doing that. Do you get an error in `HeifFile::read_from_file` too?

tomalakgeretkal: Also please give your mingw-64 version.

Jamaika1: I am using version 10.2.1 for the x265 codec. Unfortunately, there are no upgrades to new versions of GCC and you should create it yourself.
http://msystem.waw.pl/x265/

tomalakgeretkal: Interestingly, it looks like MinGW does sometimes define `_MSC_VER` as it relates more to the CRT than to Visual Studio itself. Still, that suggests this should all be okay now, as long as you can confirm that `HeifFile::read_from_file` doesn't cause an error for you.

Jamaika1: For MINGW is
`#if (defined(_MSC_VER) && _MSC_VER >= 1900) || defined(__MINGW32__) || defined(__MINGW64__)`
`#if (defined(_MSC_VER) && _MSC_VER < 1900) || !defined(__MINGW32__) || !defined(__MINGW64__)`
I am amateur, not programmer. This isn't a rule and value 1900 depends on the MSC version and thus how much the library is overdue on github and isn't updated.

tomalakgeretkal: That's for detecting Visual Studio specifically, not VS CRT compatibility. That's why I asked whether you get an error from `HeifFile::read_from_file`: if you do, then we'll need the full check that you showed; if you don't, then this is already the right code as MinGW is "simulating" Visual Studio for this feature (and its intention to do this in many cases is why it defines `_MSC_VER`).

Jamaika1: i don't get an error from HeifFile::read_from_file.

tomalakgeretkal: Excellent. :) Then I think we're good to go!

(not sure of the relevance of avif here?)

farindk:
Thank you, @tomalakgeretkal . Could you please add one example case here in the comments so that we know how to reproduce the issue?

tomalakgeretkal:
> Thank you, @tomalakgeretkal . Could you please add one example case here in the comments so that we know how to reproduce the issue?

Sure - just have a valid heif file at, say, `c:\folderé\file.heif` and attempt to load it!

TimmvonderMehden:
@tomalakgeretkal : i just had the same error with a path containing the german ü (p:/Münster), which i fixed by renaming the path to p:/Muenster. Irfanview's most recent version fails as well on those files ( i guess they are using libheif) 

I am using libbheif 1.12.0 via heif_context_read_from_file on windows 7 x64 Visual Studio 2019. I did not verify if your fix applies implicitly to this function as well, your fix was at HeifFile::read_from_file(const char* input_filename).

Anyway, i think, you expect the filename to be utf8, but this is not common on windows using std::string ot CString or const char* (QString from QT could work). At least the source code should have a comment about utf8 filenames.

please tell me if i shall open a new issue!

i guess, heif_context_write_to_file is affected as well.

silverbacknet:
If you're using std::string or CString or const char* without using UTF-8, then you already know that Unicode will never work. UTF-16 and related wstring/wchar_t/WCHAR storage types are all in various states of deprecation in the Win32 API, better just to ask for filenames in UTF-8 directly if your framework allows it, WideCharToMultiByte(CP_UTF8,...) / wcstombs(utf8,...) if not, or just pay attention and not assume you can completely ignore wide/narrow conversions.

Issue #365, opened 2020-10-23
<16 pixels encoding issue with aom


kleisauke:
Since commit https://github.com/strukturag/libheif/commit/b22820a276d7597f7f633bc57c34c28212701cdb AVIF images smaller than <16 pixels are encoded to 16×16 with cropping. This works great with rav1e, but for some reason it cannot be opened in Chrome when encoding with aom. libvips also seems to fail in processing this image (issue https://github.com/libvips/libvips/issues/1808 might be relevant here).

Test images:
https://t0.nl/x-aom.avif - 10×10 encoded with aom v2.0.0
https://t0.nl/x-rav1e.avif - 10×10 encoded with rav1e v0.3.4

For reference, the images encoded as 16×16:
https://t0.nl/x2-aom.avif
https://t0.nl/x2-rav1e.avif

The test images are generated with:
```bash
# aom v2.0.0
vips crop zebra.jpg x-aom.avif[compression=av1] 0 0 10 10
vips crop zebra.jpg x2-aom.avif[compression=av1] 0 0 16 16

# rav1e v0.3.4
vips crop zebra.jpg x-rav1e.avif[compression=av1] 0 0 10 10
vips crop zebra.jpg x2-rav1e.avif[compression=av1] 0 0 16 16
```

Building a debug build of Chromium reveals this verbose message:
```bash
[29580:29588:1023/105621.829690:VERBOSE1:avif_image_decoder.cc(534)] Frame size "10x10" differs from container size "16x16"
```
https://github.com/chromium/chromium/blob/d3f2d5be6762b18b922d109a19250812ea88f9e0/third_party/blink/renderer/platform/image-decoders/avif/avif_image_decoder.cc#L534

0xC0000054:
The issue is that the AOM encoder does not extend the image size to 16x16 before encoding it.
Chrome requires that all images match the size that is declared in the AVIF container.

kleisauke:
Curiously, the actually encoded image of `x-aom.avif` is 7×8 in size, see:
```bash
$ heif-info x-aom.avif
MIME type: image/avif
image: 10x10 (id=1), primary
  color profile: nclx
  alpha channel: no
  depth channel: no
$ heif-convert x-aom.avif x.jpg
File contains 1 images
Written to x.jpg
$ vipsheader x.jpg
x.jpg: 7x8 uchar, 3 bands, srgb, jpegload
```
[...]

PR #402, opened 2020-12-15
Update plane dimensions in extend_to_size

kleisauke:
Fixes: #365.


farindk:
@kleisauke Thank you for spotting this bug.

PR #354, opened 2020-10-03
pkg-config: add Requires.private field to allow static builds

lovell:
This ensures the linker is made aware of symbols from transitive dependencies.

PR #408, opened 2020-12-23
pkg-config: add missing Requires.private field for dav1d/rav1e

kleisauke:
The pull requests containing pkg-config support for dav1d/rav1e landed at about the same time as https://github.com/strukturag/libheif/pull/354. This PR ensures that these dependencies are also included in the `Requires.private` field.


farindk:
Thank you!

PR #415, opened 2021-01-04
Fix heif_cxx get_metadata with 0-size metadata

hjmallon:
Some Apple HEIC files seem to have 0-size Exif data which causes the cxx wrapper to throw an error. The equivalent code in the examples dir checks for 0-size, so we should here too I think.

See https://github.com/OpenImageIO/oiio/pull/2809 and https://github.com/OpenImageIO/oiio/issues/2794

### DIFF
--- a/libheif/heif_cxx.h
+++ b/libheif/heif_cxx.h
@@ -765,6 +765,9 @@ namespace heif {
                                                            metadata_id);
 
     std::vector<uint8_t> data(data_size);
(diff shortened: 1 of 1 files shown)


farindk:
Thanks. Is that problem also fixed by the commit `1c67755` ?
I would prefer that since it would apply to all language bindings at once.

hjmallon:
Yep that looks like a better fix, thanks! Closing this

PR #419, opened 2021-01-07
Allow reading of 'avis' ftyp/filetype/brand images

lovell:
Hello, libheif already has some support for HEIF images with an `ftyp` of "avis" e.g. in the `heif_main_brand` function:

https://github.com/strukturag/libheif/blob/63be81f6e2800ccfc7eaece277669d677e126c89/libheif/heif.cc#L184-L186

This PR extends that support to allow these images to be decoded (or at least progress to the next step of validation).

An example AVIF image that currently fails but which this PR will allow can be found at
https://colinbendell.github.io/webperf/animated-gif-decode/2.avif

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -120,6 +120,9 @@ heif_filetype_result heif_check_filetype(const uint8_t* data, int len)
     else if (brand == heif_avif) {
       return heif_filetype_yes_supported;
     }
+    else if (brand == heif_avis) {
+      return heif_filetype_yes_supported;
+    }
     else if (brand == heif_unknown_brand) {
       return heif_filetype_no;
     }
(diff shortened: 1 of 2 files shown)


hjmallon:
When I tried this the other day I found that this avis files have no primary image, which didn't fit in with how libheif worked. If that is true and the cause of the 'next step of validation' failure then it seems fair to blacklist the brand as they are not supported?

farindk:
Yes. Let's postpone this until we have proper sequence support. I'm still not sure how important sequence support is, though.

hjmallon:
I assume people have a use for it, and it is part of `HEIF` so it seems like it should be part of `libheif`. Maybe there shoudl be an open issue for 'Support HEIF Image Sequnces (`msf1` / `avis` / etc)' or something to track interest?

lovell:
I would be so very happy for the whole concept of "animated images" to disappear entirely and become the videos they deserve.

Some `avis` images in the wild, such as the example above, do appear to store the first frame in the primary item and can be parsed by libheif.

However I will apologise for kowtowing to these and promise never to speak of this again.

PR #435, opened 2021-01-29
Support both python and python3 for running cpplint.

### LINKED ISSUE #380: The pre-commit hook prevents to commit without a word
When I wanted to make a patch for libheif (#379), I got stuck because `git commit` (and all variants adding files first, or with `-a`, I tested so much…) was immediately returning with code 1 and no output whatsoever. As I just upgraded my machine, I even thought that `git` broke (but I could commit on every other repository).

I finally discovered that the autogen.sh installs a pre-commit hook (`.git/hooks/pre-commit`), which itself calls `./scripts/cpplint.py`. The problem is this cpplint.py. First of all, it looks like it still uses Python 2 (which was not installed on my machine as it's now deprecated and distributions don't ship it by default, or now even at all for some). So it would need to be ported.

At the very least, there should be a warning somewhere. Maybe the pre-commit hook could check for Python 2 existence and write an error if absent. Also cpplint.py shebang should be `#!/usr/bin/env python2` until it's ported (many distributions now symlink `python` to `python3`; `python` should be reserved for scripts both python 2 and 3 compatible).

Ah and by the way, the reason for which `git commit` was silent (which was the most annoying thing, I wasted a lot of time on this issue) was because of:

```
  # Change stderr to write with replacement characters so we don't die 
  # if we try to print something containing non-ASCII characters.
  sys.stderr = codecs.StreamReaderWriter(sys.stderr,
                                         codecs.getreader('utf8'),
                                         codecs.getwriter('utf8'),
                                         'replace')
```

Apparently this is about encoding (cf. comment), but when run with Python 3, it just make stderr disappear, which was very bad, because we cannot see the Python errors when the script crashed.

fancycode:
Fixes #380

### DIFF
--- a/scripts/cpplint.py
+++ b/scripts/cpplint.py
@@ -55,7 +55,7 @@ import sys
 import unicodedata
 
 
-_USAGE = """
+_USAGE = r"""
 Syntax: cpplint.py [--verbose=#] [--output=vs7] [--filter=-x,+y,...]
                    [--counting=total|toplevel|detailed] [--root=subdir]
                    [--linelength=digits]
@@ -188,6 +188,7 @@ _ERROR_CATEGORIES = [
     'build/header_guard',
     'build/include',
     'build/include_alpha',
+    'build/include_directory',
     'build/include_order',
     'build/include_what_you_use',
     'build/namespaces',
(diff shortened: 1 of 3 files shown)


farindk (approved):
Looks good. Works fine on Ubuntu 20.04.

PR #440, opened 2021-02-18
3rd party: Update dav1d to 0.8.2

Jamaika1:


farindk:
Thanks. Has been changed already.

PR #451, opened 2021-02-26
Change speed integer maximum

Jamaika1:
AOM 2.1.0-rc1


adityapatadia:
@farindk can we merge this one? Libaom has now 9 speed modes.

farindk:
I've added a check so that speed=9 is only enabled for AOM 3.0.0 and above.
Seems like v2.1.0-rc didn't make it past an RC, so I did not bother to include this version.

PR #456, opened 2021-02-28
Correction of rav1e include path

1div0:
Please review this small correction of build.


1div0:
Please consider only the very last commit https://github.com/strukturag/libheif/pull/456/commits/1c0552cac8f38e3301be025134b829e0863183ea

farindk:
Needs more changes to CMakeLists.txt. Only changing the include does not work with CMake on Linux with local rav1e build.

PR #470, opened 2021-03-27
Refactor AppVeyor configuration

EwoutH:
I rewrote the AppVeyor configuration file to modernize the build process a bit:
 - Updates environment to the Visual Studio 2019 image (running on Windows Server 2019)
 - Use [vcpkg](https://github.com/microsoft/vcpkg) for installing dependencies, using the latest available versions
 - Caches the vcpkg depenencies for faster consecutive builds
 - Remove x86 job, add arm64 job
 - Publish individual .exe artifacts

The first build will be slow, because all the dependencies will need to be build, but all builds after that will be significantly faster because libde265, x265 and dav1d are cached with vcpkg.

The AppVeyor setup now doesn't depend on any external scripts, so `scripts\prepare-appveyor.cmd` can be deleted or renamed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,10 @@ image: Visual Studio 2019
 configuration: Release
 cache: c:\tools\vcpkg\installed\
 
+matrix:
+  allow_failures:
+    - arch: arm64 # libde265 currently doesn't support arm64 on vcpkg
+
 environment:
   matrix:
   - arch: x64
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,29 +1,30 @@
 # stats available at
 # https://ci.appveyor.com/project/strukturag/libheif
-version: 1.0.{build}
-
-os:
-  - Windows Server 2012 R2
+image: Visual Studio 2019
+configuration: Release
+cache: c:\tools\vcpkg\installed\
 
 environment:
   matrix:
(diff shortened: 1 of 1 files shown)


EwoutH:
As you can see [on my branch](https://ci.appveyor.com/project/EwoutH/libheif/history), a build now takes about 1 minutes and 30 seconds.

Since libde265 currently doesn't [support](https://github.com/microsoft/vcpkg/blob/master/ports/libde265/portfile.cmake#L1) Arm64 targets when building with vcpkg, I allowed those to fail for now. When support is added to libde265 and the vcpkg port, that line can be removed.

EwoutH:
@farindk Could you (assign a) review(er)?

farindk:
Look good to me (but I'm no AppVeyor user). I'll merge after @fancycode had a look.

EwoutH:
Thanks! I've rebased the PR in the meantime.

farindk:
@fancycode Any comments on this? Otherwise I'm going to merge it for v1.12.0.

farindk:
I'll merge this without review. Looks good to me. Thank you @EwoutH for the contribution.

fancycode:
Sorry for the late response, lgtm overall. Now that `scripts\prepare-appveyor.cmd` is no longer required, it should be deleted imho.

PR #474, opened 2021-04-08
WASM output added to Emscripten build process

ravindranrahul:
-s WASM=1 — Specifies that we want wasm output. If we don’t specify this, Emscripten will just output asm.js, as it does by default.

The WASM build will provide significant performance improvement for the dependent JavaScript library: [libheif-js](https://github.com/catdad-experiments/libheif-js)

### DIFF
--- a/build-emscripten.sh
+++ b/build-emscripten.sh
@@ -40,6 +40,7 @@ export TOTAL_MEMORY=16777216
 echo "Running Emscripten..."
 emcc libheif/.libs/libheif.so \
     --bind \
+    -s WASM=1 \
     -s NO_EXIT_RUNTIME=1 \
     -s TOTAL_MEMORY=${TOTAL_MEMORY} \
     -s ALLOW_MEMORY_GROWTH=1 \
--- a/post.js
+++ b/post.js
@@ -1,67 +1,72 @@
-function StringToArrayBuffer(str) {
+(function () {
+  function StringToArrayBuffer(str) {
     var buf = new ArrayBuffer(str.length);
     var bufView = new Uint8Array(buf);
-    for (var i=0, strLen=str.length; i<strLen; i++) {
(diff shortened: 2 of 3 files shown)


farindk:
Why did you close this? Is there any problem with WASM output? Did you compare the WASM speed with JS?

ravindranrahul:
> Why did you close this?

WASAM binary loads asynchronously, I need to figure out how to initialize the post.js code correctly during the build (I have manually moved post.js code to libheif-post.js script now). I also need a onRuntimeInitialized callback in the client script, so i know when the library has loaded. Let me know if you have any comments on this.

> Is there any problem with WASM output

No

> Did you compare the WASM speed with JS

Yes. WASM is faster. You can test it here: https://heif-wasm-ravindranrahul.vercel.app/
 Source: https://github.com/ravindranrahul/heif-wasm

farindk:
Thanks for the insight. I'm looking forward to having WASM support !

ravindranrahul:
I have completed the refactoring of the JavaScript code; summary below:

1. pre.js and post.js was wrapping the entire library in an inner function. This is a bad practice. I have separated them.

> In particular, --pre-js code should not alter the main output from emscripten in ways that could confuse the optimizer, such as using --pre-js + --post-js to put all the output in an inner function scope (see MODULARIZE for that)

Source: https://emscripten.org/docs/tools_reference/emcc.html#emcc-pre-js

2. Moved the code that exposes the C API to to libhef.load(). This needs to be called after the WASM has loaded asynchronously using Module. Module.onRuntimeInitialized

```
window["Module"] = {
  onRuntimeInitialized: function () {
    libheif.load();
    init();
  },
};

````
3. Formatted the JS code. Let me know if its difficult to review after formatting. I can remove that.

aminya:
Why is this PR closed?

PR #476, opened 2021-04-10
Make the vendored rav1e completely static

lu-zero:
It sidetracks the problem of linking the static library without using the information provided by the pkg-config file.

Discovered when investigating https://github.com/lu-zero/cargo-c/issues/181


farindk:
Thank you

PR #482, opened 2021-04-26
Add rav1e.dll to make cmake find rav1e on MSVC targets

lu-zero:
Useful if the user wants to distribute the `rav1e.dll` along with `heif.dll`


farindk:
I don't understand why this is needed. In particular I'm wondering why the dll is added for rav1e and not for other libraries like dav1d or aom.

lu-zero:
Because msvc impl libs are free-form and `rustc` for the msvc uses the pattern `{name}.dll.lib`.
meson produces a different pattern (`{name}.lib`) and potentially aom might produce yet a different one, but I did not check it.

PR #483, opened 2021-04-28
Install examples when building w/ CMake

kmilos:
This makes install behaviour equivalent to autotools if examples are enabled
Also, use PNG on MinGW (both libpng and winpthreads are available)


farindk:
Thank you.

PR #484, opened 2021-04-29
Also install heif-thumbnailer example with CMake

kmilos:


farindk:
The normal `bin` directory is not the correct place for the thumbnailer. Instead it has to be installed in `/usr/share/thumbnailers`, together with the `heif.xml` in `/usr/share/mime/packages`.

kmilos:
This is for the `heif-thumbnailer` binary only, and I believe the correct location (autotools also install the same binary in `bin`).

This is a standalone binary that does not really depend on those other files, correct?

I presume your comment was for the diferent `heif.thumbnailer` file that should go to `share/thumbnailers`, which indeed does depend on `bin/heif-thumbnailer` and is part of Gnome? (Together w/ those xml files for registering the mime types in Gnome)? Should Gnome related files be installed by default on all systems?

farindk:
Yes, you are right. I misread my own documentation :-)

kmilos:
Thanks.

I in the long term I guess there should be a CmakeLists.txt in `/gnome` as well and an option in the main CmakeLists.txt to include it...

farindk:
Yes. I'll add an issue for it.

PR #485, opened 2021-04-30
Fix an out-of-bounds read with the AOM tune parameter array

0xC0000054:
The kParam_tune_valid_values array was missing the NULL pointer that
marks the end of the list.


farindk:
Thank you.

PR #489, opened 2021-05-03
Remove prepare-appveyor.cmd script

EwoutH:
The script is no longer needed for AppVeyor after the refactor #470.


fancycode:
Thanks for the quick PR!

PR #490, opened 2021-05-03
AppVeyor: Also publish build .dll files separately as artifacts

EwoutH:
Currently the build `.exe` files are separately uploaded to AppVeyor as artifacts, but since libheif is build dynamically those are useless without the `.dll` library. This PR fixes that.

Small follow up on #470.


farindk:
thanks

PR #497, opened 2021-05-08
aom: expose decoder error messages to aid debugging

lovell:
Errors that occur in the aom decoder currently return a generic "Success" message.

```sh
$ heif-convert testimage.avif out.png
File contains 1 images
Could not decode image: 0: Invalid input: Unspecified: Success
```

To make such problems easier to debug, this PR calls the `aom_codec_err_to_string` public API to lookup the underlying error mesage.

```sh
$ heif-convert testimage.avif out.png
File contains 1 images
Could not decode image: 0: Invalid input: Unspecified: Bitstream not supported by this decoder
```

As originally reported at https://github.com/lovell/sharp/issues/2688

### DIFF
--- a/libheif/heif_decoder_aom.cc
+++ b/libheif/heif_decoder_aom.cc
@@ -97,7 +97,7 @@ struct heif_error aom_new_decoder(void** dec)
 
     delete decoder;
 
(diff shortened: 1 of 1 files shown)


farindk:
Thank you

PR #501, opened 2021-05-12
cmake: install gnome thumbnailer files

### LINKED ISSUE #486: Install gnome thumbnailer files with CMake
There should be an option in CMakeLists.txt to also install the gnome thumbnailer files. See #484.
These are the files `heif.thumbnailer` and `heif.xml`/`avif.xml` in the gnome directory.

kmilos:
Closes #486


farindk:
Thank you!

PR #504, opened 2021-05-13
Detect and prevent negative overflow of clap box dimensions

lovell:
The numerators and denominators are subject to an implicit 32-bit unsigned to signed conversion, which can lead to negative values causing divide-by-zero when later rounded.

This PR detects and prevents negative values, although an alternative, slightly more complex approach might be to remove the unsigned to signed conversion.

The included example found via fuzz testing crashes the process with a floating point exception.


farindk:
Thanks, @lovell . I checked 14496-12 again and it says that the values in the `clap` box are unsigned. Hence, I'd like to try to solve this in a different way.

Unfortunately, I did not manage to download the input from the fuzzer. Not sure if this is a problem on my side or whether the file is not there. Could you please attach this file here or send it to me if you still have it?

lovell:
Thanks Dirk, the fuzzer-found file is part of this PR so you should be able to download it via https://github.com/strukturag/libheif/blob/e625a702ec7d46ce042922547d76045294af71d6/fuzzing/corpus/clap-overflow-divide-zero.heic

farindk:
Ok, got it now. I was confused by the github UI... :-)

farindk:
Thanks. I have merged it to get the .heic test case file, but I replaced the fix to make use of the fact that all `clap` fractions are in fact unsigned values.

lovell:
Great, that makes sense, thank you.

PR #506, opened 2021-05-14
Fix some of the issues found by Coverity Scan

fancycode:


libheif/heif_context.cc
```diff
@@ -1966,7 +1966,8 @@ Error HeifContext::encode_image_as_hevc(std::shared_ptr<HeifPixelImage> image,
                  err.message);
   }
 
-  int encoded_width, encoded_height;
+  int encoded_width = image->get_width(heif_channel_Y);
+  int encoded_height = image->get_height(heif_channel_Y);
```
farindk, line 1970: Could we initialize this 0 instead and return an "internal-error" if it's still 0 after parsing the SPS?

farindk: Even better: output an encoder plugin error, since every encoded stream should contain the SPS.

fancycode: Good idea, changed to ` heif_error_Encoder_plugin_error` in https://github.com/strukturag/libheif/pull/506/commits/98b867ea575ecce7039458b71f2c320742489e30.

farindk:
Thank you

Issue #509, opened 2021-05-15
gdk-pixbuf module memory-leaking


maclomhair:
Thanks for the good work of creating a gdk-pixbuf module. When looking at multiple images after each other using eog, the program start accumulating more and more memory with each image seen (up until I close it or the OOM killer kills it)

Since this only happens with heif photos (in my case, taken by Apple smartphones), I suppose it might be a memory leak in the gdk-pixbuf loader implementation provided here. If there are further details I can provide to help fix this bug, please let me know.

Platform used: Arch Linux
Software versions: libheif 1.12.0, gdk-pixbuf2 2.42.6, eog 40.1

maclomhair:
So I dug a little deeper and I think I might have found the problem. `gdk_pixbuf_new_from_data` initiates a GdkPixbuf with a reference count of one [1]. This probably needs to be unref'ed during cleanup in order for the actual freeing of the image data to occur.

The solution might be a simple `g_object_unref()` or maybe even better, a `g_clear_object()` in the cleanup routine.

[1] https://developer.gnome.org/gdk-pixbuf/unstable/gdk-pixbuf-Reference-Counting-and-Memory-Mangement.html

PR #511, opened 2021-05-15
Fix: decrease reference count on GdkPixbuf object so it can be freed

maclomhair:
GdkPixbuf objects are created with a reference count of 1. This fix decreases the reference count after use so it may actually be freed after use.

This fixes issue #509 - I tested on my machine and eog now really doesn't accumulate memory anymore.


fancycode:
lgtm

farindk:
Nice. Thank you @maclomhair .

PR #529, opened 2021-07-16
cmake: use same .pc VERSION substitute as autotools

kmilos:
Omits the fourth (tweak) version number and makes the .pc files identical between autotools and cmake


kmilos:
Obsoleted by https://github.com/strukturag/libheif/commit/da9eea36d3815ec064cafe47bf9a0ac59809fd8b

Issue #161, opened 2019-10-06
decoding and encoding progress report from library to client application


cgilles:
Decoding and encoding process can take a while especially with large image.

I see in API a progress handler in decoding structure:

struct heif_decoding_options
{
  uint8_t version;

  // version 1 options

  // Ignore geometric transformations like cropping, rotation, mirroring.
  // Default: false (do not ignore).
  uint8_t ignore_transformations;

  void (*start_progress)(enum heif_progress_step step, int max_progress, void* progress_user_data);
  void (*on_progress)(enum heif_progress_step step, int progress, void* progress_user_data);
  void (*end_progress)(enum heif_progress_step step, void* progress_user_data);
  void* progress_user_data;
};

I check the rest of the code to see if progress_user_data or on_progress are used, and i found nothing excepted an init to NULL at structure allocation.

It's planed to extend this progress handler in the future ?

Best

Gilles Caulier

farindk:
I'd like to have progress information, but that information ultimatively has to come from the decoder plugins, and as far as the decoders don't support it, I cannot add it to libheif.

We could add _some_ progress information for tiled images (number of tiles already decoded), which is maybe a good first step, because iPhone images are all stored in small tiles.

farindk:
Has been implemented in [009c34e](https://github.com/strukturag/libheif/commit/009c34e4c41e8cadc0933664c855e1957a9fcbaf).

cgilles:
Thanks for the job. We will take a look how to handle progress information in digiKam HEIF image loader.

farindk:
You just have to implement the `on_progress()` callback in the `heif_decoding_options`.
At the top of the above commit ([009c34e](https://github.com/strukturag/libheif/commit/009c34e4c41e8cadc0933664c855e1957a9fcbaf)), you can see how the callbacks are registered in `heif_dec`.

PR #546, opened 2021-09-06
Add progress report to tiles loading

mnaydenov:
Add a no-brainer implementation of the progress report, based on the suggestion in #161 
~~Currently only to the non-threaded implementation.~~

### DIFF
--- a/libheif/heif_context.cc
+++ b/libheif/heif_context.cc
@@ -1165,7 +1165,7 @@ Error HeifContext::decode_image_planar(heif_item_id ID,
       return error;
     }
 
-    error = decode_full_grid_image(ID, img, data);
+    error = decode_full_grid_image(ID, img, data, options);
     if (error) {
       return error;
     }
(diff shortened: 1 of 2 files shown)


mnaydenov:
@farindk Please consider changing the return type of `on_progress` to bool so that one can implement cancelation support. Considering these functions are not used, this will hardly be a breaking change. If this is considered unexpectable, adding `on_progress2` is an option, or `is_canceled` as a separate function. 
Progress reporting without the ability to cancel is needlessly handicapped.

bobfriesenhahn:
I used https://grep.app/ to check for how the decode options passed to heif_decode_image() are used.  I did not find any free software which currently attempts to use the progress monitor feature.  I have implemented such code for GraphicsMagick but will disable it.

Since there are no apparent users, it should be reasonable to change the function signature.

farindk:
Thanks @mnaydenov 

I've added progress reporting and user cancelation.
Sorry, I didn't use this PR because the code has changed so much in the meantime that it was easier to add this than trying to merge it.

PR #550, opened 2021-10-15
expose set_max_decoding_threads

mnaydenov:
Gives the C API access to the `HeifContext::set_max_decoding_threads`. For some reason it was missing, making it impossible to control threading, if enabled.


fancycode:
Thanks!

PR #555, opened 2021-10-28
Expose transformations

homm:
This PR introduce interface for reading transformations for given `heif_image_handle`.

# Rationale

The current rotation implementation is pretty straight and that is why inefficient on modern CPUs. For example, It could be up to 8.5 times slower than [Pillow's implementation](https://pillow.readthedocs.io/en/stable/index.html).

```python
In [2]: from HeifImagePlugin import Image, pyheif

In [3]: data = open('./full.heic', 'rb').read()

In [5]: %%timeit 
   ...: heif_file = pyheif.read(data) 
   ...: image = Image.frombytes( 
   ...:     heif_file.mode, heif_file.size, heif_file.data,
   ...:     "raw", heif_file.mode, heif_file.stride)
   ...:
683 ms ± 11 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %%timeit 
   ...: heif_file = pyheif.read(data, apply_transformations=False) 
   ...: image = Image.frombytes( 
   ...:     heif_file.mode, heif_file.size, heif_file.data,
   ...:     "raw", heif_file.mode, heif_file.stride)
   ...:
387 ms ± 10.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [13]: %%timeit 
    ...: heif_file = pyheif.read(data, apply_transformations=False) 
    ...: image = Image.frombytes( 
    ...:     heif_file.mode, heif_file.size, heif_file.data,
    ...:     "raw", heif_file.mode, heif_file.stride)
    ...: image = image.transpose(Image.ROTATE_270) 
    ...:
422 ms ± 5.66 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

> Time spent to rotation in libheif: 683 - 387 ms = 296 ms.
> Time spent to rotation in Pillow: 422 - 387 ms = 35 ms.

There is already an flag `ignore_transformations` which could be used to prevent `libheif` from transforming the image. Unfortunately there was no way to get image transformation properties from the file.

So, the new struct `heif_transformations` is introduced. It describes image's transformations (information from `irot`, `imir` and `clap` boxes) in strict order (unlike in the file itself, where operations could be in arbitrary order).

In addition, for some cases, transformation from this struct could be used for EXIF construction and actual image rotation could be totally avoided.

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -1643,8 +1643,9 @@ struct heif_error heif_image_handle_get_transformations(const struct heif_image_
     return err.error_struct(handle->image.get());
   }
 
-  handle->image->read_transformations(transformations);
-  return Error::Ok.error_struct(handle->image.get());
+  Error err = handle->image->read_transformations(transformations);
+
+  return err.error_struct(handle->image.get());
 }
 
 
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -1609,6 +1609,45 @@ void heif_nclx_color_profile_free(struct heif_color_profile_nclx* nclx_profile)
 }
 
 
+struct heif_transformations* heif_transformations_alloc()
+{
+  auto transformations = (heif_transformations*) malloc(
+    sizeof(struct heif_transformations)
+  );
+
+  if (transformations) {
+    transformations->version = 1;
+    transformations->crop_left = 0;
+    transformations->crop_top = 0;
+    transformations->crop_width = 0;
+    transformations->crop_height = 0;
+    transformations->orientation_tag = 0;
+  }
+
+  return transformations;
+}
+
+
+void heif_transformations_free(struct heif_transformations* transformations)
+{
+  free(transformations);
+}
+
+
+struct heif_error heif_image_handle_get_transformations(const struct heif_image_handle* handle,
+                                                        struct heif_transformations* transformations)
+{
+  if (transformations == nullptr) {
+    Error err(heif_error_Usage_error,
+              heif_suberror_Null_pointer_argument);
+    return err.error_struct(handle->image.get());
+  }
+
+  handle->image->read_transformations(transformations);
+  return Error::Ok.error_struct(handle->image.get());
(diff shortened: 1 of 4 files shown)


bigcat88:
An additional flag would also be useful to perform all transformations except for the final rotation as @kmilos says.
Just expose and read transformations is not enough, that will disable cropping, that used for non-odd size images.

farindk:
> but there is no way to get context back from the handler. 

I have no problems with adding a function to get the `heif_context` from an `heif_image_handle`. Similarly, We can also add some convenience functions to use `heif_image_handle` in place of `heif_context` + `heif_item`.

I'd like to give you some background why `heif_item` is used in many of the new APIs: there are properties (e.g. `udes`) that can be attached to images and also non-image items, e.g. `rgan` items. Thus it feels logical to operate on item item-ID level. There are also some features in the pipeline where it is better to operate on items directly. E.g. accessing image grids or overlay layers separately. For these kind of features, we probably need to have a more lower-level API (with item-IDs) that works close along the standard, and a higher level API (using image handles) for those who just want a quick, easy interface.

Concerning returning the transformations in a 'canonical' order: this may have some nasty consequences. E.g. rotating and cropping an image may require a conversion to a different chroma format while the other order of rotation and cropping may not require this. Now, in HEIFv2, they specified a scaling transform, which makes things even more difficult.

But I like the idea and didn't abandon it. We just have to be very careful that we don't maneuver ourselves into a dead end.

homm:
> Concerning returning the transformations in a 'canonical' order: this may have some nasty consequences.

Just for the record, I don't have any questions to this part. It's up to library: provide "close to the file" or more general format.

homm:
There are questions about the new API https://github.com/strukturag/libheif/issues/919 https://github.com/strukturag/libheif/issues/920

farindk:
@homm I've added an API function to get the `heif_context` from an `heif_image_handle`.

PR #558, opened 2021-11-02
Add irot and imir properties to HeifFile

homm:


libheif/heif_file.cc
```diff
@@ -708,14 +708,30 @@ void HeifFile::add_clap_property(heif_item_id id, uint32_t clap_width, uint32_t
...
-  irot->set(rotation);
+  irot->set_rotation(rotation);
 
   int index = m_ipco_box->append_child_box(irot);
 
   m_ipma_box->add_property_for_item_ID(id, Box_ipma::PropertyAssociation{true, uint16_t(index + 1)});
 }
 
 
+void HeifFile::add_imir_property(heif_item_id id, int vertical)
```
farindk, line 719: I think it would be cleaner to use the `MirrorDirection` type instead of `int` for the `vertical` parameter.

farindk:
Thanks for the addition. Please see me comment above. Other than that, I am fine with the PR.

farindk:
There is now `HeifFile::add_orientation_properties()`. Do we still need this PR?

PR #565, opened 2021-11-12
Extract Exif metadata or macOS dynamic wallpaper timing metadata

xyproto:
Hi,

I created a collection of utilities named [wallutils](https://github.com/xyproto/wallutils), where the `settimed` utility is able to set dynamic wallpapers in the GNOME XML or Simple Timed Wallpaper format.

In order to also support macOS dynamic wallpaper HEIF files (with the `.heic` extension), I needed to be able to extract the timing metadata.

The Go code in `strukturag/libheif/go` did not support this, so I added a new file, `metadata.go`, that contains this functionality. I have tested the new code with the `heic2stw` utility that comes with `wallutils` and it works reliably and with good performance (thanks to the good performance of libheif).

I also included a `metadata_test.go` file, for testing the metadata extraction of macOS dynamic wallpaper files.

I could not include the test-file `examples/mont.heic`, because the license is unclear. It can be downloaded from here, though: https://dynamicwallpaper.club/wallpaper/la4wfuwtkg. I sent an e-mail to `taylorgrayphoto@gmail.com`, the author of that image, to ask if he would be willing to release it under a CC license. I will add a comment to this pull request if I get a reply.

Two Go module files were also added, to make testing and developing much easier. With these files in place, running `go test` and `go build` in the `go` directory works as expected (tested with Go 1.17).

The `metadata_test.go` test will only pass once an appropriate example image is in place, though.

Thanks for creating `libheif` in the first place, it's a useful library!

Best regards,
Alexander F. Rødseth


go/heif/metadata_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	metadataID := metadataIDs[0]
+
+	times, err := handle.ImageTimes(metadataID)
+	if err != nil {
+		t.Fail()
+	}
+	fmt.Println(times)
+
+	thumbnail := false
```
xyproto, line 57: The rest of the test, from here on and out, is the same as in heif_test.go, and can potentially be skipped.

xyproto: I added code that skips this, and also skips the entire test if the example image (which may be copyrighted) does not exist in the repository.

farindk:
Thanks. This is interesting.
I found a couple of more images and some description on how this works here: https://github.com/mczachurski/wallpapper

farindk:
I had to revert this PR because I got errors in the CI that I could not resolve (I've no idea about Go):
https://github.com/strukturag/libheif/runs/5260368540?check_suite_focus=true

Seems that this is because of the new dependencies on xmlquery, xpath, and plist.
Can you fix this and create a new PR?

Maybe it would also be better to put all code that requires these dependencies into a separate file that can be compiled optionally.

xyproto:
Thanks for looking into this. If `go get -d` is run before `go build` / `go test`, then all required dependencies should be downloaded.

If downloading dependencies during CI is not desired, vendoring the dependencies with `go tidy` and then `go vendor` is possible.

Using `xmlquery`, `xpath` and `plist` is for extracting the Apple-specific metadata from the `heic`/`heif` file, and are intentional dependencies.

PR #567, opened 2021-11-30
CI: Add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR. This is a service offered by OSS-Fuzz, on which libheif already runs.

I noticed you already had a fuzzer-related action, but I wanted to offer this as you might be interested and it was recently released (https://security.googleblog.com/2021/11/clusterfuzzlite-continuous-fuzzing-for.html)

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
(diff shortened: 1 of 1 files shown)


fancycode:
Thanks!

homm:
@fancycode could you also look at #551 ?

PR #569, opened 2021-12-01
Enable wide chars for all Windows

kmilos:
This is platform specific, not compiler specific (i.e. it is also implemented by libstdc++ and libc++ on Win32 such as MinGW).

### DIFF
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -14,7 +14,7 @@ jobs:
           - { MINGW: 32 }
           - { MINGW: 64 }
     env: ${{ matrix.env }}
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
 
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -1513,7 +1513,7 @@ static struct heif_error heif_file_writer_write(struct heif_context* ctx,
 {
   const char* filename = static_cast<const char*>(userdata);
 
-#ifdef _MSC_VER
+#if defined(__MINGW32__) || defined(__MINGW64__) || defined(_MSC_VER)
   std::ofstream ostr(HeifFile::convert_utf8_path_to_utf16(filename).c_str(), std::ios_base::binary);
 #else
   std::ofstream ostr(filename, std::ios_base::binary);
--- a/libheif/heif_file.cc
+++ b/libheif/heif_file.cc
@@ -27,13 +27,13 @@
 #include <cstring>
 #include <cassert>
 
-#ifdef _MSC_VER
+#if defined(__MINGW32__) || defined(__MINGW64__) || defined(_MSC_VER)
 
 #ifndef NOMINMAX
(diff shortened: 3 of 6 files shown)


kmilos:
Hm, the CI for ming32/mingw64 fails, presumably because your cross-compile toolchain on Ubuntu doesn't have wchar enabled? The `7.3-posix` in

```
heif_file.cc:70:149: error: no matching function for call to ‘std::basic_ifstream<char>::basic_ifstream(const wchar_t*, const openmode&)’
   auto input_stream_istr = std::unique_ptr<std::istream>(new std::ifstream(convert_utf8_path_to_utf16(input_filename).c_str(), std::ios_base::binary));
                                                                                                                                                     ^
In file included from heif_file.cc:23:0:
/usr/lib/gcc/x86_64-w64-mingw32/7.3-posix/include/c++/fstream:521:7: ...
```

makes me think this is not a real Win32 toolchain...

[There is no problem building this on Msys2 w/ "native" toolchains.](https://github.com/msys2/MINGW-packages/pull/10228/checks)

kmilos:
Yep, it seems [there is also a `7.3-win32` toolchain](https://packages.ubuntu.com/bionic/amd64/gcc-mingw-w64-x86-64/filelist) available, but no idea how prefer the use of that one...

kmilos:
Scratch that, posix/win32 is only about threading model...

Wide char support is available only from g++-mingw-w64-x86-64_8.3.0, so a CI update is more than welcome.

examples/heif_thumbnailer.cc
```diff
@@ -42,7 +42,7 @@
 #  include "encoder_png.h"
 #endif
 
-#if defined(_MSC_VER)
+#if defined(__MINGW32__) || defined(__MINGW64__) || defined(_MSC_VER)
```
lazka, line 45: I think you can just `#ifdef _WIN32`

kmilos: Thanks, I was thinking about that too, but then wasn't be sure one wants this for Cygwin/MSYS as well (perhaps multibyte UTF8 just works there)?

In either case, still stuck w/ the older libstdc++ problem...

W/ the separate MSVC/MINGW, one could potentially also check for `__GLIBCXX__` version...

lazka: I don't think that's relevant on cygwin, locale encoded filenames should just work there.

lazka:
I've created #570

fancycode:
@kmilos could you please check if your changes work with the updated toolchain from #570?

kmilos:
Well, [they certainly work on the latest MSYS2 toolchains](https://github.com/msys2/MINGW-packages/pull/10228), but I can later perhaps try and rebase on top of #570 if you want to check the ubuntu-20.04 image specifically?

fancycode:
> [...] I can later perhaps try and rebase on top of #570 if you want to check the ubuntu-20.04 image specifically?

That would be perfect, thanks!

fancycode:
Nice, thanks for confirming.

PR #570, opened 2021-12-03
CI: use Ubuntu 20.04 for the mingw build to get a newer toolchain

lazka:
Needed a workaround for installing wine32 due to the GHA image being
broken (see https://github.com/actions/virtual-environments/issues/4589) (could also be worked around by using a docker image).

Might help #569


fancycode:
Thanks!

lazka:
Thanks!

PR #583, opened 2022-01-03
aom: ensure NCLX property values are within known bounds

lovell:
Hello, the `fuzzing/corpus/unsupported-nclx-profile.avif` image included in this PR was discovered via fuzz testing and has an invalid NCLX colour primaries value of 32, which is accepted by `aom_codec_get_frame`, but this does not exist in the `heif_color_primaries` enum.

This PR ensures values for colour primaries, transfer characteristics and matrix coefficients are within the bounds of the enums they are cast to. It adds a new suberror code to return when they are not. I note some of the enums have "holes" for reserved values; perhaps should we consider these an error also?


farindk:
Thanks.

I tried to reproduce this with the provided `unsupported-nclx-profile.avif`, but for me, it already gave an error in `aom_push_data -> aom_codec_decode()` (AOM_CODEC_CORRUPT_FRAME). This is with the aom package from Ubuntu hirsute (libaom0/hirsute 1.0.0.errata1-3build1). What did you use to get past `aom_codec_decode()` ?

I am also wondering if exiting with an error is the best way to handle this. How about giving a warning and sanitizing the values to some default?

lovell:
Thanks Dirk, this was discovered with the latest aom (v3.3.0) via OSS-Fuzz using clang's `UndefinedBehaviorSanitizer`.

I guess it's possible this might be a regression in aom, but there's already existing logic in libheif to deal with aom providing images with unknown `fmt`, and I think having libheif be explicit about what it does and does not support is a good thing.

In terms of this sample image that's part of this PR, it's definitely invalid and I think I'd prefer that this generates an error, but happy to switch to a warning if you think that's more appropriate.

Is there an example elsewhere in the libheif code case of how this might work in practice? Writing to stderr probably isn't the best approach for this when libheif is consumed as a library, its primary use case. What alternative value(s) should we sanitise to?

farindk:
Sure, writing to `stderr` would not be a good option. My proposal would be to store all decoding warnings in the `heif_image` object. After decoding, one can query the warnings. The warnings could simply be a list of `heif_error`s.

Moreover, there could be a `strict` flag in the decoding options that lets the decoder return an error instead of adding warnings to the image.

lovell:
Thanks, introducing a new `strict` flag seems like a good idea as it allows people to opt-in to an "I trust this input, do your best" approach. Shall I create a separate enhancement issue to track this?

In the mean time, would you prefer this PR switches to an approach where invalid values fall back to using sensible defaults without generating warnings?

lovell:
I've opened https://github.com/strukturag/libheif/pull/612 with an alternative approach based on your feedback, thank you, and with what I think is a better approach overall.

farindk:
I have implemented the whole thing. Sorry for the double work, but this now includes everything with warnings stored in the images and a command-line switch `-s` for `heif-convert` to turn on strict mode.

I've tried to keep the boiler-plate code small in the decoder plugin:
https://github.com/strukturag/libheif/blob/7041ef60a05428e11e667c096b7fd206a7e2858e/libheif/heif_decoder_aom.cc#L216

lovell:
Thanks, no worries, I'm reviewing your recent commits now.

PR #594, opened 2022-01-25
Remove reference to "XMP" in add_generic_metadata

wantehchang:
Remove a reference to "XMP" in a comment in the add_generic_metadata()
method. This method is used to add all kinds of data, not just XMP data.


wantehchang:
@farindk Please review. Thanks!

PR #599, opened 2022-02-02
AOM: Add separate quality options for the alpha image

0xC0000054:
This allows callers to use different compression settings for the color
and alpha images.

Because of the backwards compatibility concerns I did not make lossless compression the default for the alpha image.
There are other AVIF encoders (e.g. libheif, Paint.NET) that do use lossless alpha compression as the default option.

### DIFF
--- a/libheif/heif_encoder_aom.cc
+++ b/libheif/heif_encoder_aom.cc
@@ -44,14 +44,24 @@ struct encoder_struct_aom
   int cpu_used;  // = parameter 'speed'. I guess this is a better name than 'cpu_used'.
 
   int quality;
+  int alpha_quality;
   int min_q;
   int max_q;
+  int alpha_min_q;
+  int alpha_max_q;
   int threads;
   bool lossless;
+  bool lossless_alpha;
 
   aom_tune_metric tune;
 
   heif_chroma chroma = heif_chroma_420;
+
+  // --- input
+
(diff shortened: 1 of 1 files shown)


farindk:
Thanks.

I just wanted to note that the old `lossless` now also implies `lossless-alpha`. I.e. `lossless=true` + `lossless-alpha=false` will still encode the alpha lossless. I am not sure if I should be concerned about this. It is the backward compatible way and probably also how it is intended, but it may still be confusing to the user. Especially if `lossless-alpha=false` is set explicitly.

0xC0000054:
> I just wanted to note that the old `lossless` now also implies `lossless-alpha`. I.e. `lossless=true` + `lossless-alpha=false` will still encode the alpha lossless. I am not sure if I should be concerned about this. It is the backward compatible way and probably also how it is intended...

That was the intended design.
I did not make `lossless-alpha=true` to preserve backwards compatibility with the previous behavior of `lossless`, but other encoders (e.g. libavif) make `lossless-alpha=true` the default.

> ...but it may still be confusing to the user. Especially if `lossless-alpha`=false is set explicitly.

The `lossless-alpha` option was intended for when `lossless=false` and the user wanted only the alpha channel to use lossless encoding. 
I did not consider the reverse case, and I am not sure how that would be fixed.

PR #600, opened 2022-02-03
Add support for the AOM key=value API

### LINKED ISSUE #596: Add support for the AOM encoder key=value API
AOM version 3.0.0 added a key=value API to configure the decoder and encoder parameters.

`aom_codec_err_t aom_codec_set_option(aom_codec_ctx_t *ctx, const char *name, const char* value);`

This would allow consumers to pass parameters to the encoder without needing to extend the libheif AOM wrapper to support each new parameter.

0xC0000054:
This allows callers to pass custom options to the AOM encoder by adding
an 'aom:' prefix to the option name.
Th custom options will be applied after libheif configures the default
options because they can override the values that were set.

The x265 encoder already has a similar feature with the 'x265:' prefix.

Resolves #596

### DIFF
--- a/libheif/heif_encoder_aom.cc
+++ b/libheif/heif_encoder_aom.cc
@@ -35,6 +35,19 @@
 #include <aom/aom_encoder.h>
 #include <aom/aomcx.h>
 
+// Detect whether the aom_codec_set_option() function is available.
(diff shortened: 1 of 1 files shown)


farindk:
Thank you !

PR #610, opened 2022-02-21
limit exif data size

jl2005:
When the exif data of the HEIF file exceeds the limit, it cannot be converted to a jpeg image. It is necessary to judge the size of the exif data during conversion.

[bad2.heif.zip](https://github.com/strukturag/libheif/files/8108030/bad2.heif.zip)

### DIFF
--- a/examples/encoder_jpeg.cc
+++ b/examples/encoder_jpeg.cc
@@ -166,10 +166,12 @@ bool JpegEncoder::Encode(const struct heif_image_handle* handle,
 
   size_t exifsize = 0;
   uint8_t* exifdata = GetExifMetaData(handle, &exifsize);
-  if (exifdata && exifsize > 4) {
-    static const uint8_t kExifMarker = JPEG_APP0 + 1;
-    jpeg_write_marker(&cinfo, kExifMarker, exifdata + 4,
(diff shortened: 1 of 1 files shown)


jl2005:
@alex @xyproto @piponazo @homm 
Please review this PR.

silverbacknet:
APPx markers can be repeated to construct a larger unit of data. Rather than simply limiting APP1 to 64k, it might be better to split it, the way APP2 markers usually are.

But as a sanity check that an APP1 shouldn't overflow 16-bit, it's better than nothing.

farindk:
Yes, let's go to the multiple APP1 markers right away.

farindk:
@jl2005 Please see my change in [5de6ed3](https://github.com/strukturag/libheif/commit/5de6ed30131c402460e2dbb74588ef38e273287b) to save Exif markers >64k. Does this solve it?

jl2005:
> @jl2005 Please see my change in [5de6ed3](https://github.com/strukturag/libheif/commit/5de6ed30131c402460e2dbb74588ef38e273287b) to save Exif markers >64k. Does this solve it?

Yes, this is better than limiting.

PR #612, opened 2022-02-21
Add API to sanitise enums relating to colour profiles

lovell:
This is an alternative to and more substantial approach than #583 to help sanitise values relating to colour profiles, which covers the aom and dav1d decoders as well as internal colour conversion. There are no errors generated, and unknown/invalid values become "unspecified". This approach also deals with the "hole" problem with enums, e.g. the missing colour primary value 3 can be handled.

It exposes the ability to do so via three new public functions (US English spelling to match existing API):

```c
enum heif_color_primaries heif_color_primaries_sanitize(uint16_t cp);

enum heif_transfer_characteristics heif_transfer_characteristics_sanitize(uint16_t tc);

enum heif_matrix_coefficients heif_matrix_coefficients_sanitize(uint16_t mc);
```

For unknown input integer values, these functions all return their respective `heif_*_unspecified` enum value.

All existing public functions remain as-is, so they still accept and return unsigned 16-bit integer values, ensuring both backwards and forwards compatibility. This allows for an image created with a future libheif that supports a wider range of colour primaries to be parsed by an older libheif with a narrower range.

This PR also:

- Uses these new functions to sanitise values provided from the command line in the example encoder tool.
- Uses these new functions to sanitise values read from inputs in the aom and dav1d decoders.
- When calculating Kr/Kb, ensures `heif_color_primaries_unspecified` does not cause divide-by-zero.
- Adds a couple of images discovered via fuzz testing with invalid/unknown values for these enums.


farindk:
Ah, I'm just writing the same at the very moment...

lovell:
Closing as this is superseded by your great work, as always, thanks Dirk.

PR #625, opened 2022-03-27
Fix CMake dev warnings on names for find_package

wantehchang:
CMake emits a dev warning if the package names passed to find_package()
and find_package_handle_standard_args() do not match (case-sensitive).
Here is an example of the warning for the find_package(Libde265) call in
CMakeLists.txt and the find_package_handle_standard_args(LIBDE265 ...)
call in cmake/modules/FindLibde265.cmake:

-- Found LIBDE265
CMake Warning (dev) at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (LIBDE265)
  does not match the name of the calling package (Libde265).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/modules/FindLibde265.cmake:32 (find_package_handle_standard_args)
  CMakeLists.txt:62 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found LIBDE265: /usr/local/include (found version "2.0.0")


wantehchang:
@farindk @fancycode Please review. Thanks.

There is more than one way to fix the mismatches. Please let me know if you prefer a different fix.

wantehchang:
I abandoned this pull request. I believe the issue has been fixed.

PR #634, opened 2022-04-23
msys problems: do not rely on an extern variable, use api instead

dk:
Hello,

The msys/win32 problems seems to be still unsolved in 2022 when one relies on exported variables, rather than exported exported functions. There, code trying to reach char* libx265.x265_version fails with dll resolve error. I would guess one could fix that either by nudging libx265 to properly export their symbols on msys/win32, or some other way, but this fix I applied seems the cleanest.

Kindly review
Regards,
Dmitry

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -96,6 +96,28 @@ int heif_get_version_number_maintenance(void)
   return ((LIBHEIF_NUMERIC_VERSION) >> 8) & 0xFF;
 }
 
+static int use_multithreading =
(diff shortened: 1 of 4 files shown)


farindk:
Thanks, I have added this in commit ffeb827. I didn't merge the PR directly, because there was also another commit about multi- threading, which is obviously not related to this. If you think that is also worth merging, please add it in a new PR.

dk:
Great! Just submitted it anew

PR #642, opened 2022-05-31
Added extra check to prevent recursion that could result in a stack overflow.

dlemstra:
This PR adds an extra check to prevent a possible stack overflow when the first value of `image_references` is the same as the `id` argument. I did not add an extra error message but if this is required then please let me know what the error message should be.


fancycode:
Thanks!

PR #647, opened 2022-06-28
Enable GdkPixbuf loader for mingw as well

kmilos:
Also, a DLL is not a "LIBRARY" type on Windows, let CMake handle this automatically.


farindk:
Thank you

Issue #535, opened 2021-08-13
Possible leak in x265 while decoding 


pierrejoye:
Example code, relatively classic usage:

https://github.com/libgd/libgd/blob/master/src/gd_heif.c#L113

I can only reproduce using our test:

https://github.com/libgd/libgd/blob/master/tests/heif/heif_im2im.c#L36

which reads using heif_context_read_from_memory_without_copy.

I am not completely sure where the allocations (2x leak):

=================================================================
==31634==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2048 byte(s) in 2 object(s) allocated from:
    #0 0x7f049d924aa5 in posix_memalign (/lib/x86_64-linux-gnu/libasan.so.5+0x10eaa5)
    #1 0x7f049c33eb47 in x265::x265_malloc(unsigned long) (/lib/x86_64-linux-gnu/libx265.so.179+0x45db47)

SUMMARY: AddressSanitizer: 2048 byte(s) leaked in 2 allocation(s).

I also attached the valgrind log (ran it too), but there may be a lot of false positive, or a lot of other issues in x265 ;-)

[vg_logfile.txt](https://github.com/strukturag/libheif/files/6980662/vg_logfile.txt)

pierrejoye:
Same in encoding

=================================================================
==32711==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2048 byte(s) in 2 object(s) allocated from:
    #0 0x7f2f7c195aa5 in posix_memalign (/lib/x86_64-linux-gnu/libasan.so.5+0x10eaa5)
    #1 0x7f2f7ab92b47 in x265::x265_malloc(unsigned long) (/lib/x86_64-linux-gnu/libx265.so.179+0x45db47)

SUMMARY: AddressSanitizer: 2048 byte(s) leaked in 2 allocation(s).

pierrejoye:
Additional info, expected ;)

![image](https://user-images.githubusercontent.com/282408/130344872-570f6c0c-26a5-47ba-acf0-98c6783c0ac2.png)

libde265_free_decoder or other are never used.

What was the idea or design behind it? As it may matter less in CLI apps, it is absolutely critical to free them in server environments. Happy to work on a PR however I would need to know what was the idea here :)

bradh:
@pierrejoye Are you still seeing this problem?

Issue #531, opened 2021-07-20
memory leak in de265_init()


liuchang3:
libheif version: 1.10.0

==25882== 11,296 bytes in 1 blocks are still reachable in loss record 7 of 7
==25882==    at 0x4C30FB5: malloc (vg_replace_malloc.c:380)
==25882==    by 0x53B7637: alloc_and_init_significant_coeff_ctxIdx_lookupTable() (in /home/ADC/80301694/Work/2021-06-24-Base/TBLMultiMediaCloud/tbl_image/out/lib/Tbl_Image_Kit/lib/libtbl_image_kit.so.1.0.8)
==25882==    by 0x539F229: de265_init (in /home/ADC/80301694/Work/2021-06-24-Base/TBLMultiMediaCloud/tbl_image/out/lib/Tbl_Image_Kit/lib/libtbl_image_kit.so.1.0.8)
==25882==    by 0x5363B41: heif::register_decoder(heif_decoder_plugin const*) (in /home/ADC/80301694/Work/2021-06-24-Base/TBLMultiMediaCloud/tbl_image/out/lib/Tbl_Image_Kit/lib/libtbl_image_kit.so.1.0.8)
==25882==    by 0x4F4AD38: _GLOBAL__sub_I_heif_plugin_registry.cc (in /home/ADC/80301694/Work/2021-06-24-Base/TBLMultiMediaCloud/tbl_image/out/lib/Tbl_Image_Kit/lib/libtbl_image_kit.so.1.0.8)
==25882==    by 0x4010732: call_init (dl-init.c:72)
==25882==    by 0x4010732: _dl_init (dl-init.c:119)
==25882==    by 0x40010C9: ??? (in /lib/x86_64-linux-gnu/ld-2.27.so)
==25882==    by 0x1: ???
==25882==    by 0x1FFF00004A: ???
==25882==    by 0x1FFF000060: ???

i check the code,   there is de265_free function but not be called, looks like missing call deinit_plugin
the issue is found by imagemagick test .

bradh:
I cannot reproduce this problem. Can you show the repro? Exact configuration, exact usage, please.

bradh:
It looks like this was fixed by #654 and can be closed.

PR #654, opened 2022-07-26
Unregister default plugins on deinitialization

stepkamipt:
Unregister default plugins to prevent memory leaks in plugin's objects, for example, ctxIdxLookup in libde265

### DIFF
--- a/libheif/heif_plugin_registry.cc
+++ b/libheif/heif_plugin_registry.cc
@@ -101,6 +101,15 @@ public:
     heif::register_decoder(get_decoder_plugin_dav1d());
 #endif
   }
+
+  ~Register_Default_Plugins()
+  {
+     for (const auto* plugin : heif::s_decoder_plugins) {
+       if (plugin->deinit_plugin) {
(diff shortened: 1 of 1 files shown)


farindk:
I'm wondering when the destructor will ever be called since this is a static, global singleton variable.

stepkamipt:
@farindk yes, you're right, thank you. So, I created another PR with API methods to deinitialize plugins manually:
https://github.com/strukturag/libheif/pull/656

And I found that there are couple of issues related to problem I'm facing: https://github.com/strukturag/libheif/issues/531 and https://github.com/strukturag/libheif/issues/535

farindk:
Thank you. I have merged this, but changed the usage. Instead of calling `heif_unregister_*()`, it is now good practice to initialize and cleanup with `heif_init()` and `heif_deinit()`. These are reference counted, `heif_deinit()` will then effectively call `heif_unregister_*()`.
Please try in the master branch and let me know if you see any issues with this.

stepkamipt:
@farindk thank you! I've try the master branch and everything works perfect.

PR #662, opened 2022-08-18
Added missing return value checks that could cause an access violation because the output plane was not allocated.

dlemstra:


fancycode (changes requested):
Thanks, marked two places where the result should be checked, too.

Also line 2555 is missing a check:
https://github.com/strukturag/libheif/pull/662/files#diff-ae32c56934c4dee29e46c00f3acba5056fb682f16c47a8fc3562962a7dbd42f4R2555

dlemstra:
I have updated the PR to include those two other checks.

fancycode:
This one is still missing (only added as comment in my review):
https://github.com/strukturag/libheif/blob/92c3951eaf6ee543d6733143977886d73fd7bd50/libheif/heif_colorconversion.cc#L2559

dlemstra:
Updated the PR again.

fancycode (approved):
Thanks, lgtm!

fancycode:
Could you please rebase this on master?

dlemstra:
It looks like some of the tests are failing now. Will take a look at this later today/this weekend. Feel free to give me some pointers on how to fix this.

dlemstra:
Found the mistake that I made. I also made a tiny change the go unit tests while I was investigating this.

fancycode:
> Found the mistake that I made. I also made a tiny change the go unit tests while I was investigating this.

Perfect, thanks. Merging then.

Issue #635, opened 2022-04-28
crashes: add api for turning off multithreading


dk:
Applications that would like to have a tight control over usage of threads may crash if implicit multithreading is used by libheif. While 3rd party installations may configure libheif with multithreading, and explicit call to heif_global_set_multithreading(0) or something similar would be very much appreciated for these

farindk:
There is no api function to turn off multithreading, but there is a ``configure`` option to do so. This also excludes importing threading-related types (mutex, thread) which is also useful if they aren't available.

dk:
here I quickly cooked a patch to demonstrate what I mean: not the compile-time unavailablilty, but the runtime control

farindk:
Thanks. Can you submit the patch as a pull request?

farindk:
Oh, I see you put it into the other PR. Never mind, I'll get it from there.

dk:
Thank you! It's the githubs way of doing things, apparently, it automatically assumes the subsequent commits belong to the same pr/pull requests.

farindk:
You would have to make separate branches for this to work as two separate PRs. But don't mind. This patch is simple enough so that we don't have to split it.

PR #667, opened 2022-08-31
refs #635: add api for turning off multithreading

dk:
Applications that would like to have a tight control over usage of threads may crash if implicit multithreading is used by libheif. While 3rd party installations may configure libheif with multithreading, and explicit call to heif_global_set_multithreading(0) or something similar would be very much appreciated for these

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -96,6 +96,28 @@ int heif_get_version_number_maintenance(void)
   return ((LIBHEIF_NUMERIC_VERSION) >> 8) & 0xFF;
 }
 
+static int use_multithreading =
+#ifdef ENABLE_PARALLEL_TILE_DECODING
+  1
+#else
+  0
+#endif
+;
+
+int heif_global_get_multithreading(void)
+{
+	return use_multithreading;
+}
+
+int heif_global_set_multithreading(int flag)
+{
+#ifdef ENABLE_PARALLEL_TILE_DECODING
+  use_multithreading = flag ? 1 : 0;
+  return use_multithreading;
+#else
+  return 0;
+#endif
+}
 
 heif_filetype_result heif_check_filetype(const uint8_t* data, int len)
 {
--- a/libheif/heif.h
+++ b/libheif/heif.h
@@ -83,6 +83,11 @@ LIBHEIF_API int heif_get_version_number_maintenance(void);
 #define LIBHEIF_MAKE_VERSION(h, m, l) ((h) << 24 | (m) << 16 | (l) << 8)
 #define LIBHEIF_HAVE_VERSION(h, m, l) (LIBHEIF_NUMERIC_VERSION >= LIBHEIF_MAKE_VERSION(h, m, l))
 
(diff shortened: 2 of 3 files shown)


erdogan61:
Hi, what was the reason to close? Was it just a prototype?

dk:
Hi, the reason was that it doesn't seem to attract much of attention, and nothing happened since Apr 28, so I deleted my fork. It seems that as an unintended action, github also closed the pr - so I'll reopent it. But please consider to add that option, because my project that uses perl5 binding, simply crashes without it because perl5 wants to do the thread management itself. @farindk kindly review

farindk:
I'm wondering how any client code can crash just because a library is using multithreading. In fact, even if multithreading is turned off for libheif, the codecs will still use multithreading. Thus, I don't see the point yet why this makes a difference. Any reference to any perl docu that describes such a behavior?

dk:
This is just my best guess - actually my project also uses multithreading (though indirectly via libgomp), and all is fine. So the only thing I can say with certainty is that it is not the multithreading per se that causes the crash, but the ifdef'd code (IIRC I tracked the crashed line to be either "Error e = errs.front().get();" or "errs.pop_front();"). Windows wasn't particularly helpful with providing a good, gdb-style report, I found that by recompiling the library with printf's. Whether that is a combination of C++, threading, dlls, win32, or something else, no idea, I'm not a C++ expert. But I won't be able to submit a PR either that wouldn't involve the whole bunch of my project, perl5, etc etc just to demonstrate the crash.

But of course if you don't see the different then nevermind, I don't think I have more arguments.

erdogan61:
Hi dk,
in the project which i am working on, also multithreading is used to speed up the process but only if several heif is processed.
With activated libheif multiencoding you have then the situation of multithreads in every multithread:
Project processing:
 -> Multithreading of several decoding of heic files
      -> Each decoding also activated multithreading
So you probably have here no real benefit maybe even worse only overhead and maybe thread in thread problems as you have.

In spite of that if the project process wants to open just one heif file then its own multithreading is useless and has no benefits. But the multiencoding of libheif will speed up the decoding of the heif file.

Furthermore the default config in vcpkg is multithreading off, which implies to do some reconfiguration of the compiling. With an api it would be also much easier to handle, internally it can still be default off.

Thats why I am also interested in this feature.

dk:
@erdogan61 I see and I agree - libgomp that I use for implicit multithreading also gives real benefits withs speed, BUT - it is runtime-configurable, while libheif is not.

farindk:
Ok, I see the reason why this can be useful.
However, instead of a global flag to turn it off completely, I would favor a flag in the encoding and decoding options to selectively turn this off, so that if the library is used in different parts of an application, they do not interfere with each other.

dk:
That is even better, I agree

dk:
https://github.com/strukturag/libheif/pull/737

dk:
@farindk do you have a possibility to review the updated request?

farindk:
Closing since we merged #737 instead.

PR #670, opened 2022-09-05
build: allow to disable dav1d

kleisauke:
Adds a new optional `--disable-dav1d` autotools flag,
which defaults to "no".

Diff is best viewed with the "Hide whitespace" option.


farindk:
Thank you.

PR #679, opened 2022-10-01
AOM: Add monochrome decoding support

0xC0000054:


libheif/heif_decoder_aom.cc
```diff
@@ -231,6 +239,10 @@ struct heif_error aom_decode_image(void* decoder_raw, struct heif_image** out_im
 
 
   for (int c = 0; c < 3; c++) {
+    if (chroma == heif_chroma_monochrome && c > 0) {
+      break;
+    }
+
```
wantehchang, line 245: Nit: We can use a `num_planes` local variable:

```
  int num_planes = chroma == heif_chroma_monochrome ? 1 : 3;
  for (int c = 0; c < num_planes; c++) {
```

wantehchang (changes requested):

libheif/heif_decoder_aom.cc
0xC0000054: I can change it, but the current code matches what is used in `heif_decoder_dav1d.cc`:
https://github.com/strukturag/libheif/blob/18291ddebc23c924440a8a3c9a7267fe3beb5901/libheif/plugins/heif_decoder_dav1d.cc#L259

farindk:
Thank you.

PR #683, opened 2022-10-12
Replace libavif's CMake options with equivalents

wantehchang:


third-party/dav1d.cmd
```diff
@@ -1,6 +1,6 @@
 : # This install script was originally taken from libavif but might have been modified.
 
-: # If you want to use a local build of dav1d, you must clone the dav1d repo in this directory first, then enable CMake's AVIF_CODEC_DAV1D and AVIF_LOCAL_DAV1D options.
+: # If you want to use a local build of dav1d, you must clone the dav1d repo in this directory first, then enable CMake's WITH_DAV1D option.
```
wantehchang, line 3: @farindk Dirk, Since libheif's WITH_xxx CMake options are ON by default, we don't really need to enable them. So this sentence probably should be rewritten.

farindk:
I have updated the comments in all the download scripts.

PR #684, opened 2022-10-12
Fix a comment error: aom_image_t => Dav1dPicture

wantehchang:
This is a copy-and-paste error (copied from heif_decoder_aom.cc).


fancycode:
Thanks!

PR #687, opened 2022-10-17
Some C++ cleanup

wantehchang:
heif_encoder_aom.cc: In encoder_struct_aom::add_custom_option(), use the erase() method to remove an element from a std::vector.

heif_init.cc: In heif_unregister_encoder_plugin(), pre-increment the iterator, which does slightly less work than the post-increment operator (although compiler optimization is likely to take care of this for us). Also the get() method of std::unique_ptr can be avoided.

### DIFF
--- a/libheif/heif_encoder_aom.cc
+++ b/libheif/heif_encoder_aom.cc
@@ -95,12 +95,9 @@ void encoder_struct_aom::add_custom_option(const custom_option& p)
 {
   // if there is already a parameter of that name, remove it from list
 
-  for (size_t i = 0; i < custom_options.size(); i++) {
-    if (custom_options[i].name == p.name) {
-      for (size_t k = i + 1; k < custom_options.size(); k++) {
-        custom_options[k - 1] = custom_options[k];
-      }
-      custom_options.pop_back();
+  for (auto iter = custom_options.begin(); iter != custom_options.end(); ++iter) {
+    if (iter->name == p.name) {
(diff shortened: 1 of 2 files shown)


libheif/heif_encoder_aom.cc
```diff
@@ -95,12 +95,9 @@ void encoder_struct_aom::add_custom_option(const custom_option& p)
...
 
-  for (size_t i = 0; i < custom_options.size(); i++) {
-    if (custom_options[i].name == p.name) {
-      for (size_t k = i + 1; k < custom_options.size(); k++) {
-        custom_options[k - 1] = custom_options[k];
-      }
-      custom_options.pop_back();
+  for (auto iter = custom_options.begin(); iter != custom_options.end(); ++iter) {
+    if (iter->name == p.name) {
+      custom_options.erase(iter);
```
wantehchang, line 100: Dirk: If the order of elements in the std::vector doesn't matter, the following alternative may be more efficient:

```
  for (size_t i = 0; i < custom_options.size(); i++) {
    if (custom_options[i].name == p.name) {
      std::swap(custom_options[i], custom_options.back());
      custom_options.pop_back();
      break;
    }
  }
```

farindk: Efficiency doesn't matter here. I'd prefer to make sure that the options are in the correct order as sometimes, this is important. E.g. a preset option followed by an option that overwrites a parameter.

PR #690, opened 2022-10-19
heif_encoder_aom: change defaults of speed, min-q

wantehchang:
Change the default of the "speed" parameter from 5 to 6.

Change the default of the "min-q" parameter from 1 to 0 (the default of cfg.rc_min_quantizer in libaom). Also change the range of the "min-q" and "alpha-min-q" parameters from 1..62 to 0..63.

Assert that i is valid for the final "d[i++] = nullptr;" statement.

### DIFF
--- a/libheif/plugins/heif_encoder_aom.cc
+++ b/libheif/plugins/heif_encoder_aom.cc
@@ -187,11 +187,10 @@ static void aom_init_parameters()
   p->version = 2;
   p->name = kParam_speed;
   p->type = heif_encoder_parameter_type_integer;
-  p->integer.default_value = 5;
+  p->integer.default_value = 6;
   p->has_default = true;
   p->integer.have_minimum_maximum = true;
   p->integer.minimum = 0;
-
   if (aom_codec_version_major() >= 3) {
     p->integer.maximum = 9;
   }
@@ -258,11 +257,11 @@ static void aom_init_parameters()
   p->version = 2;
   p->name = kParam_min_q;
   p->type = heif_encoder_parameter_type_integer;
-  p->integer.default_value = 1;
+  p->integer.default_value = 0;
   p->has_default = true;
   p->integer.have_minimum_maximum = true;
-  p->integer.minimum = 1;
-  p->integer.maximum = 62;
+  p->integer.minimum = 0;
+  p->integer.maximum = 63;
   p->integer.valid_values = NULL;
   p->integer.num_valid_values = 0;
   d[i++] = p++;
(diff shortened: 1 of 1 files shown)


libheif/plugins/heif_encoder_aom.cc
```diff
@@ -187,11 +187,10 @@ static void aom_init_parameters()
   p->version = 2;
   p->name = kParam_speed;
   p->type = heif_encoder_parameter_type_integer;
-  p->integer.default_value = 5;
+  p->integer.default_value = 6;
   p->has_default = true;
   p->integer.have_minimum_maximum = true;
   p->integer.minimum = 0;
-
```
wantehchang, line 194: I deleted this blank line because it appears to separate two parameters.

libheif/plugins/heif_encoder_aom.cc
```diff
@@ -324,6 +323,7 @@ static void aom_init_parameters()
   p->has_default = true;
   d[i++] = p++;
 
+  assert(i < MAX_NPARAMETERS + 1);
```
wantehchang, line 326: I used `MAX_NPARAMETERS + 1` in this assertion to match the corresponding array definition:

```
static const struct heif_encoder_parameter* aom_encoder_parameter_ptrs[MAX_NPARAMETERS + 1];
```

We could use the equivalent `assert(i <= MAX_NPARAMETERS);` instead.

libheif/plugins/heif_encoder_aom.cc
```diff
@@ -258,11 +257,11 @@ static void aom_init_parameters()
   p->version = 2;
   p->name = kParam_min_q;
   p->type = heif_encoder_parameter_type_integer;
-  p->integer.default_value = 1;
+  p->integer.default_value = 0;
```
wantehchang, line 260: Dirk: I confirmed with Jingning Han (who leads our libaom team) that setting min-q to 0 or 1 doesn't make a difference unless we are doing lossless encoding. So the original default value of 1 is harmless. But it's better to use libaom's default of 0.

farindk: What does that mean exactly? Q=0 is lossless quantization, so I assume that when in normal (non-lossless) encoding mode, the encoder will never choose Q=0?

wantehchang: Dirk: That's correct. In lossy encoding mode, the encoder will never choose Q=0. So setting the minimum of Q to 0 or 1 doesn't make a difference.

PR #691, opened 2022-10-19
heif_encoder_aom: use all intra mode cfg defaults

wantehchang:
Set cfg.g_lag_in_frames, cfg.kf_mode, cfg.kf_max_dist to the default values of the AOM_USAGE_ALL_INTRA mode. This is useful when aomUsage is set to AOM_USAGE_GOOD_QUALITY or AOM_USAGE_REALTIME.

Also replace 0 with the symbolic constant AOM_USAGE_GOOD_QUALITY for readability.

### DIFF
--- a/libheif/plugins/heif_encoder_aom.cc
+++ b/libheif/plugins/heif_encoder_aom.cc
@@ -838,7 +838,7 @@ struct heif_error aom_encode_image(void* encoder_raw, const struct heif_image* i
   }
 
 
-  unsigned int aomUsage = 0;
+  unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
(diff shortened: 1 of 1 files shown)


farindk:
Thank you for optimizing the aom parameters. This is very helpful.

wantehchang:
Dirk: In the common case, we will use the `AOM_USAGE_ALL_INTRA` mode, and this change is a no-op. This change is helpful only when an older version of libaom that doesn't have `AOM_USAGE_ALL_INTRA` is used, or when `AOM_USAGE_REALTIME` is selected.

Some libaom encoder options are set by calling `aom_codec_control()`. I am going to review the `aom_codec_control()` calls next.

PR #694, opened 2022-10-20
Change AOM_USAGE_GOOD_QUALITY back to 0

wantehchang:
Undo a cosmetic change in
https://github.com/strukturag/libheif/pull/691.

The AOM_USAGE_GOOD_QUALITY macro was not defined in libaom v1.0.0-errata1. heif_encoder_aom.cc apparently tries to support libaom releases older than v2.0.0, so change AOM_USAGE_GOOD_QUALITY back to 0.

### DIFF
--- a/libheif/plugins/heif_encoder_aom.cc
+++ b/libheif/plugins/heif_encoder_aom.cc
@@ -838,7 +838,7 @@ struct heif_error aom_encode_image(void* encoder_raw, const struct heif_image* i
   }
 
 
-  unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
+  unsigned int aomUsage = 0;
 #if defined(AOM_USAGE_ALL_INTRA)
   // aom 3.1.0
   aomUsage = (encoder->realtime_mode ? AOM_USAGE_REALTIME : AOM_USAGE_ALL_INTRA);


silverbacknet:
I'd rather see this as an #if VERSION_PACKED < 0x20000... but I also wonder if there's much value in supporting pre-2.0.0, if it hasn't already come up.

farindk:
I also think we should have an `#if` clause around this so that we can easily migrate to the symbolic name.
I'm not sure whether support for a 1.x AOM is still needed for packing in old Linux distributions.
For Ubuntu, we provide 2.0 AOM packages even back to Ubuntu 14.04:
  https://launchpad.net/~strukturag/+archive/ubuntu/libheif

But there might be other distributions around, still with AOM 1.x.

wantehchang:
Hi Emily and Dirk,

Thank you for your comments. I've edited this pull request so that it is a pure revert. The surrounding code already has rather complicated `#if` clauses:

```
  unsigned int aomUsage = 0;
#if defined(AOM_USAGE_ALL_INTRA)
  // aom 3.1.0
  aomUsage = (encoder->realtime_mode ? AOM_USAGE_REALTIME : AOM_USAGE_ALL_INTRA);
#elif defined(AOM_USAGE_REALTIME)
  // aom 2.0
  aomUsage = (encoder->realtime_mode ? AOM_USAGE_REALTIME : AOM_USAGE_GOOD_QUALITY);
#endif
```

I am not quite sure how you want to add an `#if` to the first line. You can first merge this revert PR and then edit the code as you see fit.

Note: I think it is not necessary to support libaom releases older than v2.0.0. If we can assume libaom is v2.0.0 or later, then the code can be simplified as follows:

```
  unsigned int aomUsage =
#if defined(AOM_USAGE_ALL_INTRA)
    // aom 3.1.0
    AOM_USAGE_ALL_INTRA;
#else
    AOM_USAGE_GOOD_QUALITY;
#endif
  if (encoder->realtime_mode) {
    aomUsage = AOM_USAGE_REALTIME;
  }
```

farindk:
Thank you. I have simplified the `#if` construct according to your suggestion by dropping support for AOM 1.x. It seems that it is save to assume 2.x as the minimum version. If not available, this can be obtained through our PPA (https://launchpad.net/~strukturag/+archive/ubuntu/libheif).

PR #709, opened 2022-11-12
Change "constant QP" to "constant rate factor"

wantehchang:
If rate_control_mode is set to 0 (the default), the rate control mode depends on the value of enable_adaptive_quantization. If enable_adaptive_quantization is 2 (the default), the rate control mode is constant rate factor (CRF). If enable_adaptive_quantization is 0, the rate control mode is constant QP. Since we don't set enable_adaptive_quantization, the rate control mode is CRF.

### DIFF
--- a/libheif/plugins/heif_encoder_svt.cc
+++ b/libheif/plugins/heif_encoder_svt.cc
@@ -624,7 +624,7 @@ struct heif_error svt_encode_image(void* encoder_raw, const struct heif_image* i
   // disable 2-pass
   svt_config.rc_stats_buffer = (SvtAv1FixedBuf) {nullptr, 0};
 
-  svt_config.rate_control_mode = 0; // constant QP
+  svt_config.rate_control_mode = 0; // constant rate factor
   svt_config.qp = encoder->qp;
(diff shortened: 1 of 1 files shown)


libheif/plugins/heif_encoder_svt.cc
```diff
@@ -624,7 +624,7 @@ struct heif_error svt_encode_image(void* encoder_raw, const struct heif_image* i
   // disable 2-pass
   svt_config.rc_stats_buffer = (SvtAv1FixedBuf) {nullptr, 0};
 
-  svt_config.rate_control_mode = 0; // constant QP
+  svt_config.rate_control_mode = 0; // constant rate factor
```
wantehchang, line 627: Note that line 622 is commented out:

```
  //svt_config.enable_adaptive_quantization = 0;
```

Therefore `svt_config.enable_adaptive_quantization` has the default value of 2.

wantehchang (commented):
@farindk @hassount Please review. Thanks.

farindk:
Right. Thank you.

Issue #713, opened 2022-11-16
libheif 1.14.0 fails to build for MinGW


kmilos:
Up until 1.14.0, the build worked pretty much out of the box, but now it fails, see the log here: https://github.com/msys2/MINGW-packages/actions/runs/3473745090/jobs/5806121808#step:11:386

Major: missing symbols when linking the new `libheif-svtenc.dll`

```
Linking CXX shared module libheif\plugins\libheif-svtenc.dll
FAILED: libheif/plugins/libheif-svtenc.dll
cmd.exe /C "cd . && C:\msys64\ucrt64\bin\g++.exe -march=nocona -msahf -mtune=generic -O2 -pipe -DX26
5_API_IMPORTS       -O3 -DNDEBUG  -pipe -shared -o libheif\plugins\libheif-svtenc.dll -Wl,--major-im
age-version,1,--minor-image-version,14 libheif/plugins/CMakeFiles/heif-svtenc.dir/heif_encoder_svt.c
c.obj libheif/plugins/CMakeFiles/heif-svtenc.dir/__/heif_plugin.cc.obj  C:/msys64/ucrt64/lib/libSvtA
v1Enc.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -
ladvapi32 && cd ."
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
 libheif/plugins/CMakeFiles/heif-svtenc.dir/heif_encoder_svt.cc.obj:heif_encoder_svt.cc:(.text+0xb5a
): undefined reference to `heif_image_get_width'
```

... and many more

Minor: unused variable (`-Werror=unused-variable` is default on MSYS2)

```
Building CXX object libheif/plugins/CMakeFiles/heif-svtenc.dir/heif_encoder_svt.cc.obj
FAILED: libheif/plugins/CMakeFiles/heif-svtenc.dir/heif_encoder_svt.cc.obj
C:\msys64\ucrt64\bin\g++.exe -DHAVE_UNISTD_H -DHAVE_VISIBILITY -DLIBHEIF_EXPORTS -DPLUGIN_SvtEnc=1 -
Dheif_svtenc_EXPORTS -IC:/msys64/home/kmilos/MINGW-packages/mingw-w64-libheif/src/libheif-1.14.0 -
[...]

PR #714, opened 2022-11-16
Remove unused variable

kmilos:
Addresses part of https://github.com/strukturag/libheif/issues/713


farindk:
I think I'll add a `(void)input_value;` instead, as the variable is still used in an `assert()` below.

PR #718, opened 2022-11-16
Defer NCLX profile integer-to-enum-value conversion

lovell:
This takes the great work started in commit ffdac1e and extends it to ensure the cast from integer to enum value occurs as late as possible by storing integers in the `known_*` sets.

Without this change, UBSan currently reports the following:
```
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior plugins/heif_decoder_aom.cc:226:3
plugins/heif_decoder_aom.cc:226:3: runtime error: load of value 228, which is not a valid value for type 'aom_color_primaries_t' (aka 'aom_color_primaries')
0xab83f4 in aom_decode_image(void*, heif_image**)
0xa678db in heif::HeifContext::decode_image_planar(unsigned int, std::__1::shared_ptr<heif::HeifPixelImage>&, heif_colorspace, heif_decoding_options const*, bool) const
0xa66417 in heif::HeifContext::decode_image_user(unsigned int, std::__1::shared_ptr<heif::HeifPixelImage>&, heif_colorspace, heif_chroma, heif_decoding_options const*)
0xa39044 in heif_decode_image
```
If required, I can provide an example image with a colour primary value of 228 that triggers this error.

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -1364,7 +1364,7 @@ size_t heif_image_handle_get_raw_color_profile_size(const struct heif_image_hand
 }
 
 
-static const std::set<enum heif_color_primaries> known_color_primaries{
+static const std::set<typename std::underlying_type<heif_color_primaries>::type> known_color_primaries{
     heif_color_primaries_ITU_R_BT_709_5,
     heif_color_primaries_unspecified,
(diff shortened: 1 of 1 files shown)


farindk:
Great catch. Thank you.

PR #719, opened 2022-11-17
Handle relative vs absolute paths for .pc file variables

### LINKED ISSUE #717: Improve pkg-config file generation
[As Autotools is intended to be removed, the CMake generated .pc file needs to be improved for relocatibility](https://github.com/msys2/MINGW-packages/blob/043607c29b4f26550c7cd04ee3be61f340748f13/mingw-w64-libheif/PKGBUILD#L64)  - namely, it should use [`${prefix}` etc. variables when possible](https://people.freedesktop.org/~dbn/pkg-config-guide.html), but still support absolute paths when required.

See e.g. [libjxl](https://github.com/libjxl/libjxl/blob/3f6ec60ce6890132c1e2ea9719de5981042a1d96/lib/jxl.cmake#L642-L663) and [libtiff](https://gitlab.com/libtiff/libtiff/-/blob/31d9e9f73dc8b6fa599c931d9ff058d8ea32795c/cmake/PkgConfig.cmake#L48-63)

Also, there is no need to substitute

```
Libs: -L@libdir@ -lheif
Cflags: -I@includedir@
```

as, being defined correctly at the start, they can be left as

```
Libs: -L${libdir} -lheif
Cflags: -I${includedir}
```

kmilos:
Fixes #717


farindk:
Thank you.

PR #722, opened 2022-11-20
Spelling

jsoref:
I'm following up on a PR I sent to a downstream consumer (and then accidentally to the wrong fork of libheif...).

A few notable bits:

commit | notes |
-|-
9238834373c07c162668c0a7d3962483083c34c2 | `heif_encoder_descriptor_supportes_lossless_compression` is deprecated, but it still appears to be used internally. I've included changes to update the internal uses.
d9dedde4d860cb13c233001826a4a7fc4230b154 | I didn't include deprecation bits here. I did for `supports` to show that I could. I'm not quite sure what the preferences are for this repository, so I'm using this pair as a point for discussion.

---

This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/libheif/commit/65da0de0791f07ce9a7472ff7d150feb70a96b87#commitcomment-90537659

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/libheif/commit/8772000ff91d4c39a3d4111e49964352d20900da

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

### DIFF
--- a/examples/heif_enc.cc
+++ b/examples/heif_enc.cc
@@ -1422,7 +1422,7 @@ int main(int argc, char** argv)
     //heif_image_set_nclx_color_profile(image.get(), &nclx);
 
     if (lossless) {
-      if (heif_encoder_descriptor_supportes_lossless_compression(active_encoder_descriptor)) {
+      if (heif_encoder_descriptor_supports_lossless_compression(active_encoder_descriptor)) {
         heif_encoder_set_lossless(encoder, lossless);
       }
       else {
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -1769,14 +1769,14 @@ int heif_encoder_descriptor_supports_lossless_compression(const struct heif_enco
 }
 
 
-// DEPRECATED: typo in functino name
+// DEPRECATED: typo in function name
 int heif_encoder_descriptor_supportes_lossy_compression(const struct heif_encoder_descriptor* descriptor)
 {
   return descriptor->plugin->supports_lossy_compression;
 }
 
 
-// DEPRECATED: typo in functino name
+// DEPRECATED: typo in function name
 int heif_encoder_descriptor_supportes_lossless_compression(const struct heif_encoder_descriptor* descriptor)
 {
   return descriptor->plugin->supports_lossless_compression;
--- a/libheif/heif_cxx.h
+++ b/libheif/heif_cxx.h
@@ -383,14 +383,19 @@ namespace heif {
 
     enum heif_compression_format get_compression_format() const noexcept;
 
+    // DEPRECATED: typo in function name
     bool supportes_lossy_compression() const noexcept;
 
+    // DEPRECATED: typo in function name
     bool supportes_lossless_compression() const noexcept;
 
 
     // throws Error
     Encoder get_encoder() const;
 
+    bool supports_lossy_compression() const noexcept;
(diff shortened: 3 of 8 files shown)


examples/heif_enc.cc
```diff
@@ -1422,7 +1422,7 @@ int main(int argc, char** argv)
     //heif_image_set_nclx_color_profile(image.get(), &nclx);
 
     if (lossless) {
-      if (heif_encoder_descriptor_supportes_lossless_compression(active_encoder_descriptor)) {
+      if (heif_encoder_descriptor_supports_lossless_compression(active_encoder_descriptor)) {
```
jsoref, line 1425: as noted...

libheif/heif.cc
```diff
@@ -1769,14 +1769,14 @@ int heif_encoder_descriptor_supports_lossless_compression(const struct heif_enco
 }
 
 
-// DEPRECATED: typo in functino name
+// DEPRECATED: typo in function name
```
jsoref, line 1772: I can't tell if this is intentionally cute or not...

libheif/heif_cxx.h
```diff
@@ -284,7 +284,7 @@ namespace heif {
 
     bool is_full_range() const;
 
-    void set_color_primaties(heif_color_primaries cp);
+    void set_color_primaries(heif_color_primaries cp);
```
jsoref, line 287: As noted, I could include the old name as a deprecated version.

C++ classes are brittle which makes things interesting. In some ways, it might make more sense to just force people to update when they upgrade. But, I'm not sure what the repository's policy is for C++...

scripts/cpplint.py
```diff
@@ -3237,7 +3237,7 @@ def CheckOperatorSpacing(filename, clean_lines, linenum, error):
   elif not Match(r'#.*include', line):
     # Look for < that is not surrounded by spaces.  This is only
     # triggered if both sides are missing spaces, even though
-    # technically should should flag if at least one side is missing a
+    # technically it should flag if at least one side is missing a
```
jsoref, line 3240: this is an atypical correction

jsoref (commented):
Most corrections automatically suggested by Google Sheets

libheif/heif_cxx.h
silverbacknet: If you add the old name to the very end, it should not present any C++ ABI or API problem. In this case, it should work even if they're building with mismatched header/libs, since the signatures and output are identical. Just add a line with `// DEPRECATED, typo in function name` like the end of heif.h.

farindk: Since there is no C++ shared library, there is no issue here. It will fail during compilation with old application source, but that is easy to fix. If there were a shared library, we would need ABI stability so that a newer installed library does not break existing binaries.

farindk:
Thank you.
The file `tests/catch.hpp` is generated from this library: https://github.com/catchorg/Catch2
Maybe you want to report the spelling errors in that file there.

alvinhochun:
This change appears to have caused API-breaking changes in a patch release: https://bugs.kde.org/463889

farindk:
Right, that slipped through. I'll do a v1.14.2 which restores the old names as an alternative.

PR #725, opened 2022-11-23
Fix processing of force_screen_content_tools in AVIF

pelle-capture:


farindk:
Thank you for spotting this.

PR #733, opened 2022-12-02
Ensure incoming NCLX enum values are cast to integer before validation

lovell:
This is a follow-up to #718 as the undefined behaviour was still being reported for some (fuzzer-found) inputs. I tracked this down to the use of a cast in the de265 decoder that is missing from the aom and dav1d decoders.

This change ensures all the decoders are consistent in their behaviour, using a `static_cast` for compile-time safety.
```
decoder enum value => uint16 => validate => libheif enum value
```


lovell:
Marking as draft due to dependence on CI fixes in #734.

farindk:
Thank you

PR #737, opened 2022-12-06
refs #635: second attempt to extend API to disable MT decoding

dk:
I propose here to reuse an existing m_max_decoding_threads, that, if set to 0, effectively reverts to the same effect as if the library was compiled without ENABLE_PARALLEL_TILE_DECODING . That setting is context-dependent and allows both threaded and non-threaded encoding to coexist in a single application.

### DIFF
--- a/libheif/heif_context.cc
+++ b/libheif/heif_context.cc
@@ -1547,7 +1547,8 @@ Error HeifContext::decode_full_grid_image(heif_item_id ID,
   };
 
   std::deque<tile_data> tiles;
-  tiles.resize(grid.get_rows() * grid.get_columns() );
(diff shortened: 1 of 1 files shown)


Edi61:
Hi dk,

i have review and checked your code, its principially working, but how to change m_max_decoding_threads?
It's a private variable, where you have no influence from outside.

Regards,
Edi61

dk:
Hi Edi61,

I believe heif_context_set_max_decoding_threads was exposed by f0e38af?

/dk

Edi61:
Ah Ok, sorry, overseen it. So it’s fine.

farindk:
Thank you, this looks good.

PR #743, opened 2022-12-19
cmake: Use correct install path for plugins

TingPing:


TingPing:
I'm confused looking at the code. It seems like `get_plugin_paths()` requires the `LIBHEIF_PLUGIN_PATH` env var and doesn't actually use the value of CMakes `PLUGIN_DIRECTORY`.

farindk:
Thanks. Yes, this makes more sense.

PR #744, opened 2022-12-20
Fix errors in plugin loading

TingPing:
Currently `heif_init()` expects all paths to exist otherwise its a fatal error. So now we always install a directory.


farindk:
Thanks.

PR #750, opened 2022-12-29
added API for mux

roiei:
added API for mux


roiei:
@farindk 
As I told at e-mail, I made a pull request that has implementation for specific APIs. But, I know, it would be better that this PR should be split into several PR/issues. But for discussion, I first made this PR. Please let me get your opinion about that. For example, some part of the PR should be split as another issue, and some part should be discarded as well.

PR #752, opened 2023-01-03
New plugin: Hardware acceleration by FFMPEG hevc decoder

Edi61:
Hello,

to be able to use video hardware acceleration (by dxva, nvdec etc…) i have implemented a first version of a ffmpeg hvec decoding library by using libavcodec from FFMPEG.
I thought first of an own implementation for hardware accelerations,
but it makes no sense to reimplement same things.

So the plugin works in the following way:
- Create a hevc frame out of the information given in the heif file
- Use of libavcodec to decode this hevc frame including hardware acceleration provided by ffmpeg
- Copy decoded YUV data to a new heif_image and add nclx data

With the new plugin i was able to measure speed ups up over 100%, but i have only weak benchmarks.
Due to some CPU profiling there‘s still optmization potential (image transformations, color conversions),
with maybe more then 20-30% speedup, but this would be step 2, if this pullrequest is fine.

Thanks,
Regards


farindk:
This is great.
Please give me some time to review and merge this.

Edi61:
Thanks, no problem.

When doing the review it would be great if you could also look at this functions:
- HeifPixelImage::rotate_ccw
- Op_YCbCr_to_RGB<Pixel>::convert_colorspace
- static inline uint16_t clip_f_u16
- Op_RGB_to_RGB24_32::convert_colorspace

According to CPU Profiler (see picture below) these function are also consuming very high CPU load and it would be great if they could also be replacable by some "plugin" feature. Maybe you have directly an idea.... Otherwise I would make a suggestion after PR is merged.
CPU-Profiler

farindk:
From where can I get the `FFMPEGConfig.cmake` ?

Edi61:
I have used vcpkg to get ffmpeg, so it should be based on this:

https://github.com/microsoft/vcpkg/blob/13224dcaaa80d3453fbe52b9b3dc3dce6f226a27/ports/ffmpeg/FindFFMPEG.cmake.in

Conan-Kudo:
Note that FFmpeg can also do hardware accelerated AV1 too, so if you want to use ffmpeg for this purpose, you might want both. You'll want to probe ffmpeg for codecs at runtime, though.

farindk:
@Edi61 I finally merged this. It's a pity that it took so long, but during my first tries, I always was stuck with FFMPEG detection. I have now found a way that works for me (on Linux). I only did some minor adjustments.

Thank you for your contribution!

Neoclassic:
`./examples/heif-dec  ~/ffmpeg_test/testfiles/LiveOff.HEIC  out.jpg
File contains 1 image
[hevc_cuvid @ 0x744ab8002e40] Invalid pkt_timebase, passing timestamps as-is.
[hevc_cuvid @ 0x744ab0006900] Invalid pkt_timebase, passing timestamps as-is.
[hevc_cuvid @ 0x744aac008c00] Invalid pkt_timebase, passing timestamps as-is.
[hevc_cuvid @ 0x744aa40075c0] Invalid pkt_timebase, passing timestamps as-is.
Could not decode image: Decoder plugin generated an error: Unspecified: avcodec_receive_frame returned EAGAIN or ERROR_EOF`

I am getting this error. Any compilation flags required besides
cmake .. --preset=testing -WITH_FFMPEG_DECODER=ON -WITH_FFMPEG_DECODER_PLUGIN=ON

PR #754, opened 2023-01-06
Fix examples building with pkg-config being present on llvm-mingw

amyspark:
By using CMake's official Find Modules and sidestepping pkg-config altogether, we're able to use imported targets without further complexity.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -15,19 +15,15 @@ set (additional_link_directories)
 set (additional_libraries)
 set (additional_includes)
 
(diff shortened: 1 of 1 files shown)


farindk:
Thank you. By accident, I did much of the same work as in your patch just yesterday.
Could have saved me a lot of time had I seen this earlier :-)

I added the `unset()` lines from your patch (see above).
We cannot yet use the `JPEG::JPEG` syntax, because we still depend on an older cmake version for packaging on ancient Linux versions. Thus, I'll leave your pull request open as a reminder to use this syntax at a later time.

amyspark:
Great to hear @farindk, thanks!

Issue #745, opened 2022-12-20
symbol lookup error in codec plugin


chunronglai:
hi, colleagues:

   As the attached picture shows, I met error when run simple command line as "./heif-enc -A -e rav1e -q 45 ./lake-3000.jpg -o dst2.avif".
   The error info are  "symbol lookup error: ..../libheif-rav1e.so: undefined symbol: _ZNK4heif14HeifPixelImage18get_bits_per_pixelE12heif_channel".
   I also checked it with objdump and does see the symbol in libheif.so instead of in the plugin.
    Anything I should notice or improve during the library build or example run? Thanks a lot.

<img width="656" alt="symbol lookup error" src="https://user-images.githubusercontent.com/80802212/208628216-bf831d52-18ec-43f3-a644-2ab7c96a2548.png">

chunronglai:
I think the error come from some configurations of the plugin building. 
Anyway, the symbol lookup error disappear with below patch:
--- a/libheif/plugins/heif_encoder_rav1e.cc
+++ b/libheif/plugins/heif_encoder_rav1e.cc
@@ -522,7 +522,8 @@ struct heif_error rav1e_encode_image(void* encoder_raw, const struct heif_image*
     rav1eRange = nclx->get_full_range_flag() ? RA_PIXEL_RANGE_FULL : RA_PIXEL_RANGE_LIMITED;
   }

-  int bitDepth = image->image->get_bits_per_pixel(heif_channel_Y);
+  int bitDepth = heif_image_get_bits_per_pixel_range(image, heif_channel_Y);

@@ -607,11 +608,14 @@ struct heif_error rav1e_encode_image(void* encoder_raw, const struct heif_image*
   //} else
   {
     int strideY;
-    const uint8_t* Y = image->image->get_plane(heif_channel_Y, &strideY);
+    const uint8_t* Y = heif_image_get_plane_readonly(image, heif_channel_Y, &strideY);
     int strideCb;
-    const uint8_t* Cb = image->image->get_plane(heif_channel_Cb, &strideCb);
+    const uint8_t* Cb = heif_image_get_plane_readonly(image, heif_channel_Cb, &strideCb);
     int strideCr;
-    const uint8_t* Cr = image->image->get_plane(heif_channel_Cr, &strideCr);
+    const uint8_t* Cr = heif_image_get_plane_readonly(image, heif_channel_Cr, &strideCr);

[...]

PR #764, opened 2023-01-14
Update heif_encoder_rav1e.cc

arjendekorte:
Properly formatted patch for https://github.com/strukturag/libheif/issues/745


farindk:
Thanks.
I have implemented this in 2b3a662, also replacing the other non-public function calls.

PR #769, opened 2023-01-18
pkg-config `Requires.private` when codecs are compiled as plugins

farindk:
I guess that codecs that are compiled as separate plugins should not be included in the pkg-config `Requires.private` field.

@lovell Since you added the `Requires.private` field, could you please confirm that my change is correct?


lovell (approved):
LGTM

farindk:
Thanks for checking.

Issue #731, opened 2022-11-26
plugin on Windows


vtorri:
dlopen --> LoadLibrary
dlclose --> FreeLibrary
dlsym --> GetProcAddress

see
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya
and
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getprocaddress

it should be easy to write a class to embed this for linux and Windows

vtorri:
and you add a completely useless big dependency... Honestly, the calls are almost exactly the same, no difference except the fct names and the types.

farindk:
The implementation is probably easy enough that we can add the system-specific implementation for Windows without adding any dependencies.
libltdl would have the advantage to support even more systems, but HP-UX and BeOS are probably not that significant anymore.
libltdl mentions NSAddImage for macOS, but that seems to be deprecated and should be replaced with dlopen anyway.
I.e. we just need a Windows implementation for now.

kmilos:
For Glib I agree, but libltdl is not that big... In any case, sure, it's simple enough to be done directly if its only Windows. There's even the https://github.com/dlfcn-win32/dlfcn-win32 under MIT license that can be leveraged. It's already packaged for MSYS2 (MinGW) and also available through vcpkg (MSVC). Or just use the source directly, like it's already being done for `getopt`...

Looks like there is a bit more involved in order to properly handle path/separator differences for MSVC/MinGW/Cygwin, errors, etc.

farindk:
@kmilos Thanks, dlfcn-win32 is also a good proposal.

vtorri:
for path separator, \ and / are supported with the win32 API and win32 crt. Also, for the libheif plugin system, i doubt that there will be a lot of problems with path.

farindk:
Implemented with 38ea1ba75c7022318596c1be5bc289f0195a2bfd

PR #770, opened 2023-01-23
Windows plugins

farindk:
Implementation of plugin loading for Windows.


libheif/heif_init.cc
```diff
@@ -22,11 +22,19 @@
 #include "heif.h"
 #include "error.h"
 #include "heif_plugin_registry.h"
+#include "common_utils.h"
 
 #if ENABLE_MULTITHREADING_SUPPORT
 #include <mutex>
 #endif
 
+#if defined(__MINGW32__) || defined(__MINGW64__) || defined(_MSC_VER)
```
kmilos, line 31: Just use `_WIN32` if this covers both MSVC and MinGW (i.e. is related to Windows as a platform and not a specific toolchain)... (This applies to all occurrences below btw.)

OTOH, this seems to be at odds with the `if(UNIX OR MINGW)` CMakeLists.txt above? Perhaps you want to invert the logic to `if(WIN32)`, or use something like `if(UNIX OR CYGWIN OR MSYS)` instead?

libheif/heif_init.cc
```diff
@@ -22,11 +22,19 @@
...
+#include <windows.h>
```
kmilos, line 32: Seems to be included by plugins_windows.h already

libheif/heif_init.cc
farindk: If I understand correctly, MINGW supports both, `dlopen` and `LoadLibrary`, right?
Then using `_WIN32` and `WIN32` in cmake would probably be the easiest variant, right?

kmilos: Nope, there is no dlopen on MinGW out of the box - there is implementations of it like dlfcn and similar wrappers ([like we discussed](https://github.com/strukturag/libheif/issues/731)), but you have to include the source or link another dep to your project (at which point probably MSVC can build it as well, which was the alternative way...)

So if you're doing the pure Win32 API like you decided, then it applies to both MSVC and MinGW equally...

(Cygwin and MSYS are different - they are POSIX **_emulation_** platforms, and ship all the POSIX calls as such... MinGW is "native" Windows)

farindk: Ok, thanks for the background info on this.
Let's try with simple WIN32 and _WIN32.

kmilos: Looking better, thanks! Still missing replacing w/ _WIN32 on this line though...

PR #772, opened 2023-01-28
heif_context: test options before use

lovell:
Avoids a null-dereference in `decode_image_planar()`

The libvips fuzz tests caught this regression in commit https://github.com/strukturag/libheif/commit/e7515bb737423283aa2aad4d6ef74e93d1badea5


farindk:
Right. Thank you.

PR #780, opened 2023-02-08
Remove MIME files

kmilos:
These are included in the freedesktop.org shared mime info for some time now:

https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in

https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/d7ba16d1b8dae1d08b6dfa45b4b619cb1a372539
https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/d23e9fa537b84950cdb518f87e3bf1f3f2577523


farindk:
Could this PR be reverted for 20.04 by including a patch diff for that Ubuntu version only?

fancycode:
> Could this PR be reverted for 20.04 by including a patch diff for that Ubuntu version only?

Done, packages are rebuilding on Launchpad.

PR #783, opened 2023-02-15
Add ICC extraction to gdk-pixbuf loader

kmilos:
See e.g. the default JPEG, [PNG](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/aeda644f9690b346de77c8f7fae68f5661801833/gdk-pixbuf/io-png.c#L357-365) and TIFF loaders.


farindk:
Thank you.

BTW: I plan to do a v1.15.0 release today or tomorrow. Let me know if you have something that should still make it into that release.

kmilos:
> I plan to do a v1.15.0 release today or tomorrow. Let me know if you have something that should still make it into that release.

Don't think so, thanks!

PR #786, opened 2023-02-17
Fix unused variable nPlugins in libheif/plugins_windows.cc

brechtsanders:
Fix error:
```
libheif/plugins_windows.cc:65:7: error: unused variable 'nPlugins'
```


farindk:
Thank you

PR #793, opened 2023-03-05
iloc: fix parsing with version == 1

bradh:
The current ItemLocationBox (iloc) parsing produces incorrect results when `version == 1`.

There are two problems, with the initial parsing of `index_size` which occurs in
```
if ((version == 1) || (version == 2)) {
    unsigned int(4) index_size;
} else {
    unsigned int(4) reserved;
}
```

and then in the extent parsing part:
```
if (((version == 1) || (version == 2)) && (index_size > 0)) {
```

(see ISO/IEC 14496-12:2015 Section 8.11.3.2).

The existing code does > 1, instead of >= 1 for both of those. That causes incorrect parsing of the `extent_offset` and `extent_length`.

The PR attached corrects the first case, and removes the second case (because its redundant - we already skip the case where `index_size` == 0). It would be equally valid to just make them both >= 1 though.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -939,7 +939,7 @@ Error Box_iloc::parse(BitstreamRange& range)
   int base_offset_size = (values4 >> 4) & 0xF;
   int index_size = 0;
 
-  if (get_version() > 1) {
+  if (get_version() >= 1) {
     index_size = (values4 & 0xF);
   }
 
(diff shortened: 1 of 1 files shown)


farindk:
Right. Thank you. Strange that this did not fall onto our feet earlier. Maybe `version=1` is not actively used in practice.

bradh:
Certainly possible `version==1` doesn't have much use. I used it in my (POC) generation code because it looked like the efficient case when values are small, and you want `index` support. I can see that just using `version==2` would be a simpler choice though.

farindk:
I checked this. Actually, libheif also generates `version=1` boxes, but the `index` field is only used for `construction_method=2`, which is never used (it is not even supported). Thus, the error in reading the index field never showed up.

Issue #788, opened 2023-02-21
Libheif 1.15.1 failed to encode image in AVIF format(1.14.2 is ok)


bigcat88:
Sorry for little information, still looking for what the problem is.

Picture modes: Monochrome, RGB and RGBA, 10 bit, regardless of image size.
For the same 12-bit images, everything is ok.

Error: "The encoder plugin generated an error: not specified: Failed to encode frame"

The same tests for version 1.14.2 pass normally.

I'll post more info when I find what the problem is.

bigcat88:
It is failed in my python library and I was able to replicate it without it.
`heif-enc L_16__29x100.png -L -p chroma=444 -A -b 12 -o test.avif`

is ok.

`heif-enc L_16__29x100.png -L -p chroma=444 -A -b 10 -o test.avif`
```Could not encode HEIF/AVIF file: Encoder plugin generated an error: Unspecified: Failed to encode frame```

**Comment by bigcat88** [comment]
```
> heif-enc -A                
 heif-enc  libheif version: 1.15.1

> heif-enc -A --list-encoders

HEIC encoders:
- x265 = x265 HEVC encoder (3.4+31-6722fce1f) [default]
AVIF encoders:
- aom = AOMedia Project AV1 Encoder 3.6.0 [default]
```

File example:
[L_16__29x100.png.zip](https://github.com/strukturag/libheif/files/10824139/L_16__29x100.png.zip)

**Comment by bigcat88** [comment]
MSYS2 updated libaom to `3.6.0` and now on the Windows is the same error:
```
C:\msys64\mingw64\bin>heif-enc RGBA_16__128x128.png -A
Could not encode HEIF/AVIF file: Encoder plugin generated an error: Unspecified: Failed to encode frame
```
_using old libheif(**1.14.2**) **OR** old aom(**3.5.0**) - and it works fine_

**Comment by farindk** [comment]
@wantehchang Wan-Teh Chang, do you know what could have changed in AOM 3.6.0 ?

**Comment by wantehchang** [comment]
@farindk Dirk: I will debug this today.

**Comment by wantehchang** [comment]
Dirk: The bug was introduced in commit 135f4ba0e2f13bedff4c81fc62bb49ae4bc61181. Things go wrong in the following code in libheif/plugins/heif_encoder_aom.cc:

```
  int seq_profile = compute_avif_profile(heif_image_get_bits_per_pixel(image, heif_channel_Y),
                                     heif_image_get_chroma_format(image));
```

[...]

PR #798, opened 2023-03-20
heif_encoder_aom.cc: Get image bit depth correctly

wantehchang:
Call heif_image_get_bits_per_pixel_range(), not
heif_image_get_bits_per_pixel(), to get the bit depth of the image. The bug was introduced in commit 135f4ba0e2f13bedff4c81fc62bb49ae4bc61181.

Fix https://github.com/strukturag/libheif/issues/788.


farindk:
Thank you.

Issue #587, opened 2022-01-19
Add support for sharp yuv


maryla-uc:
As a follow up to #586 (smooth chroma upsampling), a second (bigger) step would be to add support for "sharp yuv", a slower rgb->yuv420 conversion technique that minimizes arifacts caused by chroma subsampling. The algorithm assumes that the image is decoded with bilinear upsampling.

It was first implemented in webp, see an intro/example here https://www.ctrl.blog/entry/webp-sharp-yuv.html
libwebp code is here: https://github.com/webmproject/libwebp/blob/36a6eea3bc2871c7f0166193a6fa42241af4a464/src/enc/picture_csp_enc.c#L447

Would you consider adding it as an optional subsampling method to libheif? If so, would you rather have it be reimplemented in libheif, or use an implementation from an external library, if available? (we are considering adding it to libyuv or possibly creating a separate library).

farindk:
Do you know which algorithm is used by SharpYUV ? I don't think it is simple bilinear upsampling. Maybe bicubic or Lanczos, or something else?

maryla-uc:
SharpYuv is a subsampling technique but it assumes the upsampling is bilinear (you can see it here: https://github.com/webmproject/libwebp/blob/36a6eea3bc2871c7f0166193a6fa42241af4a464/src/dsp/yuv.c#L229 )

SharpYuv is an iterative process. Basically, it starts with the naive subsampled version (first candidate), upsamples it with bilinear the way it expects the decoder to do, and subsamples it again. It checks how far the result is from the target (= subsampled original), shifts the candidate by that much, and does another iteration.

This is a bit of a simplification but it's the gist of it. The fact that it's iterative means it's slower than regular subsampling. This is why it's not enabled by default in cwebp, although with more complex image formats/slower encoders, the overhead might be less noticeable.

farindk:
Understood, thanks for the explanation.

Sounds like this could be done with a large matrix inversion. Since it's an overdetermined system, one might look for a least squares solution.
Maybe the webp code is just doing that iterative inverse. I didn't have a close look yet.

tongyuantongyu:
I did some work on porting sharp yuv to libavif (https://github.com/AOMediaCodec/libavif/pull/444), and I'd like to share some information here.

> this could be done with a large matrix inversion

[...]

PR #799, opened 2023-03-21
Add support for 'sharp' RGB to YUV420 conversion for 8 bit images.

maryla-uc:
The 'sharp' RGB to YUV conversion algorithm improves the visual quality of sharp edges when viewing 420 images with bilinear chroma upsampling.

Caveats: it's slower than normal RGB to YUV conversion, and it can make images look worse if nearest neighbor upsampling is used at decoding time.

Uses libsharpyuv (from libwebp). It's disabled by default.

Issue #587


farindk:
I have merged this into a separate 'sharpYUV' branch: https://github.com/strukturag/libheif/tree/sharpYUV

Before we can merge it into the master branch, there are still some issues required to be solved:
- Currently, sharpYUV is only used for 8bit 4:2:0 images when the input is interleaved RGB. It is not used yet for planar RGB, chroma 4:2:2, or HDR images. Hence, the option "enable-sharp-yuv" is a bit misleading as it is not guaranteed that this is really used. Can we use the sharpYUV algorithm also for the other cases?
- We should have bilinear YUV->RGB conversion that matches the sharpYUV conversion.
- Finally, I'm not sure yet about the best way to decide which algorithm to use. Ideally, that decision should be carried out automatically by the conversion pipeline planning algorithm. However, as it is currently implemented as a flag that switches the behavior of a conversion stage and at the same time even might have a silent fallback to another algorithm, this seems to fight against the decision making system. I would prefer sharpYUV be in a separate conversion step and let the planning algorithm make the decision. However, this is not possible at the moment because we currently cannot feed user choices into the planning cost function.

maryla-uc:
Hi Dirk, thank you for reviewing the PR and creating that branch!
I agree with you that this is a partial implementation and it makes sense to wait for fuller support before releasing it.

Regarding full support:
- planar RGB is supported by libhsarpyuv
- 422 is not currently supported, it was considered low priority as it's rarely if ever used for images, but it could be reprioritized if needed
- higher bit depths (I assume that's what you mean by HDR?) are supported: 10/12/16 bits for the input rgb, and 10/12 for the output yuv. The caveat is that libsharpyuv expects a buffer of uint16_t values as input, i.e. it assumes that the input uses the platform's native endianness. This is unlike libheif that manually handles endiannness (I'm not sure why this is needed btw?). To use libhsarpyuv as is, libheif would need a function to determine if the image's current endiannness matches the platform's, and if not, swap the bytes to change the endiannness. When I realized this, I decided to start small with a PR that only deals with 8bit. Other options would be to make libsharpyuv more flexible, or to change the way libheif handles endianness. 

I also understand your concern about choosing to use sharpyuv or not. One tricky thing is handling failures, even though they should only happen if bad values are passed in(e.g. an unsupported bit depth, a zero or negative size, a null input our output pointer), or if a malloc fails (out of memory). Do you think the whole encoding process should fail in this case rather than falling back to a simpler conversion?

farindk:
Hi Maryla,

endianness is probably no problem. If we have a color conversion stage and that, for example, only accepts little endian, but the input is big endian, the planning algorithm will automatically call `Op_RRGGBBaa_swap_endianness`. You just have to specify in `state_after_conversion()` what inputs are accepted and what the output will then look like.

In the same way, if there is no planar color conversion available, it will automatically first convert to interleaved, then do the conversion and then convert back again. That's inefficient, but better than nothing.
But here is the problem with the current partial sharpYUV implementation: when this is enabled, the planning should always go through the sharpYUV algorithm and not find a workaround to do the conversion without. I am still not sure about the best method. Maybe the nearest-neighbor Ops should simply be removed from the planning when sharpYUV is selected.

> Do you think the whole encoding process should fail in this case rather than falling back to a simpler conversion?

I am not sure, but probably yes. It seems that the only case that we cannot handle is 4:2:2 and that will consistently fail. Out of memory is also a valid reason to fail and should not lead to the software just doing something different instead.

farindk:
I guess we can close this in favor of #813 , right?

maryla-uc:
Yes, since this was merged in a the sharpYUV branch, this PR can be closed.

PR #803, opened 2023-03-28
Uncompressed Codec Placeholder

dukesook:
MPEG is working on ISO/IEC 23001-17 which defines the uncompressed codec. Add it to the list of compression formats to be implemented in the future.

### DIFF
--- a/libheif/heif.h
+++ b/libheif/heif.h
@@ -960,7 +960,8 @@ enum heif_compression_format
   heif_compression_AV1 = 4,
   heif_compression_VVC = 5,
   heif_compression_EVC = 6,
-  heif_compression_JPEG2000 = 7  // ISO/IEC 15444-16:2021
+  heif_compression_JPEG2000 = 7,  // ISO/IEC 15444-16:2021
+  heif_compression_UNCOMPRESSED = 8 // ISO/IEC 23001-17:2023
 };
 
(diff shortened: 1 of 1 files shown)


farindk:
@dukesook Thank you for mentioning this. I had a look at the draft text.

@bradh Thanks for the proposal. The issue with this is that we cannot pass C++ structures into the codec plugins because that would induce an ABI dependency and plugins could not be compiled independently from the main library.

The uncompressed codec seems to be quite different from other codecs in the sense that it uses a lot of custom MP4 boxes to describe its contents. Because of this and because it is comparatively easy to implement, I suggest that we simply add this codec into the main library instead of a plugin.

Are you aware of any sample images that could be used for testing while implementing this?

bradh:
@farindk I'm not aware of any samples that are authoritative, or even partly validated.  I do have some I've generated (with a separate implementation). The big gap in my generated samples is nothing that is tiled yet, and limited padding / blocking options. If you'd like to see them, ping me an email at bradh at frogmouth dot net (with hopefully obvious replacement) and I can send you what I have.
 
The implementation would probably work better internally - I had not considered that. Making a plain C struct for each box would also be a possibility,  but the main library would be probably need to be split into a box-parsing layer and an image producing layer to allow the plugin thing to work well.

PR #804, opened 2023-03-30
Work in progress on uncompressed codec

bradh:


libheif/box.cc
```diff
@@ -1003,12 +1013,14 @@ Error Box_iloc::parse(BitstreamRange& range)
     for (int e = 0; e < extent_count; e++) {
       Extent extent;
 
-      if (index_size == 4) {
-        extent.index = range.read32();
-      }
-      else if (index_size == 8) {
-        extent.index = ((uint64_t) range.read32()) << 32;
-        extent.index |= range.read32();
+      if (get_version() >= 1 && index_size > 0) {
```
farindk, line 1016: I guess that this is accidentally reverting commit a11e10cab275097971089b227e23f8dcc5ca3f93.
Please remove this from the PR or explain why you reverted the above commit.

bradh: Indeed accidental. My bad.

farindk:
Looks good to me. We could merge this basically without change.

Two things:
I would prefer the lowercase `heif_compression_uncompressed` instead of `heif_compression_UNCOMPRESSED` because all enum constants in libheif are lowercase except for the abbreviations.

To structure the growing code base, I would move all boxes IO for the uncompressed codec into a separate file. But I can do that later myself. I didn't make up my mind yet whether it would be better to sort the code by decoding phase (directory with all box decoders) or by task (directory for uncompressed codec, directory for region descriptors). I have a tendency towards the latter.

bradh:
I have switched the enum to lower case.

I did not do the box move - it likely requires other refactoring, and that isn't a job for tonight.

farindk:
Thank you. I'll do the move to a separate directory with everything related to 'uncompressed' images.

PR #805, opened 2023-04-02
uncompressed: draft work on uncompressed decoder

bradh:
@farindk Is this the kind of approach you had in mind for the "internal" implementation?

Clearly this is not complete, and will need significant cleanup, just looking for comments on the tie-in.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ plugin_option(AOM_DECODER AOM "aom" "AVIF decoder" OFF)
 plugin_option(SvtEnc SvtEnc "Svt-av1" "AVIF encoder" ON)
 plugin_option(RAV1E RAV1E "Rav1e" "AVIF encoder" ON)
 
+option(ENABLE_UNCOMPRESSED_DECODER "Support internal ISO/IEC 23001-17 uncompressed decoder" OFF)
 
 # --- Create libheif pkgconfig file
 
--- a/libheif/CMakeLists.txt
+++ b/libheif/CMakeLists.txt
@@ -95,6 +95,10 @@ if (ENABLE_MULTITHREADING_SUPPORT)
     endif()
 endif()
 
+if (ENABLE_UNCOMPRESSED_DECODER)
+    target_compile_definitions(heif PRIVATE ENABLE_UNCOMPRESSED_DECODER=1)
+endif()
+
 write_basic_package_version_file(${PROJECT_NAME}-config-version.cmake COMPATIBILITY ExactVersion)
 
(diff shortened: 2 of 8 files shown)


bradh:
One thing I hit while testing this on a grid image was:

```
Could not decode image: 400: Invalid input: Wrong tile image chroma format: Image tile has different chroma format than combined image
```

The `grid` implementation appears to assume planar (or at least, that seems to be what comes down from the heif-convert tool), and my test file uses RGB pixel interleave. 

Would it be better to make the uncompressed codec always output planar RGB(A) given any kind of R/G/B/A input file (planar / pixel interleave / row interleave, channel order, sample size, etc)? Or should we implement conversion?

farindk:
That's a difficult question. The history is that in the beginning, I allowed any kind of image format for the tiles (in my case YCbCr with different chroma formats). Then, issues appeared what would happen if different tiles used different chroma formats, or when tiles would have odd dimensions. Also other parts of the decoder run into problems, like how should we turn a 4:2:2 image by 90 degrees?

Thus, I decided to use 4:4:4 and preferably RGB internally to avoid these problems.
Now, MIAF seems to put further restrictions to HEIF to avoid these cases.

For the moment, I think it would be easiest to just use planar RGB, but I might relax that restriction in the future.
I don't think that we should have native formats for all options (row interleave, ...) because in my opinion, these add complexity without enough benefit.

bradh:
> For the moment, I think it would be easiest to just use planar RGB, but I might relax that restriction in the future. I don't think that we should have native formats for all options (row interleave, ...) because in my opinion, these add complexity without enough benefit.

I can work with that. 

Can I get feedback on the broader approach (as in, how it is tied in)? Is that roughly what you had in mind?

farindk:
Yes, it is ok. I would just move all the "uncompressed"-related functions out of HeifContext into a separate file or even a separate directory so that HeifContext/HeifFile will not grow too much. They have to be refactored anyway in the future.

If you want, we can merge this now and I'll help with the refactoring, or we do it later. Just as you prefer.

bradh:
I have some of the pixel-interleave to planar cleanup done, and will push that soon.

bradh:
@farindk please refactor as you see fit, including any squashing of commits.

farindk:
Beware: I'm going to move things around a bit now. Probably you'd like to wait until I am finished with this.

farindk:
Ok, that's it for the moment. I hope I didn't mess up too much :-)

farindk:
I have written the framework for adding an uncompressed encoder. No actual encoding code yet.

PR #808, opened 2023-04-03
uncompressed: additional component types

bradh:
Also renames Component_x to Component_Type_x to allow easier distinguish from Components

Also removes an obsolete comment on component ordering.


farindk:
On second thought: in the future, we are probably going to put `ComponentType` into the public API. Hence, it would make sense to already use the naming scheme of the public API for this:
`heif_component_type` with entries `heif_component_type_red` ...

bradh:
I will follow up on this later today.

PR #809, opened 2023-04-04
uncompressed: add tiling support for planar and interleaved images

bradh:
This enables correct rendering of the  `uncC_2_rgb_planar_tiled.heif` and `uncC_3_rgb_interleaved_tiled.heif` samples.

### DIFF
--- a/libheif/uncompressed_image.cc
+++ b/libheif/uncompressed_image.cc
@@ -19,6 +19,7 @@
  */
 
 
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


farindk:
Great. I've tested it. Yes, works on all 4 example images.

bradh:
I wondered about adding the decoding (and samples) to CI, but since this is in draft, and there are potentially changes before it gets to release, probably better to keep it offline for now.

PR #813, opened 2023-04-05
Refactor sharp yuv conversion to be a separate conversion step.

maryla-uc:
If sharp yuv is requested but is not supported or fails, the whole encode fails rather than falling back to regular RGB->YUV conversion.

Reverts changes to Op_RGB24_32_to_YCbCr and adds
Op_RGB24_32_to_YCbCr_Sharp instead.


farindk:
Thank you. I thought about the conversion options again and replaced the `enable_sharp_yuv` boolean with an enum to select the preferred chroma subsampling algorithm and a boolean to force using this preferred algorithm. This is still open for discussion if you have another idea.

Issue #726, opened 2022-11-23
Remove autoconf / automake files


fancycode:
Now that the cmake build system is the new default and supports more features, we should get rid of the deprecated autoconf / automake build files.

@farindk what do you think?

farindk:
I think we should still leave it in the repository for some time so that everyone has some time to adapt.
I'll add a deprecation warning to `configure`.

maryla-uc:
Note that some of the scripts in https://github.com/strukturag/libheif/tree/master/scripts and the oss fuzz build script at https://github.com/google/oss-fuzz/blob/master/projects/libheif/build.sh need to be updated before autoconf/automake can be removed.

farindk:
I've replaced the autotools with cmake in all the CI tests and OSSFuzz.

OSSFuzz fails for the "AFL" fuzzer with
````
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x25
````
https://github.com/google/oss-fuzz/actions/runs/4818842628/jobs/8594181803?pr=10177

I can reproduce this locally, but I have not yet figured out how to fix this.

farindk:
I think I fixed the AFL fuzzer. Still waiting for the OSSFuzz CI run.

bradh:
The .gitignore content still needs update. Which entries should remain depends on whether we want to support in-tree cmake builds. Any thoughts on that?

farindk:
I see that the current CI scripts do cmake in-tree builds, but I would not recommend that because it will clutter files everywhere. I just got a similar problem with the CPack configuration that was including all of those clutter files.

bradh:
Is anything still required on this?

farindk:
Seems like we can close it.

PR #814, opened 2023-04-05
Add c++ tests to cmake.

maryla-uc:
Will help with removing autoconf/automake #726

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,7 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/libheif.pc
 # ---
 
 option(WITH_EXAMPLES "Build examples" ON)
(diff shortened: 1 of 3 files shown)


maryla-uc:
Mh the fuzzer shows as failed but the error is just "This request was automatically failed because there were no enabled runners online to process the request for more than 1 days"

Note that I have disabled the tests on windows (they're on by default otherwise) as I didn't manage to make them build. They're on by default on other platforms.

Also note that the tests depend on a new library, heif_internal, which exports all symbols (so that tests can use functions that are not part of the public API). This seems cleaner than modifying the way the main libheif lib is built when tests are enabled, as is done in the automake file. It does slow down the first compile quite a bit though.

farindk:
Thank you

Issue #590, opened 2022-01-23
Stack Overflow in libheif


CarlosAndresRamirez:
Hello libheif team,

I have found an issue in libheif that severely impacts the availability and potentially the integrity (depending on context) of software systems that utilize the library. As the issue triggers a stack overflow on some systems, code execution could be possible.

Issue has been tested with the "heif-convert" sample program. Other programs such as Google Photos in Android enter into an infinite loop.

**ISSUE:**
Specially crafted .heic files can trigger a stack-based overflow at get_id_of_non_virtual_child_image() in heif_context.cc.

**PoC:**
1. [https://github.com/CarlosAndresRamirez/PoCs/blob/main/libheif-PoC-Stack-Overflow-02.heic](https://github.com/CarlosAndresRamirez/PoCs/blob/main/libheif-PoC-Stack-Overflow-02.heic)
2. [https://github.com/CarlosAndresRamirez/PoCs/blob/main/libheif-PoC-Stack-Overflow.heic](https://github.com/CarlosAndresRamirez/PoCs/blob/main/libheif-PoC-Stack-Overflow.heic)

>  ./heif-convert libheif-PoC-Stack-Overflow.heic output.png                                                                                                                                           
File contains 1 images
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2081262==ERROR: AddressSanitizer: stack-overflow on address 0x7fffff7feff8 (pc 0x0000004c2fb5 bp 0x7ffff5690000 sp 0x7fffff7ff000 T0)
    ...
SUMMARY: AddressSanitizer: stack-overflow
==2081262==ABORTING
>

Thanks,
Carlos Andres Ramirez
Security Engineer

PR #815, opened 2023-04-06
Add image reference cycle detection

fairay:
Resolves  https://github.com/strukturag/libheif/issues/590.

Some derived images may contain cycles of image references in iref. Decoding such images leads to an out of memory error (which in turn may cause a DoS).

This PR adds an iref check, which detects cycles during image parse.


farindk:
Thank you.
I've changed it slightly to hide the `parent_items` parameter, which is an implementation detail.

fairay:
Thanks for merging this, @farindk 
Could you, please, tag a release which includes this fix?

PR #802, opened 2023-03-25
rgan: initial work on Region Annotation

bradh:
This is initial work on supporting parsing of RegionItem. It is the first substantive change I'm proposing for libheif, and wanted to get feedback before investing too much in the wrong direction.

Region items and region annotations are described in ISO/IEC 23008-12:2022 Section 6.10. AFAICT, it is not present in HEIF 2017 edition.

The current scope I am targeting is parsing and info display. I'm not planning writing at this stage. 

An example of what it might look like in an info display is:

```
MIME type: image/heif
main brand: mif1
compatible brands: mif1, mif2, ns01, unif

image: 1280x720 (id=6007), primary
  color profile: no
  alpha channel: no 
  depth channel: no
metadata:
  none
region annotations:
  id: 6024, reference_width: 1280, reference_height: 720, 2 regions
    Rectangle Region, x: 320, y: 480, width: 320, height: 240
    Point Region, x: 480, y: 600
```

This is a WIP. Remaining tasks include:

  - [ ] Support for polyline
  - [ ] Support for polygon
  - [ ] Support for ellipse
  - [ ] Support for inline mask (at least flagging it is not supported)
  - [ ] Support for mask item

Any additional items and general contribution guidance (e.g. recommended code formatters) would be appreciated.

PR #816, opened 2023-04-07
Regions

bradh:
@farindk This is a port of #802 to the proposed region API. There are some changes from master branch - only the last commit in the sequence is new.

I had to change the API in a few aspects - some are trivial (like signed vs unsigned), but some are more extensive, and possibly not desired.  In particular, I had trouble mapping from the C++ class to the C API without context. That led to this:

```
 LIBHEIF_API
-int heif_region_item_get_number_of_regions(const struct heif_region_item* region_item);
+long unsigned int heif_region_item_get_number_of_regions(const struct heif_image_handle* handle,
+                                                         const struct heif_region_item* region_item);
 
 LIBHEIF_API
-int heif_region_item_get_list_of_regions(const struct heif_region_item* region_item,
-                                         struct heif_region* regions, int max_count);
+long unsigned int heif_region_item_get_list_of_regions(const struct heif_image_handle* handle,
+                                                       const struct heif_region_item* region_item,
+                                                       struct heif_region* regions,
+                                                       long unsigned int max_count);
 
 LIBHEIF_API
 int heif_region_item_get_number_of_annotations(const struct heif_region_item* region_item);
@@ -1768,10 +1783,16 @@ LIBHEIF_API
 enum heif_region_type heif_region_get_type(const struct heif_region* region);
 
 LIBHEIF_API
-struct heif_error heif_region_get_point(const struct heif_region* region, int32_t* x, int32_t* y);
+struct heif_error heif_region_get_point(const struct heif_image_handle* handle,
+                                        const struct heif_region_item* region_item,
+                                        const struct heif_region* region, int32_t* x, int32_t* y);
 
 LIBHEIF_API
-struct heif_error heif_region_get_rectangle(const struct heif_region* region, int32_t* x, int32_t* y, int32_t* width, int32_t* height);
+struct heif_error heif_region_get_rectangle(const struct heif_image_handle* handle,
+                                            const struct heif_region_item* region_item,
+                                            const struct heif_region* region,
+                                            int32_t* x, int32_t* y,
+                                            uint32_t* width, uint32_t* height);
```

That doesn't seem very satisfactory, and will be worse again with the annotations. The alternative would be to have the `heif_region_item` and `heif_region` keep some kind of opaque context, and reinterpret that as a pointer to the parent. I didn't have a sound approach to ensuring it was safe though. 

If you have any improvements, would be grateful for whatever you can offer.


farindk:
@bradh I thought that `heif_region_item` could be implemented similar to `heif_image_handle` (see in `heif_api_structs.h`).
I can do the change if you want. If I understood you correctly, I just have to cherry-pick 61e0f06 for this, right?
Let me know when I can do this, I want to avoid that we both work on the same code concurrently.

bradh:
I'm currently working on some related parts in my separate (java) implementation, so would appreciate you working on this as you have the time available.

farindk:
I've reworked the regions API. The extra parameters are gone now. I've also changed back the unsigned integers to plain `int`. That may be a matter of taste, but I think that the full range will never be needed and unsigned calls for a lot of problems, like simple for loops counting backwards `while (i>=0)`.

I've also temporarily removed the annotations API because I think that this is easier done in a unified way not differentiating between images and regions (`udes` can be assigned to both).

bradh:
Thanks.

PR #817, opened 2023-04-07
udes: implementation for user description property

bradh:
This supports the implementation of region annotations.

See ISO/IEC 23008-12:2022(E) Section 6.5.20.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -517,6 +517,10 @@ Error Box::read(BitstreamRange& range, std::shared_ptr<heif::Box>* result)
       box = std::make_shared<Box_mdcv>(hdr);
       break;
 
+    case fourcc("udes"):
+      box = std::make_shared<Box_udes>(hdr);
+      break;
+
 #ifdef ENABLE_UNCOMPRESSED_DECODER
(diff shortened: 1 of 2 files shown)


bradh:
See https://github.com/nokiatech/heif/raw/gh-pages/content/heifv2/lights_1440x960.heic for a sample that uses this.

farindk:
Thank you. Works fine.

I'll think of an API to access these.

farindk:
I have added an API for reading the `udes` properties. Please check if you find some problems with this.

Note that I have deviated from the proposal in the regions branch, because in the regions branch, the user-description property was defined specifically for regions, which is not ideal. I guess it makes more sense to access them by `heif_item_id` such that it works with both images and regions.

bradh:
I think the `heif_item_id` is a good choice. I just noticed https://github.com/nokiatech/heif_conformance/blob/master/conformance_files/C050.heic has udes, which is used with the `albc` (Album Collection) entity group. That is, it's a way to put a name / description / tags on the album. 

(That may require some more work with group handling to realise, but the API seems workable for that case).

PR #820, opened 2023-04-13
JPEG2000 Encoder

dukesook:
While there are still many TODOs, here is a simple yet functional JPEG2000 encoder. 

I'm starting with the encoder since no sample files exist other than the ones created by this initial encoder. I publicly shared some [sample files](https://github.com/dukesook/enoji/tree/main/images) which are being verified by the JPEG2000 community. 

[MP4Box.js](https://gpac.github.io/mp4box.js/test/filereader.html) can now parse the item 'j2kH' property. 

I'm using the OpenJPEG library which can be installed on Linux with:
>$ sudo apt install libopenjp2-7-dev 

It's worth noting that HTJ2K is so similar to JPEG2000 that it might get implemented in the same file.


farindk:
Thank you. I have merged this into the branch `jpeg2k`. I suggest that we continue development there until we can merge it into `master`.
I will close this. You can then create a new PR with the future changes against `jpeg2k`.

PR #821, opened 2023-04-17
Improve sharp yuv support and tests.

maryla-uc:
Add sharp yuv support for planar rgb and higher bit depths.

Improve tests: check that output image has the expected planes/bit depths etc., and check psnr after roundtrip when possible. 

Fix makefile build.

### DIFF
--- a/libheif/Makefile.am
+++ b/libheif/Makefile.am
@@ -61,6 +61,22 @@ endif
 libheif_la_SOURCES = \
   bitstream.h \
   bitstream.cc \
+  color-conversion/colorconversion.cc \
+  color-conversion/colorconversion.h \
+  color-conversion/rgb2yuv.cc \
+  color-conversion/rgb2yuv.h \
+  color-conversion/rgb2yuv_sharp.cc \
+  color-conversion/rgb2yuv_sharp.h \
+  color-conversion/yuv2rgb.cc \
+  color-conversion/yuv2rgb.h \
+  color-conversion/rgb2rgb.cc \
+  color-conversion/rgb2rgb.h \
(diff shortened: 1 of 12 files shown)


farindk:
Thank you. Very good idea to convert the alpha plane on-the-fly with your `has_alpha` / `wants_alpha`. I'm just wondering whether it is more efficient to simply add a complete (planar) alpha plane afterwards (for the case `has_alpha=false, wants_alpha=true`), since that can use a simple `memset(...0xFF...)` to fill the alpha plane instead of doing it byte by byte.

farindk:
I have fixed the copying of the input decoding options. The background is that the user code may be coded to a lower version of libheif. Hence, we first fill the options with all defaults for the current version and then overwrite those options that have been set by the user (which may be a lower version).

maryla-uc:
> I have fixed the copying of the input decoding options. The background is that the user code may be coded against a lower version of libheif. Hence, we first fill the options with all defaults for the current version and then overwrite those options that have been set by the user (which may be a lower version).

That makes sense, thank you!

> Thank you. Very good idea to convert the alpha plane on-the-fly with your `has_alpha` / `wants_alpha`. I'm just wondering whether it is more efficient to simply add a complete (planar) alpha plane afterwards (for the case `has_alpha=false, wants_alpha=true`), since that can use a simple `memset(...0xFF...)` to fill the alpha plane instead of doing it byte by byte.

Yes you are right, I made this change because I noticed a discrepancy between the comments that said "Note: no input alpha channel required. It will be filled up with 0xFF", and the actual behavior, but when adding planard alpha it can definitely be optimized.

PR #822, opened 2023-04-19
regions: add polyline to regions API

bradh:
This proposes two somewhat related changed to the region API.

The first is to represent polyline as distinct from polygon in the API. That avoids the need for checking if its closed, and more closely represents the way the standard describes it (and hopefully the way the user thinks about it). The underlying implementation is still shared - the API change just avoids exposing that implementation detail.

The second change is to switch both polygon and polyline to returning `heif_error`, rather than just not filling in any data in the points array. That should be checked, and also makes the API more consistent with the equivalent calls for point, rectangle and ellipse.

This should be seen as a proposal towards a final API, still subject to discussion.

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -3224,7 +3224,8 @@ struct heif_error heif_region_get_ellipse_scaled(const struct heif_region* regio
   return heif_error_invalid_parameter_value;
 }
 
-int heif_region_get_polygon_num_points(const struct heif_region* region)
+
+static int heif_region_get_poly_num_points(const struct heif_region* region)
 {
   const std::shared_ptr<RegionGeometry_Polygon> polygon = std::dynamic_pointer_cast<RegionGeometry_Polygon>(region->region);
   if (polygon) {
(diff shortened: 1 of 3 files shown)


farindk:
Yes, makes sense. Thanks.

I'm wondering whether we need all the error return values since basically nothing can go wrong in the `heif_region_get_*" functions except the user passing wrong arguments. Checking all the error values can become tedious.
On the other hand, getting some indication of user error is better than silently ignoring it.

Maybe some comment in the API would do that the error check can be omitted if one is sure that the parameters are ok.
But that's also dangerous...

bradh:
> Yes, makes sense. Thanks.
> 
> I'm wondering whether we need all the error return values since basically nothing can go wrong in the `heif_region_get_*" functions except the user passing wrong arguments. Checking all the error values can become tedious. On the other hand, getting some indication of user error is better than silently ignoring it.

I see it as something that you can check if you are paranoid or trying to debug something. It is at least possible.

> Maybe some comment in the API would do that the error check can be omitted if one is sure that the parameters are ok. But that's also dangerous...

I don't think most users will need to be told not to check errors - that'll come pretty easily :-)

farindk:
Ok, we leave it as it is now.

BTW: I did not check the `*_scaled` functions. I did some hacking to modify the `reference_width` sizes and check that this works, but I have not checked all the transformations yet.

PR #823, opened 2023-04-20
region: add referenced mask support

bradh:
This extends the region item support to support a referenced mask image as the region geometry.

It splits up the referenced mask item from the inline mask, partly because there is no parsing support yet for inline, but also because we might need a slightly different approach - the referenced item isn't readily available from just the region info.

It also includes partial support for the mask configuration property (`mskC`), not yet exposed in API.


libheif/heif.h
```diff
@@ -1934,6 +1935,11 @@ struct heif_error heif_region_get_polyline_points_scaled(const struct heif_regio
                                                          double* pts,
                                                          heif_item_id image_id);
 
+LIBHEIF_API
+struct heif_error heif_region_get_referenced_mask(const struct heif_region* region,
+                                                  int32_t* x, int32_t* y,
+                                                  uint32_t* width, uint32_t* height);
+
```
farindk, line 1942: I think this should also return the ID of the referenced mask image. If I understand the standard correctly, we would get the ID from the 'iref' box.
Do we have an example image?

bradh: I do not have a test image, and am working on creating some.

Will extend the API and implementation to provide the item id.

bradh: I have added the API and implementation, along with special case handling for when width and height are 0 in the property (which flags that we need to get the values from the referenced item `ispe`).

Still need to make representative test data.

bradh:
Closing this because of conflicts. A replacement is at #845

PR #826, opened 2023-04-23
uncompressed: fix decode when image width is not equal to stride

bradh:
I had not realised that stride was rounded up from actual width.

This PR also includes a new test suite that verifies the solution, and should help with avoiding regression. The test files were generated with https://github.com/gpac/gpac/wiki/uncvg


farindk:
Thank you.

PR #828, opened 2023-04-24
Test all valid src/dst color state combinations in color test.

maryla-uc:
Fix uncovered bugs/inconsistencies:
- make sure we don't downsample to 420 unless necessary
- fix sign in Op_YCbCr420_to_RRGGBBaa
- for monochrome, only accept heif_colorspace_monochrome+heif_chroma_monochrome (WARNING: behavior change, previously heif_colorspace_YCbCr+heif_chroma_monochrome was also valid)

### DIFF
--- a/libheif/color-conversion/colorconversion.cc
+++ b/libheif/color-conversion/colorconversion.cc
@@ -30,6 +30,7 @@
 #include <set>
 #include <cmath>
 #include <limits>
+#include <string>
 #include "rgb2yuv.h"
 #include "rgb2yuv_sharp.h"
 #include "yuv2rgb.h"
@@ -298,7 +299,7 @@ bool ColorConversionPipeline::construct_pipeline(const ColorState& input_state,
       assert(m_conversion_steps.back().output_state == target_state);
 
 #if DEBUG_ME
-      debug_dump_pipeline();
+      std::cerr << debug_dump_pipeline();
 #endif
 
       return true;
@@ -367,13 +368,15 @@ bool ColorConversionPipeline::construct_pipeline(const ColorState& input_state,
 }
 
 
-void ColorConversionPipeline::debug_dump_pipeline() const
+std::string ColorConversionPipeline::debug_dump_pipeline() const
 {
-  std::cerr << "final pipeline has " << m_conversion_steps.size() << " steps:\n";
+  std::ostringstream ostr;
+  ostr << "final pipeline has " << m_conversion_steps.size() << " steps:\n";
   for (const auto& step : m_conversion_steps) {
     auto& op = *step.operation;
-    std::cerr << "> " << typeid(op).name() << "\n";
+    ostr << "> " << typeid(op).name() << "\n";
   }
+  return ostr.str();
 }
 
 
--- a/libheif/color-conversion/colorconversion.cc
+++ b/libheif/color-conversion/colorconversion.cc
@@ -135,18 +135,6 @@ static void __attribute__ ((unused)) print_spec(std::ostream& ostr, const std::s
   ostr << "\n";
 }
 
-
-static void __attribute__ ((unused)) print_state(std::ostream& ostr, const ColorState& state)
-{
-  ostr << "colorspace=" << state.colorspace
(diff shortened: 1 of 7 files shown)


farindk:
Thanks for the rigorous testing.

Concerning the `monochrome` issue:
I guess we have two options:
a) consider `colorspace_YCbCr` and `colorspace_monochrome` equivalent if we have `chroma_monochrome`, or
b) require that `colorspace_monochrome` and `chroma_monochrome` may only be used in this combination.

I would favor option b) since `colorspace_monochrome` would be useless if it is just a synonym in a special case. Moreover, there are often monochrome images used in special cases like 'alpha channels' or 'masks'. Option a) would require to test both, colorspace and chroma, for this, or we have to test only chroma for monochrome, but colorspace for everything else (RGB vs YCbCr).

it would be clearer if we assume that 'colorspace' should reflect the semantic intention and 'chroma' the memory layout (as a coarse general guideline).

maryla-uc:
Hi Dirk, I agree that it's simpler to have one canonical representation of "monochrome", as colorspace_monochrome+chroma_monochrome.

I added this fix because I noticed that the current conversion.cc test uses both combinations, colorspace_monochrome+chroma_monochrome and colorspace_YCbCr+chroma_monochrome, and that one conversion operator supported both, so I thought this was intentional. But if you're fine with only supporting one, I'm totally fine with that.

However, this is actually a small API behavior change, right? Images created with `heif_image_create(w, h, heif_colorspace_YCbCr, heif_chroma_monochrome, &image)` will now fail to encode with "Unsupported color conversion", even though they worked previously. Is this behavior change ok?

farindk:
In my view, it is ok to clarify this even if it is a behavior change. It will make things easier in the long term.
We can also do this in two steps:
`heif_image_create` might accept both variants and change them internally and then print a warning message that this will be deprecated.

maryla-uc:
So I tried checking some of the projects that use libheif: Krita and ImageMagick do the right thing, while libvips, libgd, gimp and darktable don't use monochrome.
Krita https://github.com/KDE/krita/blob/46b7df109b0e8d21513cdfab7a9ff73626de0d74/plugins/impex/heif/HeifExport.cpp#L329
ImageMagick https://github.com/ImageMagick/ImageMagick/blob/699085f91e996e19c92e2da4fff3952ab19d0a4a/coders/heic.c#L1270

The only problematic usage I found was in OpenImageIO/oiio which uses... heif_colorspace_RGB with heif_chroma_monochrome, which was never valid: https://github.com/OpenImageIO/oiio/blob/e88e4e4b932372120965605e9e7ca18565e39ef1/src/heif.imageio/heifoutput.cpp#L110

Overall I think it's fairly unlikely to break anyone and is not worth adding a warning.

maryla-uc:
By the way do you know why the "fuzzer" presubmit is broken with `/usr/bin/ld: cannot find /home/runner/work/libheif/libheif/clang/lib/clang/17/lib/linux/libclang_rt.fuzzer-x86_64.a: No such file or directory`?

farindk:
> By the way do you know why the "fuzzer" presubmit is broken with `/usr/bin/ld: cannot find /home/runner/work/libheif/libheif/clang/lib/clang/17/lib/linux/libclang_rt.fuzzer-x86_64.a: No such file or directory`?

No, I tried to fix it as the old Ubuntu distribution was not supported anymore, but in the current clang version, there apparently is no `libclang_rt.fuzzer-x86_64.a` anymore. I didn't have the time yet to dig deeper how this is supposed to work now.

farindk:
> I tried checking some of the projects that use libheif

Thanks. It probably really is no big issue then.

I'd still like to add a warning message, so that we can in the future output an error when an invalid combination is used. I'll add this.

kmilos:
> No, I tried to fix it as the old Ubuntu distribution was not supported anymore, but in the current clang version, there apparently is no `libclang_rt.fuzzer-x86_64.a` anymore. I didn't have the time yet to dig deeper how this is supposed to work now.

@farindk Should be in the `libclang-common-XX-dev` package (depending on the Clang version you're installing)

PR #835, opened 2023-04-30
uncompressed: add support for tile alignment padding

bradh:
This PR removes a restriction on tile padding (i.e. UncompressedFrameConfigBox `tile_align_size` being 0) to allow support for padding. 

This is only meaningful for tiled items - we ignore it (treating it as 0) for non-tiled items, which is consistent with the DIS requiring `tile_align_size` to be 0 if a single tile is used.

Three more test files are included (one for each interleave mode, all generated with gpac uncvg filter), and the test code has been refactored to avoid duplicated code.


farindk:
thanks

PR #836, opened 2023-04-30
ci: attempt to fix windows build on appveyor

bradh:
Note: we use `<cassert>` in other places that aren't guarded, so it should be safe outside the windows section here.


farindk:
Yes. That probably was my IDE auto-inserting the #include at the wrong place... It always does that at the most inconvenient place :-)

PR #837, opened 2023-05-02
Update CMakeLists.txt

arjendekorte:
Fix inconsistent use of WITH_GDK_PIXBUF / WITH_GDKPIXBUF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,7 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/libheif.pc
 # ---
 
 option(WITH_EXAMPLES "Build examples" ON)
-option(WITH_GDKPIXBUF "Build gdk-pixbuf plugin" ON)
+option(WITH_GDK_PIXBUF "Build gdk-pixbuf plugin" ON)
 
 option(WITH_REDUCED_VISIBILITY "Reduced symbol visibility in library" ON)
 


farindk:
Thank you. Unfortunately a few minutes too late for v1.16.0...

arjendekorte:
Sorry about that, I just noticed this when attemption to package libheif for openSUSE.

farindk:
On second thought, I think it would be better to change it to "WITH_GDKPIXBUF", because when people start to use that name and we change it again in the next release, it's a bit unfortunate.

farindk:
91b91de - Use WITH_GDKPIXBUF everywhere.

arjendekorte:
Disagree here. Without the patch, adding -D WITH_GDK_PIXBUF=ON will work too, adding -D WITH_GDKPIXBUF=ON is a no-op now.

arjendekorte:
To be more precise, if one adds -D WITH_GDK_PIXBUF=ON to libheif-1.16.0 it will have the intended behaviour of enabling this (and is a workaround for the non-automatic selection of this option), while -D WITH_GDKPIXBUF will do absolutely nothing. So what you've done now, is effectively the opposite of what you mentioned earlier.

farindk:
Ok, I understand what you mean. But then we also have to change `build-emscripten.sh`, which is currently the only place that used the flag.

arjendekorte:
The flag is also used in https://github.com/strukturag/libheif/blob/25899820ecee2f8c2e9eaae06917500f5f81b59f/CMakeLists.txt#L219
This is how I noticed in the first place, since the `gdk-pixbuf` subdirectory was missing and packaging failed.

farindk:
Yes.

Unrelated question: are you going to use the plugin system and separate out the codecs into separate packages? Since there is a growing number of codecs (e.g. we will add JPEG2000 support soon), this may be helpful to reduce the library size and also to avoid licensing issues with HEIC.

arjendekorte:
We do have separate plugins:
```
/usr/lib64/libheif.so.1.16.0
/usr/libexec/libheif/libheif-rav1e.so
/usr/libexec/libheif/libheif-svtenc.so
```
Currently they are in the same package, but we might change this. In that case, they would probably be recommended (so they would be installed automatically in most cases), but it would be possible not to install them. This might indeed be a nice option, as it  would prevent pulling in some quite large libraries.

If there are licensing issues, we can't package libraries at all. Putting them in a separate package won't help in openSUSE, legal will simply not allow them to be built at all.

arjendekorte:
Done. Thanks for the heads up, this makes a lot sense.

PR #844, opened 2023-05-03
Set the heif_color_conversion_options version field

0xC0000054:
The debugger was showing this field as using uninitialized memory,


farindk:
Right. Thank you.

PR #845, opened 2023-05-03
region: add inline and referenced mask support

bradh:
This extends the region item support to support an inline mask and a referenced mask image as the region geometry.

It splits up the referenced mask item from the inline mask because they work quite differently (and have different capabilities - the inline mask is just binary per pixel, while referenced can represent probabilistic masks).

I think there could still be a unifying raster mask API, but not yet.

It also includes partial support for the mask configuration property (mskC), not exposed in API.

I'd particularly like feedback on the inline mask API for returning the packed `data[]` array. That doesn't feel quite right, although it is generally consistent.

Replaces #823.


bradh:
> * Rename `heif_region_get_referenced_mask()` to `heif_region_get_referenced_mask_ID()` for consistency with other functions.

Fixed.

> * `MaskImageCodec::decode_mask_image` should return an error when `bits_per_pixel` has an unsupported value (anything except 8, 16)

Fixed.

> * The fact that the data of inline masks is returned as stored in the file makes it difficult for the user. My proposal would be to return a `heif_image` instead and do the conversion internally. It would also make it much easier to switch between inline and referenced masks as both can be accessed as `heif_image`. There could even be a `heif_region_get_mask()` function that works for both mask types. (In case you want to keep the 1-bit depth low-level data access, this could be a special function.)

This is still to be done. Should the `heif_image` we are returning be scaled / transformed to a parent image (e.g., as for the `_transformed` version on the vector region API)?

> * Do we need plugins/encoder_mask ? It seems superfluous.

Just for convenience / consistency in using the existing encoder API. Without that, it looks like we'd need new API or a special code path in the encoder.

> * You have set the rgan item to 'hidden' (HeifContext::add_region_item). Do you have any reference that says whether it should be hidden or not? I didn't find anything.

Not aware of anything explicit. I think you suggested marking it hidden in an email, and that seems reasonable.

I did check the standard, and the definition is an image "that is not intended to be displayed", and there is a requirement that readers should not display a hidden image item. I think it could go either way - it isn't really a picture for display, and the words about a region and which pixels are in/out are more intended for processing than display. However you could choose to overlay an image mask where its the output of image segmentation (e.g. for visual validation). So I'm not sure and I'll ask around for more views.

bradh:
> > * The fact that the data of inline masks is returned as stored in the file makes it difficult for the user. My proposal would be to return a `heif_image` instead and do the conversion internally. It would also make it much easier to switch between inline and referenced masks as both can be accessed as `heif_image`. There could even be a `heif_region_get_mask()` function that works for both mask types. (In case you want to keep the 1-bit depth low-level data access, this could be a special function.)
> 
> This is still to be done. Should the `heif_image` we are returning be scaled / transformed to a parent image (e.g., as for the `_transformed` version on the vector region API)?

I have now added the ability to get the inline image back as a `heif_image`. I also included support for adding the mask as a `heif_image`. Original versions were kept (seemed like it might be useful, although I'm not hard over on it). I didn't do scaling / transformation / offset.

farindk:
Great, thank you.

farindk:
I have simplified the `heif_region_item_add_region_inline_mask()` function: https://github.com/strukturag/libheif/commit/95d4638cfa240d002989b8a2baf10419009e9435
If I understand it correctly, then the `std::min()`s and the checks against `mask_width`/`mask_height` in the for loops are not necessary. Please check if didn't miss a corner case here.

farindk:
BTW: I also changed the semantics a little bit. When converting an 8-bit mask to a 1-bit inline mask, the input pixel has to by >= `0x80` to get a '1' output. I think this makes more sense when we interpret the 8-bit value as a probability.

farindk:
Another point for discussion: I am not sure that we should keep the API functions `heif_region_get_inline_mask_data()`, `heif_region_get_inline_mask_data_len()`, `heif_region_item_add_region_inline_mask_data()`.
What would be the use case for that? Shouldn't the compression details be transparent to the user?

I'm also wondering whether an `rgan` item might use `content_encoding`.

farindk:
Another strange observation: I tried to add 'deflate' compression of the 'mski' image, since the standard says:
````
If a mask item’s data is lossless-compressed, the compression method shall be signalled using the
content_encoding parameter in the item’s ItemInfoEntry.
````
The code is in the branch `mski-deflate'.

However, I noticed that an `infe` only has a `content_encoding` entry if `item_type == "mime"`.

Can you make sense of that?

bradh:
> Another strange observation: I tried to add 'deflate' compression of the 'mski' image, since the standard says:
> 
> ```
> If a mask item’s data is lossless-compressed, the compression method shall be signalled using the
> content_encoding parameter in the item’s ItemInfoEntry.
> ```
> 
> The code is in the branch `mski-deflate'.
> 
> However, I noticed that an `infe` only has a `content_encoding` entry if `item_type == "mime"`.
> 
> Can you make sense of that?

I think that only applies to `version` 2 or 3 of `infe`.

farindk:
> >    However, I noticed that an infe only has a content_encoding entry if item_type == "mime".
>
> I think that only applies to version 2 or 3 of infe.

But `item_type` is only present in version 2 and 3. So how should an `infe` box look like for compressed `mski` ?

bradh:
> Another point for discussion: I am not sure that we should keep the API functions `heif_region_get_inline_mask_data()`, `heif_region_get_inline_mask_data_len()`, `heif_region_item_add_region_inline_mask_data()`. What would be the use case for that?

(I've split this part from the next question).

The concept was that if you have some representation out of an image segmentation algorithm, it probably isn't in something that is similar to a libheif pixel image. So its more likely you'd have to convert it (probably line by line). So possibly its more efficient just to convert the segmentation output directly to an inline mask, rather than have to convert to a pixel image, only to have libheif unconvert it.

> Shouldn't the compression details be transparent to the user?

Yes. So the "data" in the case of a compressed image would be before compression (on the `_add_region_inline_mask_data()` case) or after decompression (on the `_get_inline_mask_data[(),_len()]` cases). Possibly there should be some kind of parameter to allow the user to select whether compression occurs and the format, although that would ideally be consistent with whatever other options we're providing for compression (e.g. in metadata).


> I'm also wondering whether an `rgan` item might use `content_encoding`.

Possibly, but the only case where I can see it helping is in the inline case, and there is a `mask_coding_method` and `mask_coding_parameters` so I think it'd be more sensible to just do that for deflate.

bradh:
> > > However, I noticed that an infe only has a content_encoding entry if item_type == "mime".
> > 
> > 
> > I think that only applies to version 2 or 3 of infe.
> 
> But `item_type` is only present in version 2 and 3. So how should an `infe` box look like for compressed `mski` ?

Indeed. I think that looks like a bug in the spec. Let me reach out.

bradh:
> I have simplified the `heif_region_item_add_region_inline_mask()` function: [95d4638](https://github.com/strukturag/libheif/commit/95d4638cfa240d002989b8a2baf10419009e9435) If I understand it correctly, then the `std::min()`s and the checks against `mask_width`/`mask_height` in the for loops are not necessary. Please check if didn't miss a corner case here.

In the documentation, I said the image would be clipped or expanded (see https://github.com/strukturag/libheif/blob/master/libheif/heif.h#LL2751C17-L2755C74) to match the specified width and height. I don't think that is happening any more, and so the width and height may be bogus.

If you'd like to match the input image, I can live with that. However I think we need to update the documentation, and probably remove the width and height parameters from the function (just using the image size instead).

bradh:
> BTW: I also changed the semantics a little bit. When converting an 8-bit mask to a 1-bit inline mask, the input pixel has to by >= `0x80` to get a '1' output. I think this makes more sense when we interpret the 8-bit value as a probability.

I think that is fine. However it does not match the documentation at https://github.com/strukturag/libheif/blob/master/libheif/heif.h#L2529-L2532 . Would you like me to update that?

farindk:
I don't see a conflict with the documentation. The 8-bit semantics didn't change. I just moved the threshold for the conversion from 8-bit to 1-bit to `0x80` = probability 0.5

bradh:
> I don't see a conflict with the documentation. The 8-bit semantics didn't change. I just moved the threshold for the conversion from 8-bit to 1-bit to `0x80` = probability 0.5

The documentation currently says any non-zero value.

farindk:
Ah, we were looking at different text. The link above says lines 2529-2532, but I think you actually mean 2831-2833.
Right, let me make that small change...

farindk:
I think we should also delete the sentences after that:
````
 If the image width is less that the specified width, it is expanded                                                                                                                                                                                         
 * to match the width of the region (zero fill on the right). If the image height is                                                                                                                                                                                        
 * less than the specified height, it is expanded to match the height of the region                                                                                                                                                                                         
 * (zero fill on the bottom). If the image width or height is greater than the                                                                                                                                                                                              
 * width or height (correspondingly) of the region, the image is cropped.    
 ````
That seems to refer to the `min(mask_width, image_width)` thing that I deleted. The mask can actually be scaled to any size in the reference coordinate space and thus I assume its size should be independent from the mask image size.

bradh:
> I think we should also delete the sentences after that:
> 
> ```
>  If the image width is less that the specified width, it is expanded                                                                                                                                                                                         
>  * to match the width of the region (zero fill on the right). If the image height is                                                                                                                                                                                        
>  * less than the specified height, it is expanded to match the height of the region                                                                                                                                                                                         
>  * (zero fill on the bottom). If the image width or height is greater than the                                                                                                                                                                                              
>  * width or height (correspondingly) of the region, the image is cropped.    
> ```
> 
> and simply remove the `width`, `height` parameters from the function. These should be taken from the image. Or am I missing something?

Concur (see https://github.com/strukturag/libheif/pull/845#issuecomment-1574474353)

PR #848, opened 2023-05-03
Allow callers to get the number of properties

0xC0000054:
The code to count the number of properties when out_list == nullptr was being skipped due to a null pointer check on that parameter.

Updated the header documentation for heif_item_get_properties_of_type and heif_item_get_transformation_properties to reflect this change.


farindk:
Right. Thank you.

PR #849, opened 2023-05-04
Handle _write() conversion warning on Windows

kmilos:
There is no POSIX matching write(), even w/ unistd.h present on e.g. MinGW, just an alias for Miscrosoft's CRT _write.

See also https://github.com/msys2/MINGW-packages/issues/17063


farindk:
Thanks for going down the rabbit hole of the wonderful world of integer conversions :-)

PR #850, opened 2023-05-04
README: cmake parameter order

oittaa:
At least on my Debian Docker image the build failed with the previous parameter order. The official man page also says the options should be given first before the path.

`cmake [<options>] <path-to-source | path-to-existing-build>`

https://cmake.org/cmake/help/latest/manual/cmake.1.html


farindk:
Thank you.

PR #853, opened 2023-05-05
update .gitignore to reflect CMake conversion

bradh:
As discussed in #726, we no longer need most of the ignore entries.

I have added one for the `build` directory to match the recommendation in README.md.


farindk:
Radical clean up :-)
But I'll egoistically add a few more to cover my setup.

PR #856, opened 2023-05-05
examples: add -v version command line option

### LINKED ISSUE #571: Please add -v to heif-convert and heif-thumbnailer
Thanks for the tool.
heif-enc and heif-info give information abount themselve. Please add this feature to heif-convert und heif-thumbnailer.

bradh:
Resolves #571

### DIFF
--- a/examples/heif_convert.cc
+++ b/examples/heif_convert.cc
@@ -74,10 +74,11 @@ static void show_help(const char* argv0)
                "Usage: heif-convert [options]  <input-image> <output-image>\n"
                "\n"
                "The program determines the output file format from the output filename suffix.\n"
-               "These suffices are recognized: jpg, jpeg, png, y4m."
+               "These suffixes are recognized: jpg, jpeg, png, y4m."
                "\n"
                "Options:\n"
                "  -h, --help                     show help\n"
+               "  -v, --version                  show version\n"
                "  -q, --quality                  quality (for JPEG output)\n"
                "  -d, --decoder ID               use a specific decoder (see --list-decoders)\n"
                "      --with-aux                 also write auxiliary images (e.g. depth images)\n"
@@ -87,6 +88,7 @@ static void show_help(const char* argv0)
                "      --no-colons                replace ':' characters in auxiliary image filenames with '_'\n"
                "      --list-decoders            list all available decoders (built-in and plugins)\n"
                "      --quiet                    do not output status messages to console\n"
+               "      --show-config              show configuration information\n"
                "  -C, --chroma-upsampling ALGO   Force chroma upsampling algorithm (nn = nearest-neighbor / bilinear)\n";
 }
 
@@ -115,6 +117,8 @@ int option_with_exif = 0;
 int option_skip_exif_offset = 0;
 int option_list_decoders = 0;
 
+const int OPTION_SHOW_CONFIGURATION = 1000;
+
 std::string chroma_upsampling;
 
 static struct option long_options[] = {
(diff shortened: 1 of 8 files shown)


examples/heif_convert.cc
```diff
@@ -220,6 +222,9 @@ int main(int argc, char** argv)
           chroma_upsampling = "nearest-neighbor";
         }
         break;
+      case 'v':
+        std::cout << argv[0] << "   libheif version: " << heif_get_version() << std::endl;
```
farindk, line 226: How about only writing the plain version number? That would make it easier for scripts to parse it. And the text in front does not give much information anyway.

farindk: Or: we could output both, the libheif version and the tool version (which may theoretically be different).

bradh: You mean like:

```
$ ./heif-convert -v
1.16.1
```

?

farindk: Yes. Or:
````
$ ./heif-convert -v
libheif: 1.16.1
heif-convert: 1.15.0
````

Not sure if that is useful. Maybe for "broken" installations.

Comment (bradh) [reply examples/heif_convert.cc:226]: As a point of comparison:

```
$ ./avifdec --version
Version: 0.9.3 (dav1d [dec]:0.9.2, libgav1 [dec]:0.17.0, aom [enc/dec]:v3.3.0)
libyuv : available (1807)
```

Comment (bradh) [reply examples/heif_convert.cc]: I think I like this:

```
$ ./heif-thumbnailer -v
./heif-thumbnailer: 1.16.1 (libheif: 1.16.1)
$ ./heif-convert -v
./heif-convert: 1.16.1 (libheif: 1.16.1)
```

Comment (farindk) [reply examples/heif_convert.cc]: Yes. I think something like this would be most useful. The codec versions are already printed in `--list-encoders` / `--list-decoders`, but other things, like `libsharp-yuv` version or some build configuration could also be useful.

Comment (farindk) [reply examples/heif_convert.cc]: One item per line has the advantage that it is easy to grep:
````
heif-convert: 1.15.0
libheif: 1.16.1
libsharp-yuv: N/A
libjpeg: 6.2
...
````

Comment (bradh) [reply examples/heif_convert.cc]: OK. This was meant to be an easy fix for a tired Friday night :-)

Will give it further attention over the next few days.

**Comment by bradh** [comment]
Probably also needs to go into the other tools, noting that `heif_enc` already has an option using the short `-v` form for something else.

File: examples/heif_convert.cc (hunk 1 above)
Comment (bradh) [reply examples/heif_convert.cc]: Just one additional thought: should we do a plain `-v` / `--version` option for the tool that would be the result of `heif_get_version()` , and a `--dump-configuration` or similar for the build time and run-time versions where we can figure it out?

Comment (farindk) [reply examples/heif_convert.cc]: I also was wondering about whether we should have two options. I concluded that I have no clear opinion.
BTW: for consistency, `-v` should then output the tool version and not `heif_get_version()`.

**Comment by farindk** [comment]
> Probably also needs to go into the other tools, noting that `heif_enc` already has an option using the short `-v` form for something else.

I have no problem with removing `-v/--verbose` from `heif-enc`. At least I did not use this for years...

**Comment by bradh** [comment]
I split the functionality. The thumbnailer only has version. The others have a `--show-config` option, that looks like this:

```
$ ./heif-enc --show-config
libheif:1.16.1
plugin paths:
libsharpyuv:no
```

Happy to add more to that, or to go with this for now and add as needed.

bradh:
Hmm, looks like it doesn't like `get_plugin_paths()`. Strange that it worked locally though...

farindk:
`get_plugin_paths()` is not an exported library symbol. It worked when you did a build with `WITH_REDUCED_VISIBILITY=off`. E.g. when the tests are enabled.

bradh:
> `get_plugin_paths()` is not an exported library symbol. It worked when you did a build with `WITH_REDUCED_VISIBILITY=off`. E.g. when the tests are enabled.

Thanks. 

I plan to split this into a simple `-v` part, and rethink the configuration dumping. Will raise a new PR when I get back to it.

farindk:
I think a simple `-v` instead of `--show-config` will do. Otherwise, we have 3 places where version information is printed:
`--help`, `-v`, `--show-config`.
The typical user will be happy with the simple version from `--help` and use `-v` to get the full information.

Regarding `get_plugin_paths()`, we might add an API function for that, but it'll have to wait until v1.17.0.

PR #857, opened 2023-05-07
uncompressed: initial work on encoder

bradh:
Supports a limited range of inputs. 

Mono and simple RGB pixel interleave work OK.

YCbCr doesn't work correctly - encoded files look to have chroma problems when decoded with gpac and also with my viewer.

Additional work is needed for:

 - [X] RGBA pixel interleave
 - [X] RGB planar
 - [X] RGBA planar
 - [X] Mono + alpha from PNG
 - [ ] Strange image sizes (need to find / create test data)
 - [X] Higher bit depth (need to find / create test data)
 - [X] YCbCr encoding
 - [x] Fix clang warnings

and probably other things...

@farindk would appreciate feedback on the approach if you have time.

### DIFF
--- a/libheif/uncompressed_image.cc
+++ b/libheif/uncompressed_image.cc
@@ -25,37 +25,54 @@
 #include <map>
 #include <iostream>
 
+#include "libheif/heif.h"
 #include "uncompressed_image.h"
 
 
-enum heif_component_type
+enum heif_uncompressed_component_type
 {
-  heif_component_type_monochrome = 0,
-  heif_component_type_Y = 1,
-  heif_component_type_Cb = 2,
-  heif_component_type_Cr = 3,
-  heif_component_type_red = 4,
-  heif_component_type_green = 5,
-  heif_component_type_blue = 6,
-  heif_component_type_alpha = 7,
-  heif_component_type_depth = 8,
-  heif_component_type_disparity = 9,
-  heif_component_type_palette = 10,
-  heif_component_type_filter_array = 11,
-  heif_component_type_padded = 12
+  component_type_monochrome = 0,
+  component_type_Y = 1,
+  component_type_Cb = 2,
+  component_type_Cr = 3,
+  component_type_red = 4,
+  component_type_green = 5,
+  component_type_blue = 6,
+  component_type_alpha = 7,
+  component_type_depth = 8,
+  component_type_disparity = 9,
+  component_type_palette = 10,
+  component_type_filter_array = 11,
+  component_type_padded = 12
+};
+
+
+enum heif_uncompressed_component_format
+{
+  component_format_unsigned = 0,
+  component_format_float = 1,
+  component_format_complex = 2,
+};
+
+
+enum heif_uncompressed_sampling_type
+{
+  sampling_type_no_subsampling = 0,
+  sampling_type_422 = 1,
+  sampling_type_420 = 2,
+  sampling_type_411 = 3
 };
 
 
 enum heif_uncompressed_interleave_type
 {
-  heif_uncompressed_interleave_type_component = 0,
-  heif_uncompressed_interleave_type_pixel = 1,
-  heif_uncompressed_interleave_type_mixed = 2,
-  heif_uncompressed_interleave_type_row = 3,
-  heif_uncompressed_interleave_type_tile_component = 4,
-  heif_uncompressed_interleave_type_multi_y = 5
+  interleave_type_component = 0,
+  interleave_type_pixel = 1,
+  interleave_type_mixed = 2,
+  interleave_type_row = 3,
+  interleave_type_tile_component = 4,
+  interleave_type_multi_y = 5
 };
-
 
 Error Box_cmpd::parse(BitstreamRange& range)
 {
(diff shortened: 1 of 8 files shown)


farindk:
I just tried it with PNG input. Encoding produces a HEIF file, but when I try to decode it, I get
````
Could not decode image: 0: Unsupported feature: Unsupported data version: Uncompressed image with component_align_size 1 is not implemented yet
````

Nothing special on the command line:
````
$ ./heif-enc -U input.png -o out.heif
$ ./heif-convert out.heif dec.png
````
The same with any PNG input I tried. Let me know if you need an example HEIF.

**Comment by bradh** [comment]
> I just tried it with PNG input. Encoding produces a HEIF file, but when I try to decode it, I get
> 
> ```
> Could not decode image: 0: Unsupported feature: Unsupported data version: Uncompressed image with component_align_size 1 is not implemented yet
> ```
> 
> Nothing special on the command line:
> 
> ```
> $ ./heif-enc -U input.png -o out.heif
> $ ./heif-convert out.heif dec.png
> ```
> 
> The same with any PNG input I tried. Let me know if you need an example HEIF.

I have this fixed locally - its valid to have `component_align_size` of 1, but was not handled by the decoder. It should work with a really recent gpac if you want to try that for decoding instead of `heif-convert`.

File: libheif/uncompressed_image.cc [hunk 1]
```diff
@@ -25,38 +25,55 @@
 #include <map>
 #include <iostream>
 
+#include "libheif/heif.h"
 #include "uncompressed_image.h"
 
 
-enum heif_component_type
+enum uncompressed_component_type
```
farindk, line 32: I think we should keep the old name with the `heif_` prefix because I expect that this enum will eventually be moved to the public API where we need the prefix. Of course, we could postpone the prefix until then, but I guess it will be sooner than later anyway.

libheif/uncompressed_image.cc
```diff
@@ -25,38 +25,55 @@
...
+  component_type_depth = 8,
+  component_type_disparity = 9,
+  component_type_palette = 10,
+  component_type_filter_array = 11,
+  component_type_padded = 12
 };
 
 
-enum heif_uncompressed_interleave_type
+enum uncompressed_component_format
```
farindk, line 50: Some thing with the `heif_` prefix here and thje two below.

libheif/uncompressed_image.cc
bradh: I had hoped to hide the detail of the uncompressed codec behind the existing API. No problem with changing the names back though.

libheif/uncompressed_image.cc
```diff
@@ -25,38 +25,55 @@
...
-  heif_uncompressed_interleave_type_row = 3,
-  heif_uncompressed_interleave_type_tile_component = 4,
-  heif_uncompressed_interleave_type_multi_y = 5
+  component_format_unsigned = 0,
+  component_format_float = 1,
+  component_format_complex = 2,
 };
 
 
+enum uncompressed_sampling_type
```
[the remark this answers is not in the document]
bradh: I think `heif_chroma` does more than just sampling type (e.g. interleave type and bit depth). That is useful in that you only get combinations that work, but does represent a bit of a mismatch when parsing and when using these values as constants for write-out.

libheif/uncompressed_image.cc
farindk: Yes, I understand that, but it appears that there will be many more components (UV, X-Ray, even though they apparently have been removed from the latest draft) so that we will have to expose the components enum to the user in some way to let them attach such images. We can postpone the details until later.

libheif/uncompressed_image.cc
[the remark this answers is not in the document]
farindk: Yes, I agree. Let's keep the uncompressed chroma type separate. Just like the 'heif_components' are separate from 'heif_channel'.
Maybe we could even make this more clear by naming the uncompressed-specific types with a `heif_unc_` prefix. But this can also be done later when this is exposed in a special API.

bradh:
I think there could still be cases where unusual image sizes might be a problem, but I'd like to get this in soon to avoid ongoing merge conflicts. It remains behind a cmake flag.

farindk:
I'll give it a try

farindk:
It seems that all odd-sized images don't pass correctly through encoding / decoding. Take this as an example:

Input PNG:
rainbow-451x461

compressed/decompressed:
out

bradh:
Perhaps a problem with stride != width. Are you using `heif-enc` + `heif-convert` for testing?

farindk:
Yes, I've fixed encoding ing `uncompressed_image.cc` lines 958- as follows:
````
   {
      int src_stride;
      uint8_t* src_data = src_image->get_plane(heif_channel_interleaved, &src_stride);
      uint64_t out_size = src_image->get_height() * src_image->get_width();
      data.resize(out_size);
      for (int y = 0; y < src_image->get_height(); y++) {
        memcpy(data.data() + y * src_image->get_width(), src_data + src_stride * y, src_image->get_width());
      }
      heif_file->append_iloc_data(out_image->get_id(), data, 0);
    }
````

But now it crashes in the decoder.

bradh:
Thanks. Can you push that onto the branch?

Likely similar changes will be required in YCbCr.

I can look at the decoder a bit later.

farindk:
My patch is wrong. It has to include all interleaved components...

bradh:
You can get the number of components and the size of each from the `uncC` box content.

That saide, it might be easier to just copy the fill stride and flag the use of row fill in the `uncC` box.

farindk:
Probably not the most elegant solution, but it works

bradh:
> Probably not the most elegant solution, but it works

Thanks.

farindk:
The same for YCbCr planes. But it cannot be loaded yet, because the decoder does not support 4:2:0.

I have the feeling that we'll need some more general way to handle the different options since the number of cases is going to explode. But I'll merge the current state.

PR #859, opened 2023-05-08
tests: enable testing where practical

bradh:
[This is a somewhat speculative change - could see the case for backing some of it out]

Prior to this change, tests were only available if reduced symbol visibility was disabled.

That wasn't needed for the uncompressed codec tests. With a small change, it also isn't really needed for the encode tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,7 +212,7 @@ endif (DOXYGEN_FOUND)
 
 # --- Testing
 
-option(BUILD_TESTING "" OFF)
+option(BUILD_TESTING "" ON)
 include(CTest)
 if(BUILD_TESTING)
     # TODO: fix tests on windows.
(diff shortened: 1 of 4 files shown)


farindk:
I'm not sure this is worth the effort. Tests are simply allowed to access the whole internals. The fact that some don't need to is mainly due to the fact that we do not have many tests. Diving this into tests that need this and tests that don't will become a maintenance nightmare and I don't see a real need for this.

bradh:
The desire to do so is partly driven by not wanting to test internals, but rather to test interfaces (and make sure I'm exporting things correctly). In some ways it is a unit test vs integration test distinction.

I had expected that the tests that didn't need internal symbols would be the majority, and would always run in CI, so it wouldn't be a bit deal.

bradh:
@farindk I'd like to request reconsideration of this. It isn't trying to say that tests should not access internals, but rather that _if_ they don't need internals, they are able to be run without disabling visibility constraints. 

Its not that big a maintenance issue IMO, since you can always put it inside the "must have reduced visibility" part.

farindk:
Ok, let's see how this will evolve.

PR #860, opened 2023-05-10
example: fix heif-enc encoder option

bradh:
Using the heif-enc `--encoder ID` option (prior to this PR) produces:

```
$ ./heif-enc --encoder x265 something.png
Can't open x265
```

That is a bit confusing - its actually trying to open "x265" as an input file.

The option was missing an argument flag, which is added in this PR.


farindk:
Thanks for the fix.

PR #861, opened 2023-05-11
Allow callers to get the encoder descriptor count

0xC0000054:
Changed `heif_get_encoder_descriptors` to return the number of matching encoder descriptors when `out_encoders==nullptr`.
Updated the header documentation to reflect this change.


farindk:
Thank you.

farindk:
I have simplified the code a bit.

PR #866, opened 2023-05-16
Explicitly use std::pow.

0-wiz-0:
This fixes a build failure on SmartOS, see
https://us-central.manta.mnx.io/pkgsrc/public/reports/upstream-trunk/20230514.2249/libheif-1.16.1/build.log
```
/home/pbulk/build/graphics/libheif/work/libheif-1.16.1/libheif/hevc.cc: In function 'double read_depth_rep_info_element(BitReader&)':
/home/pbulk/build/graphics/libheif/work/libheif-1.16.1/libheif/hevc.cc:46:16: error: call of overloaded 'pow(int, int)' is ambiguous
   46 |     value = pow(2, exponent - 31) * (1.0 + mantissa / pow(2, mantissa_len));
      |             ~~~^~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libheif/hevc.cc
+++ b/libheif/hevc.cc
@@ -43,10 +43,10 @@ static double read_depth_rep_info_element(BitReader& reader)
   //printf("sign:%d exponent:%d mantissa_len:%d mantissa:%d\n",sign_flag,exponent,mantissa_len,mantissa);
 
   if (exponent > 0) {
-    value = pow(2, exponent - 31) * (1.0 + mantissa / pow(2, mantissa_len));
+    value = pow(2.0, exponent - 31) * (1.0 + mantissa / pow(2.0, mantissa_len));
   }
   else {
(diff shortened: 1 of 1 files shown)


farindk:
Your compiler output looks like the actual problem is that it cannot decide between `pow(float, int)` and `pow(double, int)`.
Hence, a better fix could be to replace the `pow(2, ...)` with `pow(2.0, ...)`.

Can you please try if that works?

0-wiz-0:
Yes, that works too. I've adapted the diff. Thanks for the suggestion!

farindk:
There is another `pow()` in the first of the two lines. Why doesn't the compiler complain about that one?
I guess this should also get a `2.0`.

0-wiz-0:
Yeah, sorry for that.
I'm reporting this for another party that provided the fix and okayed your alternative solution, and I created the new patch and missed the third `pow`; this went unnoticed because I don't have access to the system.
New version pushed.

PR #869, opened 2023-05-19
tests: initial upgrade to Catch2 version 3.3.2

bradh:
This only does the "amalgamated" migration approach, rather than a proper static library. See https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md#how-to-migrate-projects-from-v2-to-v3


bradh:
The other approach to this would be a git submodule. However that does require a bit more work on the part of developers (for the submodule init or recurse etc).

farindk:
Note: this will increase the minimum C++ version to C++14. Currently, it is C++11.
We have to check whether that is fine even for old distributions.

bradh:
> Note: this will increase the minimum C++ version to C++14. Currently, it is C++11. We have to check whether that is fine even for old distributions.

I had hoped it would only be needed to run tests. The C++14 support is default in gcc 6 and some of it is available back to later gcc 4 versions, per https://gcc.gnu.org/projects/cxx-status.html#cxx14

It might be marginal on RHEL 7, but RHEL 8 and derivatives would be fine.

farindk:
Since we are not using C++20, we can finally merge this. I have merged this and done the required changes in #1364 .

PR #870, opened 2023-05-22
file: fix argument order

bradh:
The order in the file.h is:
void add_iref_reference(heif_item_id from, uint32_t type, const std::vector<heif_item_id>& to);

And in box.[cc,h]:
void add_reference(heif_item_id from_id, uint32_t type, const std::vector<heif_item_id>& to_ids)


farindk:
Good catch :-)

PR #872, opened 2023-05-23
jpeg: brands

bradh:
A little contribution to recognise JPEG in HEIF.

Feel free to close if this dupes existing work...


farindk:
Right, thanks.

PR #874, opened 2023-05-26
jpeg2000: initial box implementation

bradh:
This adds boxes defined in ISO/IEC 15444-16:2021 for encapsulation of JPEG 2000 in ISO/IEC 23008-12 (i.e. using J2K in HEIF).

Because there is no known test data, includes unit testing.

This does not provide codec support, or image sequences, although required brands are added.


farindk:
Thank you.

PR #877, opened 2023-05-29
examples: add -v version command line options

### LINKED ISSUE #571: Please add -v to heif-convert and heif-thumbnailer

bradh:
This is a simplified (reduced scope) version of #856, just adding the version, and not attempting to do the configuration status.

Resolves #571


farindk:
Thanks. I have extended this a bit to also show the libheif version and all plugin directories. That should be useful for users that have issues with their installation.

PR #880, opened 2023-06-01
ftyp: do not add compatible brand if already present

bradh:
Minor TODO fix.

Couldn't find a good formatting for the `has_compatible_brand()` part of the change. Please feel free to adjust.


farindk:
thanks

PR #881, opened 2023-06-03
kvazaar: check for logging support

bradh:
The kvazaar encoder will not compile if kvazaar library is too old (prior to https://github.com/ultravideo/kvazaar/commit/580c6e27586d82394b1912ea1ef9932f8572a59d).

This adds a check for the problematic structure member, and only sets that field if available. The other option would be to only compile the kvazaar encoder if logging is supported, however since the output is mostly cosmetic noise, that didn't seem warranted.


farindk:
thanks

PR #885, opened 2023-06-03
Implement 1pic brand

### LINKED ISSUE #884: Support 1pic brand on write
ISO/IEC 23008-12:2022 Section 10.2.5 defines the `1pic` brand.

The short version is a single intra-coded image.

Subtasks:

 - [x] Add brand
 - [x] Add check for brand compatibility on write

bradh:
Also cleans up some existing code that was being added to, for consistency.

Resolves #884

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -17,6 +17,7 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with libheif.  If not, see <http://www.gnu.org/licenses/>.
  */
+#include "libheif/heif.h"
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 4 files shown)


farindk:
How do you interpret the `1pic` semantics? Is a grid image that is composed of several images also a `1pic`? My interpretation would be yes, because otherwise most images from phones would not be compatible even though they appear to the user as clearly one image one.

bradh:
I think from the phrasing in the spec that there can be multiple image items, but only one that is visible. 

More generally, I interpret this as the 'replace PNG / JPEG' image capability.

PR #891, opened 2023-06-05
Fix a mem leak when `heif_deinit()` was not called

kleisauke:
By switching to `std::shared_ptr` we ensure that these smart pointers' lifetimes persist throughout the entire execution time of the program, eliminating the requirement of calling `heif_deinit()`.

This is especially useful in fuzzer environments where calling `heif_deinit()` is not always feasible.

Regressed since commit 8f9ac3d6db58d91ebb36f68244cdc534253b889b.

---
Context (in chronological order, none of them would appease OSS-Fuzz):
https://github.com/libvips/libvips/pull/3481
https://github.com/libvips/libvips/pull/3489
https://github.com/libvips/libvips/pull/3498

### DIFF
--- a/fuzzing/file_fuzzer.cc
+++ b/fuzzing/file_fuzzer.cc
@@ -85,8 +85,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
   struct heif_image_handle* primary_handle = nullptr;
   int images_count;
   heif_item_id* image_IDs = NULL;
+  bool explicit_init = data[size - 1] & 1;
 
-  heif_init(nullptr);
+  if (explicit_init) {
+    heif_init(nullptr);
+  }
 
   heif_check_filetype(data, clip_int(size));
   heif_main_brand(data, clip_int(size));
@@ -151,7 +154,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
   heif_context_free(ctx);
   free(image_IDs);
 
-  heif_deinit();
+  if (explicit_init) {
+    heif_deinit();
+  }
 
   return 0;
 }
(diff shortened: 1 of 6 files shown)


farindk:
Thanks.
I understand that the color conversions leaked when `heif_init`/`heif_deinit` was not used, but I don't understand why it's a problem to use `heif_init`/`heif_deinit` in the fuzzers. Is the second commit where you reverted the use of init/deinit in the fuzzers really necessary?

kleisauke:
It should not be necessary, but it helps to catch any potential further regressions. 

If you prefer, I can omit that commit. libvips' fuzzers should be able to catch such regressions anyway.

farindk:
Thanks, now I understand your intention. But that then has the effect that potential errors in init/deinit won't be detected.
How about modifying `file_fuzzer.cc` such that init/deinit is only called `if (data[size-1] & 1)` ? Then both cases would be tested.

kleisauke:
Good point, I've just updated this PR accordingly.

fuzzing/file_fuzzer.cc
```diff
@@ -85,8 +85,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
   struct heif_image_handle* primary_handle = nullptr;
   int images_count;
   heif_item_id* image_IDs = NULL;
+  bool explicit_init = data[size - 1] & 1;
```
farindk, line 88: Seems we need an `size ? (data[size - 1] & 1) : false`

kleisauke: Indeed, I committed 4ae47b2c3f88f1b86a561cfa4bc5b60018b21063 before noticing this review comment. Let me know if you want the else-clause (`size == 0`) to be `false` instead.

farindk: I guess it doesn't matter.

farindk:
Thank you.

Issue #767, opened 2023-01-16
Static build of heif-convert


jankais3r:
Hi,

I am trying to create a static build of `heif-convert`. Apologies if this is a rookie mistake, I am not well versed in the c++ toolchain.

I first tried to use the CMake approach as autogen is deprecated. I followed instructions from #189, but the result was a dynamically linked binary.
```
$ mkdir build
$ cd build
$ cmake -DBUILD_SHARED_LIBS=OFF ..
$ make
$ file examples/heif-convert
heif-convert: ELF 64-bit LSB pie executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=b45e6dcd01374f8a7e03b46221f6559ad36e9e8b, for GNU/Linux 3.2.0, not stripped
```

I also tried autogen, but the result was the same:
```sh
$ ./autogen.sh
$ ./configure --enable-static
$ make
$ file examples/.libs/heif-convert
heif-convert: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=1eb37876e66816492eda3ce886b8fe2c5150533e, for GNU/Linux 3.2.0, with debug_info, not stripped
```

Could anyone point me to the right direction please?

bradh:
There are four parts to building a libheif example application such as `heif-convert` statically:

1. Having all the dependencies available as static libraries
2. Ensuring that the static version of those dependencies is selected
3. Building libheif as a static library
4. Statically linking the executable

The first part is out-of-scope for this description, but on Linux, you'll typically get that as part of whatever development package (-dev, -devel) is providing the dynamic library. Its possible you might only have a dynamic library of some encoder or decoder library, but if you don't need that, you can turn it off in the CMake build. (Note: On Ubuntu, libde265.a and libdav1d.a are not available in the `-dev` package. Also, the package for x265 provides `libx265.a`, but it requires methods in libnuma, which is not available as a static library. If you need those, you'll have to build it yourself.).

[...]

PR #893, opened 2023-06-11
plugins: only build plugin code when plugins loading is enabled

bradh:
While working on static build at #767, I noted that we get warnings that look like this:

```
[ 60%] Linking CXX executable heif-info
/usr/bin/ld: ../libheif/libheif.a(plugins_unix.cc.o): in function `PluginLibrary_Unix::load_from_file(char const*)':
plugins_unix.cc:(.text+0x4ad): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
[ 60%] Built target heif-info
```

Clearly it isn't very likely you'd be doing static linking on the application, then using plugins - not absolutely out of the question, but unusual.

We do have an option to disable plugin loading (the CMake `ENABLE_PLUGIN_LOADING` directive), however that currently isn't enough to avoid this (compile-time) warning. This PR extends the current use of that option to wrap the plugin loading code (for Linux and Windows) in an `#if ENABLE_PLUGIN_LOADING` stanza.

### DIFF
--- a/libheif/plugins_unix.cc
+++ b/libheif/plugins_unix.cc
@@ -18,6 +18,7 @@
  * along with libheif.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#if ENABLE_PLUGIN_LOADING
 
 #include "plugins_unix.h"
 #include "heif_plugin.h"
@@ -116,4 +117,6 @@ void PluginLibrary_Unix::release()
     m_library_handle = nullptr;
   }
 }
+
+#endif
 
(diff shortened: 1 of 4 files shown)


farindk:
Are you sure? When I disable `ENABLE_PLUGIN_LOADING`, these warnings disappear for me.
And they really should because `plugins_unix.cc` is not even going to be compiled when plugins are turned off.

bradh:
Indeed you are correct. I'm not sure where my earlier testing went wrong.

PR #900, opened 2023-06-16
Add more nclx conversion tests and fix related bugs.

maryla-uc:
Some operators were not checking for matrix coefficients 0/8/11/14 even though they were not supported. It looks like the operation succeeds but get_RGB_to_YCbCr_coefficients and get_YCbCr_to_RGB_coefficients just return default values in this case (maybe this needs to be changed?)
There is only one operator that supports matrix_coefficients 8, and none of them support 11 or 14 currently.


farindk:
Look good. Thank you.

Issue #895, opened 2023-06-13
 the quality \ the filesize of different libaom versions are quite different.


Hpsyche:
# Background
Since libaom has optimized the time consumption and image size in version 3.0.0+, we consider upgrading libaom to version 3.6.0;

# Question
Using the same image, under the same heif-enc\lieheif version, it is found that after upgrading libaom the quality and image size are quite different.
## Environment
ubuntu 20.04
libheif version: 1.14.2
## Original image
![ori_gh](https://github.com/strukturag/libheif/assets/37608082/13cff7f9-3e02-46a0-909f-fbec2dfa55e3)
image_size = 17688
## Command
`heif-enc ori.jpg -p quality=50 -A -o dest.avif`
## Result
[dest_avif.zip](https://github.com/strukturag/libheif/files/11733656/dest_avif.zip)
#### aom 2.0.0 version
```
heif-enc -A --list-encoders
aom = AOMedia Project AV1 Encoder v2.0.0
```
image_size = 14514

#### aom 3.6.0 version
```
heif-enc -A --list-encoders
aom = AOMedia Project AV1 Encoder v3.6.0
```
image_size = 7251
The image size and quality of the new version have deteriorated a lot. 
## Confuse
I try to use `heif-enc -A -P`,found that the default parameters are the same. 
So I'm confused that is this normal, or my parameters are wrong ?

wantehchang:
@Hpsyche Thank you for the bug report. I can see the quality difference between aom_2_0_0.avif and aom_3_6_0.avif. I will try to reproduce the results and take a look.

wantehchang:
@Hpsyche @farindk

I found out the reason. This is caused by a bug in libaom v2.0.0 (fixed by the change https://aomedia-review.googlesource.com/c/aom/+/113064). When the bug is present, libaom will use a lower quantizer index (which implies higher quality and larger file size) for the first frame.

I added some debugging printf statements to libaom and libheif. Here is the output for the command `heif-enc ori.jpg -p quality=50 -A -o dest.avif`.

Before the bug fix (at libaom commit 0481a781e262b90f659664135a9ba62bfb59a011):

```
Setting cfg.rc_min_quantizer = 0, cfg.rc_max_quantizer = 63
Calling aom_codec_control(AOME_SET_CPUUSED, 6)
Calling aom_codec_control(AOME_SET_CQ_LEVEL, 32)
active_best_quality = 62, active_worst_quality = 128
q = 62
```

[...]

PR #901, opened 2023-06-16
Work around a qindex bug in libaom v2.0.2 or older

wantehchang:
In libaom v2.0.2 or older, the first frame is incorrectly using a qindex (quantizer index) lower than the provided cq_level.

This bug was fixed by the libaom changelist
https://aomedia-review.googlesource.com/c/aom/+/132881. The bug fix was cherry-picked to the libaom 2.0.x release branch ("applejack"), but v2.0.3 was never released.

Fix https://github.com/strukturag/libheif/issues/895.


farindk:
Thank you.

PR #903, opened 2023-06-19
Add sample image to fuzz testing corpus to help prevent buffer overflow when parsing region mask

lovell:
This PR adds a small safety net to the feature added recently in https://github.com/strukturag/libheif/pull/845 that ensures the referenced region mask exists before trying to use it.

This problem was discovered via fuzz testing. Please let me know if you'd like a sample image that would otherwise trigger a segfault when attempting to access `refs[0]`.


bradh (commented):
LGTM.

fancycode:
Thanks, feel free to add a (small) sample file to our fuzzing corpus:
https://github.com/strukturag/libheif/tree/master/fuzzing/data/corpus

This will be automatically picked up by oss-fuzz, too.

fancycode:
Did a quick pass-by review, shouldn't the check rather be when accessing the `refs` vector here:
https://github.com/strukturag/libheif/blob/67a27f834172dff164dd965e92b55dc25a31bca9/libheif/context.cc#L985

Currently you could have a `refs` vector with one element and still crash if there are two regions with referenced masks. Or maybe I'm missing something and this is already checked elsewhere.

lovell:
@fancycode Danke schön, I agree, I've moved the check into the loop, plus added a 1.2KB sample to the fuzzing corpus that would otherwise cause a segfault.

fancycode (approved):
Thanks, lgtm now!

Waiting for @farindk for final approval.

lovell:
The libvips fuzz tests found another example of this issue that goes a bit further than previous cases and causes a buffer overflow. Please can this fix (or something similar) be considered for inclusion before the next tagged release to avoid a CVE, thank you.

lovell:
Looks like commit https://github.com/strukturag/libheif/commit/6c4016fd3d944a8903bdd3e6be8fcdd1bc73ca79 fixed this, thank you. Probably still worth adding the sample image from this PR to the fuzz test corpus.

farindk:
Thanks, @lovell .
Sorry, I didn't realize that you had already submitted a PR for this issue. I just got triggered by a clusterfuzz email that I received yesterday.

lovell:
No worries, thanks for fixing/merging.

PR #915, opened 2023-06-29
Have CMakeLists.txt compile OpenJPEG as a plugin

dukesook:
Here are the next few steps to adding a JPEG2000 codec:
  - Have CMakeLists.txt compile OpenJPEG as a plugin if found
  - Once heif_context_encode_image() is called with a JPEG2000 encoder, navigate to the jpeg2000.cc file

The OpenJPEG library can be installed on Linux with the following command:
>$ sudo apt install libopenjp2-7-dev


libheif/plugins/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ macro(plugin_compilation name varName optionName defineName)
                     LIBHEIF_EXPORTS
                     HAVE_VISIBILITY)
             target_compile_definitions(heif-${name} PRIVATE PLUGIN_${defineName}=1)
-            target_include_directories(heif-${name} PRIVATE ${PROJECT_SOURCE_DIR} ${libheif_BINARY_DIR} ${libheif_SOURCE_DIR} ${${varName}_INCLUDE_DIR})
+            target_include_directories(heif-${name} PRIVATE ${PROJECT_SOURCE_DIR} ${libheif_BINARY_DIR} ${libheif_SOURCE_DIR} ${${varName}_INCLUDE_DIR} ${${varName}_INCLUDE_DIRS})
```
kmilos, line 23: Perhaps we should simply change `${varName}_INCLUDE_DIR` to `${varName}_INCLUDE_DIRS` instead of adding?

libheif/plugins/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ macro(plugin_compilation name varName optionName defineName)
             message("Compiling '" ${name} "' as built-in backend")
             target_sources(heif PRIVATE ${${optionName}_sources})
             target_compile_definitions(heif PRIVATE HAVE_${defineName}=1)
-            target_include_directories(heif PRIVATE ${${varName}_INCLUDE_DIR})
+            target_include_directories(heif PRIVATE ${${varName}_INCLUDE_DIR} ${${varName}_INCLUDE_DIRS})
```
kmilos, line 33: Ditto

CMakeLists.txt
```diff
@@ -156,6 +158,9 @@ if ((JPEG_DECODER_FOUND AND NOT WITH_JPEG_DECODER_PLUGIN) OR (JPEG_ENCODER_FOUND
     list(APPEND REQUIRES_PRIVATE "libjpeg")
 endif()
 # TODO: JPEG
+if (OpenJPEG_FOUND AND NOT WITH_OpenJPEG_PLUGIN)
+    list(APPEND REQUIRES_PRIVATE "OpenJPEG")
```
kmilos, line 162: Seems to be `libopenjp2` on my system for the pkgconf file.

libheif/plugins/CMakeLists.txt
kmilos: > If any of the libheif supplied custom FindXYZ.cmake modules does not produce the `XYZ_INCLUDE_DIRS` variable, [it should be fixed there](https://cmake.org/pipermail/cmake/2006-October/011502.html).

Indeed. Please fix those in all cmake/modules/FindXYZ.cmake modules: the `libfind_process(XYZ)` produces those, and they should be used in all subsequentl `find_package_handle_standard_args(XYZ)` calls.

kmilos:
Also, are you sure the CMake config files for OpenJPEG are shipped an all platforms for most recent versions? Would it be an idea to supply cmake/modules/FindOpenJPEG.cmake as well just in case?

libheif/plugins/CMakeLists.txt
dukesook: Thank you for your feedback, I will study up on this issue.

dukesook: I simply removed `${varName}_INCLUDE_DIRS` which restores those two lines to code to how they originally were. It turns out that they weren't necessary.

kmilos: It should be "fixed" anyway, but that's maybe for a separate PR.

farindk: @kmilos Yes, but let's do this in a different issue so that @dukesook can concentrate on the JPEG2000 parts. (#917)

CMakeLists.txt
dukesook: @kmilos: Are you suggesting this instead?
```cmake  
if (OpenJPEG_FOUND AND NOT WITH_OpenJPEG_PLUGIN) 
    list(APPEND REQUIRES_PRIVATE "libopenjp2")

Comment (kmilos) [reply CMakeLists.txt:162]: Yep, thanks 👍

File: libheif/plugins/CMakeLists.txt (hunk 2 above)
Comment (dukesook) [reply libheif/plugins/CMakeLists.txt:33]: I'm actually still getting compilation errors without the `${varName}_INCLUDE_DIRS` variable. Perhaps I can leave it for now and a future Pull Request can address this issue more thoroughly.

Comment (farindk) [reply libheif/plugins/CMakeLists.txt:33]: Yes, leave it. I will take care of it later.

File: libheif/jpeg2000.cc [hunk 4]
```diff
@@ -286,3 +286,15 @@ std::string Box_j2kH::dump(Indent& indent) const
...
+
+Error Jpeg2000ImageCodec::encode_jpeg2000_image(const std::shared_ptr<HeifFile>& heif_file,
+                                                const std::shared_ptr<HeifPixelImage>& src_image,
+                                                void* encoder_struct,
+                                                const struct heif_encoding_options& options,
+                                                std::shared_ptr<HeifContext::Image>& out_image) 
+{
+    return Error(heif_error_Unsupported_feature,
+                heif_suberror_Unsupported_codec,
+                "JPEG2000 Encoding has not be implemented yet");
```
bradh, line 298:
```suggestion
                "JPEG2000 Encoding has not been implemented yet");
```

libheif/plugins/encoder_openjpeg.cc
```diff
@@ -0,0 +1,180 @@
...
+  return error_Ok;
+}
+
+void opj_query_input_colorspace(enum heif_colorspace* inout_colorspace, enum heif_chroma* inout_chroma) {
+}
+
+
+
+struct heif_error opj_encode_image(void* encoder_raw, const struct heif_image* image, enum heif_image_input_class image_class) {
+  return heif_error {heif_error_Unsupported_feature, heif_suberror_Unsupported_codec, "JPEG2000 Encoding has not be implemented yet"};
```
bradh, line 127:
```suggestion
  return heif_error {heif_error_Unsupported_feature, heif_suberror_Unsupported_codec, "JPEG2000 Encoding has not been implemented yet"};
```

libheif/plugins/encoder_openjpeg.cc
```diff
@@ -0,0 +1,180 @@
...
+
+}
+
+struct heif_error opj_get_compressed_data(void* encoder_raw, uint8_t** data, int* size, enum heif_encoded_data_type* type) {
+  return heif_error {heif_error_Unsupported_feature, heif_suberror_Unsupported_codec, "JPEG2000 Encoding has not be implemented yet"};
```
bradh, line 132:
```suggestion
  return heif_error {heif_error_Unsupported_feature, heif_suberror_Unsupported_codec, "JPEG2000 Encoding has not been implemented yet"};
```

farindk:
I have merged this into the `jpeg2k` branch.
Encoding just gives me a segmentation fault. Maybe I broke something when resolving all the merging conflicts, but most probably it's somewhere in the encoding plugin because the encoded data that is returned looks suspicious...

farindk:
@dukesook I have worked on improving the JPEG-2000 code in the `jpeg2000` branch, encoding and decoding basically work. RGB is currently hardcoded.
I tried to change this, but couldn't find where the colorspace should actually be stored in the file.
The HEIF boxes only specify the component mapping and according to the standard, the colorspace should be stored in the `Color Specification Box`, but that is only defined for the JP2 file format.

Any insights on this?

farindk:
@bradh FYI

farindk:
I'm also confused by the requirement of ISO/IEC 15444-16 that each image item should have exactly one `colr` box. That either contains an nclx profile (only usable for YCbCr images) or an ICC profile (for RGB images). Should that box be 'misused' for detecting YCbCr vs. RGB? And what about greyscale images where `colr` doesn't make any sense?

Either I am misunderstanding a few things, or the standard has severe defects.

farindk:
Ok, I think I found it. In 15444-16, when they talk about `colr` box, they probably do not mean the HEIF `colr` box, but the JP2 `colr` box. Same name, different content.
That would mean, that this is also wrong in the example image, which has a HEIF-`colr` box, but no JP2-`colr` box.

dukesook:
@farindk, Thanks for working on the jpeg2000 code.  The `jpeg2k` branch started to diverge from the `master` branch so I'm not surprised to see merging issues, but it looks like you already took care of them. 

I too was confused with the two different `colr` boxes, I will bring this up with a subject matter expert as well as where the colorspace should be located.

palemieux:
In a 15444-16 file, colorspace information is signaled in the `colr` box carried in the `ItemPropertyContainerBox` (see ISO/IEC 23008-12).

farindk:
I don't see how the HEIF-`colr` box can give us this information.

When I'm encoding images with OpenJPEG, I can select between several colorspaces
(sRGB, gray, SYCC, EYCC, CMYK). And also Table I-18 in ISO 15444-1 (or ITU T.800) gives me the impression
that there several options.

But where in the file can we specify which colorspace is used? The 'cdef' box just maps the channels to
IDs that change their meaning depending on the colorspace. E.g. Asoc=2 either means "green" in RGB
or "Cb" in YCbCr.

Whether RGB or YCbCr (or greyscale, ...) is used should be specified in the J2K "Colour Specification Box",
confusingly also named 'colr'. Specifically the field EnumCS (Table I-10).

The HEIF 'colr' doesn't give me this information. It is either an 'nclx', which is only meaningful when
we use a YCbCr-like space, or an ICC.

How should that fit together? How should an RGB image file look like versus an YCbCr file?

palemieux:
> How should that fit together? How should an RGB image file look like versus an YCbCr file?

AFAIK the values for `colour_primaries`, `transfer_characteristics`, `matrix_coefficients` and `full_range_flag` are intended to be selected from those from [H.273](https://www.itu.int/rec/T-REC-H.273-202107-I/en).

EXAMPLES:

* full-range RGB image that uses the colour primaries and the PQ transfer function specified at [ITU-R-BT.2100]: `9 16 0 1`
* narrow-range YCbCr image that hat uses the colour primaries and the transfer function defined at [ITU-R-BT.709]: `1 1 1 0`


Makes sense?

farindk:
Yes, that would work for the moment, but it basically means that we define that the J2K data is **always** YCbCr, since H.273 is assuming YCbCr input only.
To me, this looks very inelegant for a couple of reasons:
- we will lose the capability to store other colorspaces like CMYK, which I thought would be a main advantage of the JPEG2000 codec,
- if the input is always YCbCr, why do we need the 'cdef' box? We could as well simply say that the channels are Y,Cb,Cr,(Alpha) in this order,
- ISO 15444-16 requires that a `colr` is always present (which overrides the 'optional' in HEIF, BTW). What should it contain for greyscale images? Are J2K greyscale images allowed, after all?
- ISO 15444-16 only says that a 'colr' should be present. It says nothing about what it should contain. It could also contain an ICC. What's the default `nclx` in this case? matrix-coefficients=0 or not?
- Assuming that the nclx contains matrix-coefficients=0, should `cdef` use the mapping R=1, G=2, B=3 or G=1, B=2, R=3, since it effectively is remapped by H.273 (eq. 41-43)?

A cleaner approach, in my view, would be to have a J2K-`colr` box in the `j2KH` container to specify the colorspace. In order to stay compatible to the current definition (if that is already in use at all), this could be optional, with YCbCr the default if not present.

With that proposal, we can still use YCbCr+nclx (as basically every other codec uses), but we had a clean way to store greyscale and even RGB+ICC without any nclx hack, and would be open to other colorspaces (CMYK).

PS: I'm not sure whether the table I-10 (EnumCS) has been extended after 2002. I don't have the 2019 version of ISO 15444.1,

palemieux:
> Yes, that would work for the moment, but it basically means that we define that the J2K data is always YCbCr, since H.273 is assuming YCbCr input only.

I do not think this is accurate: RGB input is signaled by setting `matrix_coefficients = 0`.

... or maybe you mean something else by _H.273 is assuming YCbCr input only_?

farindk:
> > Yes, that would work for the moment, but it basically means that we define that the J2K data is always YCbCr, since H.273 is assuming YCbCr input only.
> 
> I do not think this is accurate: RGB input is signaled by setting `matrix_coefficients = 0`.
> 
> ... or maybe you mean something else by _H.273 is assuming YCbCr input only_?

See H.273, Eq. 41-43.  matrix-coefficients is renaming YCbCr to RGB. I.e. input YCbCr, output RGB.

palemieux:
> See H.273, Eq. 41-43. matrix-coefficients=0 is renaming YCbCr to RGB. I.e. input YCbCr, output RGB.

See Table 4, row 1: _typically used for GBR (often referred to as RGB)_

H.265, E.3.1 makes it clear that RGB is signaled using `matrix-coefficients=0`:

image

I think Eq. 41-43 are probably just misleading. It is the first I hear that H.273 cannot signal RGB :) We can triple-check.

farindk:
Yes, I agree that matrix-coefficients=0 is basically RGB. We use that in libheif for already for years.
My point is that matrix-coefficients=0 technically does not specify that its **input** is RGB, it remaps its YCbCr input to RGB **output**. I.e. for an (RGB) image with matrix-coefficients=0, we would have to set the 'cdef' to R=3,G=1,B=2 because that is still YCbCr (=GBR) before the remapping.
If matrix-coefficients=0 would mean that its **input** is RGB, we would need to set 'cdef' to R=1, G=2, B=3.

However, the above is just a pedantic detail.
The main question is: should JPEG-2000 in HEIF really be limited to this? It's giving away a lot of its potential.

And what about the other points I listed? Greyscale, for example?

palemieux:
> The main question is: should JPEG-2000 in HEIF really be limited to this? It's giving away a lot of its potential.

I agree. Thanks for asking the question.

15444-16 was written to only use 23008-12 tools and with the assumption that `ItemPropertyContainerBox -> colr` was the only method to signal color information in 23008-12. Is that a correct assumption?

If not, then what other mechanism exists?

If so, I see three options:

- modify 23008-12 to support a broader range of color/pixel characteristics
- modify 15444-16 to allow either 14496-12 `colr` or 15444-1/-2/-15 `colr`
- not use HEIF if color cannot be signaled using H.273

Any other option? Preference?

farindk:
> 15444-16 was written to only use 23008-12 tools and with the assumption that `ItemPropertyContainerBox -> colr` was the only method to signal color information in 23008-12. Is that a correct assumption?

Yes. All current codecs in HEIF assume that images are encoded as YCbCr (this includes RGB through nclx,m.coeff=0).
One notable exception is ISO23001-17 (draft), which defined other channel types, such as X-ray, IR, even though these IDs have been removed from the current draft as far as I know.

Considering that there are a lot of J2K-DICOM images out there with X-ray content, this would be an important use case.

> If so, I see three options:
>     * modify 23008-12 to support a broader range of color/pixel characteristics
>     * modify 15444-16 to allow either 14496-12 `colr` or 15444-1/-2/-15 `colr`
>     * not use HEIF if color cannot be signaled using H.273
> Any other option? Preference?

I think a necessary change is to lift the "file must contain a HEIF-`colr`" box constraint in 15444-15. This constraint will limit the use forever to YCbCr/RGB. And it is unclear to me what it should even mean for greyscale images. This includes cases in which greyscale images are used as auxiliary images like 'depth', 'alpha', ...

Moreover, adding a note to 15444-15 would be nice stating that the image data is to be interpreted as YCbCr if not indicated differently. (This still allows RGB through nclx.)

Then, in a next step, we can think whether it would be useful to additionally support 15444-1-'colr' or extend the HEIF-'colr'.

palemieux:
> I think a necessary change is to lift the "file must contain a HEIF-colr" box constraint in 15444-15.

What is the expected behavior of 23008-12 viewers in the absence of the `colr` box?

farindk:
> What is the expected behavior of 23008-12 viewers in the absence of the colr box?

Use any YCbCr->RGB transform, usually the coefficients from JPEG (ITU_R_BT_601_6). It is not specified in the standard.

palemieux:
> Use any YCbCr->RGB transform, usually the coefficients from JPEG (ITU_R_BT_601_6). It is not specified in the standard.

That is only for 4:2:2, right? If so, is Rec. 709/sRGB assumed for 4:4:4?

palemieux:
In any case, I like the idea of considering allowing the `colr` box from 15444-1/-2/15. How would we differentiate it from the `colr` box from 14496-12 since they both have the same FourCC?

farindk:
> > Use any YCbCr->RGB transform, usually the coefficients from JPEG (ITU_R_BT_601_6). It is not specified in the standard.
> 
> That is only for 4:2:2, right? If so, is Rec. 709/sRGB assumed for 4:4:4?

It's for all chroma formats. But like said, it's not specified in the standard what to do.




> In any case, I like the idea of considering allowing the `colr` box from 15444-1/-2/15. How would we differentiate it from the `colr` box from 14496-12 since they both have the same FourCC?

HEIF-`colr` is an image property, while 15444-1-`colr` is in the `j2kH` box.

palemieux:
> HEIF-colr is an image property, while 15444-1-colr is in the j2kH box.

Understood.

15444-16 already allows the `colr` box inside the `j2kH` box (_All boxes are optional unless specified otherwise, and boxes other than those listed above can be present_), so no harm in experimenting with this today.

In the meantime, I plan to (a) confirm the interpretation above of the absence of the `ItemPropertyContainerBox::colr` and (b) understand if there are any drawbacks in the including J2K `colr` in  `j2kH`.

Ok?

farindk:
Yes.

Please let me know the current version of Table I-10 in 15444-1 (2019). I only have access to the 2002 version and in that version, Table I-10 just lists YCbCr, RGB, and Greyscale. Thus, if it is still like this, we won't gain much of including the 15444-1-`colr` box at this time.

palemieux:
See your inbox.

> Please let me know the current version of Table I-10 in 15444-1 (2019). I only have access to the 2002 version and in that version, Table I-10 just lists YCbCr, RGB, and Greyscale.

We would get grayscale at least, right?

Were you thinking XRay, Infrared and other exotic channels?

farindk:
> > Please let me know the current version of Table I-10 in 15444-1 (2019). I only have access to the 2002 version and in that version, Table I-10 just lists YCbCr, RGB, and Greyscale.
> 
> We would get grayscale at least, right?

Greyscale could be the default for one-channel images, but yes, it would be safer to explicitly state this.
Yes, I thought about the more exotic channels (while I don't think X-ray is so exotic given the amount of DICOMs).
But also CMYK or Lab.

palemieux:
> But also CMYK or Lab.

CMYK is `EnumCS = 12`.

CIELab is `EnumCS = 14`.

image

farindk:
Yes, I think this looks good. It still doesn't define the exotic channel types, but what was important to me is that there is no theoretical limit. If applications will need it, it is possible to define them.

palemieux:
> If applications will need it, it is possible to define them.

Yes. That is very much the intent.

farindk:
Good. Thanks for discussing this. I'll do the necessary changes to libheif to write/read images according to what we discussed. I'll let you know when it's updated.

farindk:
@dukesook I have rewritten the J2K plugin such that encoding is done in YCbCr space or greyscale. Tested all chroma formats. RGB encoding through `--matrix_coefficients=0 --chroma=444`.
It seems to be reasonably stable now, so I have merged this into `master`.

farindk:
@dukesook Note: because of the change of semantics, your test image `lunar_eclipse.hej2` is not valid anymore.

dukesook:
I apologize, but haven't revised a test file yet. I'll do this sometime down the road.

palemieux:
@dukesook @farindk I would appreciate your feedback (before December 15) on the proposed revision of the [J2K in HEIF specification](https://ds.jpeg.org/documents/jpeg2000/wg1n100678-101-ICQ-WD_15444-16_3ED.pdf). This revision allows the presence of zero or more Colour Specification boxes, as defined in Rec. ITU-T T.801 | ISO/IEC 15444-2, in the JPEG 2000 header item. If multiple Colour Specification boxes are present, the decoder can choose the one it believes is most appropriate.

bradh:
@palemieux how would you like comments?

palemieux:
> how would you like comments?

Couple of non-mutually-exclusive options:

- create a new issue in this repo to capture and discuss comments on the spec
- join the J2K discord (https://discord.gg/2nZMtqYXud)
- email me privately at pal@sandflow.com (I chair the J2K Maintenance Activity and can help convey the comments)

Looking forward to them.

Issue #917, opened 2023-06-30
CMake: use {libname}_INCLUDE_DIRS instead of {}_INCLUDE_DIR everywhere


farindk:
See https://github.com/strukturag/libheif/pull/915#discussion_r1247530176

PR #918, opened 2023-07-03
CMake: use recommended INCLUDE_DIRS form

kmilos:
Resolves https://github.com/strukturag/libheif/issues/917


kmilos:
Could potentially drop the changes in all the FindXYZ.cmake files as superfluous, the plural variable is guaranteed by the `libfind_process()` call already...

kmilos:
There is also no need for explicit addition of link library paths (those should've been added using `target_link_directories()` anyway), as "find" returns the full path to libraries...

kmilos:
Dropped find module changes, keeping just the minimal required now.

PR #923, opened 2023-07-12
Fix build of libheif plugins on Darwin

Torrekie:
As Darwin does not treat "loadable modules" (aka. bundles) as standalone libraries, neither identifiers nor version infos can be included.

This patch removes the version info for Darwin builds so plugins can be built correctly.

### DIFF
--- a/libheif/plugins/CMakeLists.txt
+++ b/libheif/plugins/CMakeLists.txt
@@ -11,10 +11,12 @@ macro(plugin_compilation name varName optionName defineName)
                     MODULE ${${optionName}_sources} ${${optionName}_extra_plugin_sources}
                     ../heif_plugin.cc
                     )
-            set_target_properties(heif-${name}
-                    PROPERTIES
-                    VERSION ${PROJECT_VERSION}
-                    SOVERSION ${PROJECT_VERSION_MAJOR})
+            if (NOT APPLE)
+                set_target_properties(heif-${name}
+                        PROPERTIES
+                        VERSION ${PROJECT_VERSION}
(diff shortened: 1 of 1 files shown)


Torrekie:
```
[ 69%] Linking CXX shared module libheif-x265.so
cd /buildroot/libheif-1.16.2/build/libheif/plugins && /usr/bin/cmake -E cmake_link_script CMakeFiles/heif-x265.dir/link.txt --verbose=1
/usr/bin/c++ -O3 -DNDEBUG -arch arm64 -bundle -Wl,-headerpad_max_install_names -compatibility_version 1.0.0 -current_version 1.16.2 -o libheif-x265.so "CMakeFiles/heif-x265.dir/encoder_x265.cc.o" "CMakeFiles/heif-x265.dir/__/heif_plugin.cc.o"  -Wl,-rpath,/buildroot/libheif-1.16.2/build/libheif /usr/lib/libx265.dylib ../libheif.1.16.2.dylib 
clang-14: error: invalid argument '-compatibility_version 1.0.0' only allowed with '-dynamiclib'
make[2]: *** [libheif/plugins/CMakeFiles/heif-x265.dir/build.make:118: libheif/plugins/libheif-x265.so] Error 1
```

Torrekie:
MacPorts met same issue https://github.com/macports/macports-ports/blob/ca2ac6c3148c701a4fc0a96c69d89e980034cde3/multimedia/libheif/Portfile#L38

farindk:
Thank you for the additional information. Makes sense. I hoped that this would be handled by cmake as the plugin is specified as `MODULE`. Apparently it is not.

Issue #924, opened 2023-07-15
Emscripten doesn't compile


jerbob92:
I'm trying to compile the project using the `build-emscripten.sh` script to compile to WebAssembly. My goal is to build it into a server-side WebAssembly so that it can be directly embedded in go-libheif (without using CGO). 

However, the script always results in the following errors, I have tried multiple versions, including the version in the install script and the latest version, am I doing something wrong?

```
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_class
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_class
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_class
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_bindings
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_function
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_enum
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_enum_value
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_enum
wasm-ld: error: ../libheif/libheif.a(heif.cc.o): undefined symbol: _embind_register_enum_value
[...]

PR #925, opened 2023-07-18
Emscripten update and standalone support

jerbob92:
Related to #924 

This MR:

- Updates Emscripten to the latest SDK version 3.1.43 (from 1.37.26)
- Uses the default Emscripten backend (LLVM) and not the old one anymore (fastcomp)
- Updates the libde265 version that was included in the Emscripten build to 1.0.12 (from 1.0.8)
- Adds the following options to the `build-emscripten.sh` script:
  - `CORES` (default value: all available cores)
  - `ENABLE_LIBDE265` (whether to include libde265, default = 1)
  - `LIBDE265_VERSION` (which version of libde265 to include, default = 1.0.12)
  - `ENABLE_AOM` (whether to include aom, default = 0)
  - `AOM_VERSION` (which version of aom to include, default = 3.6.1)
  - `STANALONE` (whether to build a standalone WASM that runs on non-web/serverside WASM runtimes)
  - `DEBUG` (whether to create a debug build, sets the Emscripten flags `--profile` and `-g`)
- Fixes the pre.js/post.js to work with the new Emscripten version, since the Embind functions were not available yet on load in the previous implementation.


build-emscripten.sh
```diff
@@ -2,52 +2,114 @@
...
-CONFIGURE_ARGS="-DENABLE_MULTITHREADING_SUPPORT=OFF -DWITH_GDK_PIXBUF=OFF -DWITH_EXAMPLES=OFF -DBUILD_SHARED_LIBS=ON"
-#export PKG_CONFIG_PATH="${DIR}/libde265-${LIBDE265_VERSION}"
+EXTRA_EXE_LINKER_FLAGS="-lembind"
+EXTRA_COMPILER_FLAGS=""
+if [ "$STANDALONE" = "1" ]; then
+    EXTRA_EXE_LINKER_FLAGS=""
+    EXTRA_COMPILER_FLAGS="-D__EMSCRIPTEN_STANDALONE_WASM__=1"
+fi
 
+CONFIGURE_ARGS="-DENABLE_MULTITHREADING_SUPPORT=OFF -DWITH_GDK_PIXBUF=OFF -DWITH_EXAMPLES=OFF -DBUILD_SHARED_LIBS=ON -DENABLE_PLUGIN_LOADING=OFF"
```
jerbob92, line 78: I added `-DENABLE_PLUGIN_LOADING=OFF`, I thought it would make sense.

build-emscripten.sh
```diff
@@ -2,52 +2,114 @@
...
+    -DCMAKE_CXX_FLAGS="${EXTRA_COMPILER_FLAGS}" \
+    -DCMAKE_EXE_LINKER_FLAGS="${LIBRARY_LINKER_FLAGS} ${EXTRA_EXE_LINKER_FLAGS}" \
+    $CONFIGURE_ARGS_LIBDE265 \
+    $CONFIGURE_ARGS_AOM
+
 emmake make -j${CORES}
 
-export TOTAL_MEMORY=16777216
+LIBHEIFA="libheif/libheif.a"
+EXPORTED_FUNCTIONS=$($EMSDK/upstream/bin/llvm-nm $LIBHEIFA --format=just-symbols | grep "^heif_\|^de265_\|^aom_" | grep "[^:]$" | sed 's/^/_/' | paste -sd "," -)
```
jerbob92, line 89: This is necessary to make the C API's be exposed, otherwise they won't be available exported the WASM/JS code.

build-emscripten.sh
```diff
@@ -2,52 +2,114 @@
...
-    -s ASSERTIONS=0 \
-    -s INVOKE_RUN=0 \
-    -s DOUBLE_MODE=0 \
-    -s PRECISE_F32=0 \
-    -s DISABLE_EXCEPTION_CATCHING=1 \
-    -s USE_CLOSURE_COMPILER=0 \
-    -s LEGACY_VM_SUPPORT=1 \
-    --memory-init-file 0 \
-    -O3 \
+    -s ERROR_ON_UNDEFINED_SYMBOLS=0 \
```
jerbob92, line 109: I have removed all the old flags here, I don't think they are neccesary.

libheif/heif.cc
```diff
@@ -33,7 +33,7 @@
 #include <set>
 #include <limits>
 
-#if defined(__EMSCRIPTEN__)
+#if defined(__EMSCRIPTEN__) && !defined(__EMSCRIPTEN_STANDALONE_WASM__)
```
jerbob92, line 36: Note: `__EMSCRIPTEN_STANDALONE_WASM__` isn't something that Emscripten provides, I add this using `CMAKE_C_FLAGS` / `CMAKE_CXX_FLAGS`. Embind is not supported on non-web runtimes since it heavily relies on JS, so we do not want the Embind things to be injected in the standalone build.

libheif/heif_emscripten.h
```diff
@@ -112,7 +112,7 @@ static emscripten::val heif_js_decode_image(struct heif_image_handle* handle,
   result.set("width", width);
   int height = heif_image_get_height(image, heif_channel_Y);
   result.set("height", height);
-  std::string data;
+  std::basic_string<unsigned char> data;
```
jerbob92, line 115: This was needed because Embind was messing up the data transfer when using a normal std::string, however, the data transfer between strings looks very odd to me right now because I think what currently happens is:
- WASM sends pointer to string to JS
- JS reads binary data from pointer position
- JS converts binary data into string
- libheif converts string back to binary data (`post.js`)

I believe we can remove the string conversion, probably also speeds things up a bit. I'm currently talking to Emscripten what the best way to do that is in Embind.

post.js
```diff
@@ -175,40 +175,6 @@ var libheif = {
     }
 };
 
-var key;
```
jerbob92, line 178: This has all moved to pre.js

pre.js
```diff
@@ -29,5 +29,47 @@ var Module = {
         console.error(text);
     },
     canvas: {},
-    noInitialRun: true
+    noInitialRun: true,
+    onRuntimeInitialized: function() {
+        // Expose enum values.
+        var enums = {
```
jerbob92, line 35: I had to move all this logic from `post.js` to `pre.js` inside the `onRuntimeInitialized` function. This is because these properties are not available yet in `post.js` since they are loaded later using `embind`, `onRuntimeInitialized` is triggered after loading `embind`. The interface still works the same because of this trick.

pre.js
```diff
@@ -29,5 +29,47 @@ var Module = {
...
+        // Expose internal C API.
+        for (key in this) {
+            if (enums.hasOwnProperty(key.slice(1)) || key.indexOf("_heif_") !== 0) {
+                continue;
+            }
+            libheif[key.slice(1)] = this[key];
+        }
+
+        // Expose embind API.
+        for (key in Module) {
```
jerbob92, line 68: I alias the embind API after the C API because there are two functions that have the same name as in the C API (`heif_context_read_from_memory` and `heif_get_version`), we prefer the embind function over the C API function.

scripts/install-emscripten.sh
```diff
@@ -33,13 +32,13 @@ if [ ! -d emsdk ]; then
 fi
 
 cd emsdk
-echo "Updating SDK base to ${EMSDK_VERSION} ..."
+echo "Updating SDK base to ${VERSION} ..."
```
jerbob92, line 35: I have removed the `EMSDK_VERSION` and use `VERSION` now, the version that gets installed is always latest.

build-emscripten.sh
farindk: When not building to WASM, but plain JS, `--memory-init-file 0` avoids the generation of an extra file `libheif.js.mem`. Thus we may keep this to stay compatible to existing applications.

build-emscripten.sh
```diff
@@ -90,12 +91,12 @@ EXPORTED_FUNCTIONS=$($EMSDK/upstream/bin/llvm-nm $LIBHEIFA --format=just-symbols
...
 echo "Running Emscripten..."
 
-BUILD_FLAGS="-lembind -o libheif.js --pre-js pre.js --post-js post.js"
+BUILD_FLAGS="-lembind -o libheif.js --pre-js pre.js --post-js post.js -sWASM=$USE_WASM"
 RELEASE_BUILD_FLAGS="-O3"
 
 if [ "$STANDALONE" = "1" ]; then
     echo "Building in standalone (non-web) build mode"
-    BUILD_FLAGS="-s STANDALONE_WASM=1 -s WASM=1 -o libheif.wasm --no-entry"
+    BUILD_FLAGS="$BUILD_FLAGS -s STANDALONE_WASM=1 -s WASM=1 -o libheif.wasm --no-entry"
```
jerbob92, line 99: @farindk I specifically did not extend `BUILD_FLAGS` in standalone build because if you do, it will include embind imports in the generated WASM.

build-emscripten.sh
jerbob92: @farindk Ah I did not even know it was possible to only emit JS, was that what the previous version had as output?

build-emscripten.sh
farindk: Ok. I'll undo that. I must admit that I didn't really understand the 'standalone' option.

build-emscripten.sh
farindk: Yes, previously, it was a single `libheif.js` file. Thus, we have to still support this that other software that depends on this, continues to work.

build-emscripten.sh
jerbob92: @farindk [Standalone](https://github.com/emscripten-core/emscripten/wiki/WebAssembly-Standalone) builds allow you to run the generated WASM outside the browser in WebAssembly runtimes such as [Wasmer](https://wasmer.io/), [WAVM](https://github.com/WAVM/WAVM), or [wasmtime](https://github.com/CraneStation/wasmtime). Personally I'm using the standalone build in the [Wazero](https://github.com/tetratelabs/wazero) runtime to be used in the go-libheif library, this allows for a go library that is fully contained (because the WASM is embedded in the library) and doesn't need external native libraries, it also allows the library to run fully sandboxed. The only downside is that it's slower than CGO, and multithreaded support is pretty bad right now, which causes the decoding to be around 3x as slow as CGO.

build-emscripten.sh
jerbob92: @farindk Shouldn't `USE_WASM` have a default value of `0` then?

farindk: Good point. But I think in this case, I would take the risk to change the default because WASM is clearly the future and it doesn't change anything for the end-user. Just of the ones building new libheif.js files.
I am currently thinking about even doing larger changes, so it will be worth for them having a look anyways.

libheif/heif_emscripten.h
farindk: Converning passing the data around: do you think [this (answer item 3)](https://stackoverflow.com/questions/53602955/using-emscripten-how-to-get-c-uint8-t-array-to-js-blob-or-uint8array) might work?

build-emscripten.sh
jerbob92: I have notified the following repo of the change: https://github.com/catdad-experiments/libheif-emscripten/issues/19

What kind of other changes do you want to make? I think it might be interesting to look at thread support (for the browser it's possible).

libheif/heif_emscripten.h
jerbob92: Yes! `result.set("data", emscripten::val(emscripten::typed_memory_view(data.size(), data.data())));` seems to work perfectly! It will give you a `Uint8Array` directly in JS, however, I'm not completely sure who then becomes owner of the data buffer and who has to free it, would have to be looked into to prevent memory leaks.

fancycode:
While this might improve things, it broke the Emscripten CI tests (already in this PR) :cry: So from a testing point of view, we don't detect if some further change breaks Emscripten.

I'll add an issue to fix the Emscripten CI.

PR #932, opened 2023-07-23
uncompressed: fix component decode, simplify logic

### LINKED ISSUE #904: uncompressed decode for component order is incorrect
The current logic only copies the first row of pixels. That works for the tests, because its vertical bars, but is not valid in general.

I am working on a revised approach that will also handle subsampling (422, 420).

bradh:
Resolves #904

The previous logic worked really hard to copy into the destination image in pixel order, but was complicated and buggy. The new logic works in source byte order, which is much simpler. The interleave types now mostly reduce to different ordering of the `for(...)` loops.  I expect this will be extensible to handle row padding and blocking, likely with shared code, possibly an inline function.

Also adds tile-component interleave decoding, with small samples generated by gpac uncvg.


bradh:
The test files are no longer valid against the FDIS.

PR #938, opened 2023-08-08
box: add missing break in switch

bradh:
Looks like a bad merge.


farindk:
Yes, apparently an error during the merge.
Thanks.

PR #944, opened 2023-08-20
Ensure correct JPEG2000 SIZ segment bounds check

lovell:
This PR offers a possible improvement to the JPEG2000 SIZ segment bounds check introduced in https://github.com/strukturag/libheif/commit/7bf1cdf56b1a44c012b25dfeb1efdb82f289ac49 to ensure the data is long enough to read out all of the components.

It also adds a sample file discovered via fuzz testing that will currently trigger a buffer overflow.


lovell:
@farindk This is currently the only other (known) buffer overflow in libheif found via the libvips fuzzers.

PR #945, opened 2023-08-22
Remove legacy JPEG2000 encoder code

kmilos:


silverbacknet:
Jpeg2000 support was just barely added, and it is a valid format that can exist in HEIF files. Why do you think it needs to be removed?

Besides, you missed a bunch of things, like the cmakelist.

kmilos:
> Why do you think it needs to be removed?

Please note the "legacy code". These are redundant parts, as both new JPEG2000 decoder and encoder source files have indeed been added recently.

silverbacknet:
Oh, the openjpeg that's built-in instead of the outside library. I'm sorry, totally misread that diff.

farindk:
Thanks

PR #948, opened 2023-08-22
Use portable types to avoid cast warnings

kmilos:
NB: Requires API version bump (at least when it comes to Windows)

Can of course be solved the other way (explicit casting to `unsigned long` where needed), if deemed safe.

This is because LLP64 on Windows vs LP64 everywhere else...

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -3685,7 +3685,7 @@ struct heif_error heif_region_get_referenced_mask_ID(const struct heif_region* r
   return heif_error_invalid_parameter_value;
 }
 
-unsigned long int heif_region_get_inline_mask_data_len(const struct heif_region* region)
+size_t heif_region_get_inline_mask_data_len(const struct heif_region* region)
 {
   const std::shared_ptr<RegionGeometry_InlineMask> mask = std::dynamic_pointer_cast<RegionGeometry_InlineMask>(region->region);
   if (mask) {
--- a/examples/heif_info.cc
+++ b/examples/heif_info.cc
@@ -567,7 +567,7 @@ int main(int argc, char** argv)
             int32_t y;
             uint32_t w;
             uint32_t h;
-            long unsigned int data_len = heif_region_get_inline_mask_data_len(regions[j]);
+            size_t data_len = heif_region_get_inline_mask_data_len(regions[j]);
             std::vector<uint8_t> mask_data(data_len);
             heif_region_get_inline_mask_data(regions[j], &x, &y, &w, &h, mask_data.data());
             printf("      inline mask [x=%i, y=%i, w=%u, h=%u, data len=%lu]\n", x, y, w, h, mask_data.size());
--- a/libheif/heif.h
+++ b/libheif/heif.h
@@ -2593,7 +2593,7 @@ struct heif_error heif_region_get_referenced_mask_ID(const struct heif_region* r
  * @return the number of bytes in the mask data, or 0 on error.
  */
 LIBHEIF_API
-unsigned long heif_region_get_inline_mask_data_len(const struct heif_region* region);
+size_t heif_region_get_inline_mask_data_len(const struct heif_region* region);
 
 
 /**
--- a/libheif/uncompressed_image.cc
+++ b/libheif/uncompressed_image.cc
@@ -705,13 +705,13 @@ Error UncompressedImageCodec::decode_uncompressed_image(const std::shared_ptr<co
       uint8_t* dst = img->get_plane(channels[c], &stride);
       for (uint32_t row = 0; row < height; row++) {
         long unsigned int tile_row_idx = row % tile_height;
-        long unsigned int tile_row_offset = tile_width * tile_row_idx * channels.size();
+        size_t tile_row_offset = tile_width * tile_row_idx * channels.size();
         uint32_t col = 0;
         for (col = 0; col < width; col++) {
           long unsigned int tile_base_offset = get_tile_base_offset(col, row, uncC, channels, width, height);
           long unsigned int tile_col = col % tile_width;
-          long unsigned int tile_offset = tile_row_offset + tile_col * pixel_stride + pixel_offset;
-          long unsigned int src_offset = tile_base_offset + tile_offset;
+          size_t tile_offset = tile_row_offset + tile_col * pixel_stride + pixel_offset;
+          size_t src_offset = tile_base_offset + tile_offset;
           uint32_t dstPixelIndex = row * stride + col;
           dst[dstPixelIndex] = src[src_offset];
         }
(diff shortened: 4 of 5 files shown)


libheif/heif.cc
```diff
@@ -3685,7 +3685,7 @@ struct heif_error heif_region_get_referenced_mask_ID(const struct heif_region* r
   return heif_error_invalid_parameter_value;
 }
 
-unsigned long int heif_region_get_inline_mask_data_len(const struct heif_region* region)
```
farindk, line 3688: Do you think it is safe to change the return type in the public API?
Strictly speaking, it probably is not, but on the other hand, it'll work in most cases and this function is probably currently not in use yet anywhere. (At least that I know of.)

kmilos: There's probably alternative workarounds for this particular instance if you don't want to change the API, which is understandable. OTOH, having platform dependent return types that can catch one out in the API is not ideal either. Since libheif bumps the API and ABI version every once in a while anyway, I'll let you decide if and when you want to do it.

farindk: So far, we have changed the API only in backwards compatible ways. That would break compatibility.

@fancycode You are more strict in these things than I am. I would be fine to silently change the return type as this function is probably not used in any application code (maybe only in private code of @bradh). But if you would insist in announcing an incompatible API change because of this, I would go for simply casting the return type.

kmilos: > So far, we have changed the API only in backwards compatible ways.

But I think you have changed the ABI in non-backwards compatible ways, so apps required rebuilds anyway in the past?

farindk: > > So far, we have changed the API only in backwards compatible ways.
> 
> But I think you have changed the ABI in non-backwards compatible ways, so apps required rebuilds anyway in the past?

Not that I remember. Excluding bugs like the `heif_init()` thing that just came up.

bradh: I'm OK with the change - the work is so developmental, everything is moving.

fancycode: As the public API for `heif_region_get_inline_mask_data_len` was not in a released version yet, I'm fine with changing it.

farindk: Right.

kmilos:
Ooops, looks like I missed one (or it came in in the meantime):
```
C:/msys64/home/kmilos/libheif/tests/region.cc:393:115: warning: conversion from 'std::vector<unsig
ned char>::size_type' {aka 'long long unsigned int'} to 'long unsigned int' may change value [-Wconv
ersion]
  393 |   err = heif_region_item_add_region_inline_mask_data(region_item1, 20, 50, 64, 3, mask_data.
data(), mask_data.size(), NULL);
      |
        ~~~~~~~~~~~~~~^~
```

farindk:
I have added it.

kmilos:
@farindk Unfortunately we missed another one (I usually built w/o the JPEG codec):
```
[79/96] Building CXX object libheif/CMakeFiles/heif.dir/plugins/decoder_jpeg.cc.obj
C:/msys64/home/kmilos/MINGW-packages/mingw-w64-libheif/src/libheif-1.17.0/libheif/plugins/decoder_
jpeg.cc: In function 'heif_error jpeg_decode_image(void*, heif_image**)':
C:/msys64/home/kmilos/MINGW-packages/mingw-w64-libheif/src/libheif-1.17.0/libheif/plugins/decoder_
jpeg.cc:146:64: warning: conversion from 'std::vector<unsigned char>::size_type' {aka 'long long uns
igned int'} to 'long unsigned int' may change value [-Wconversion]
  146 |   jpeg_mem_src(&cinfo, decoder->data.data(), decoder->data.size());
      |                                              ~~~~~~~~~~~~~~~~~~^~
```
This one actually comes from libjpeg-turbo, so an explicit cast is needed? Haven't checked the API of the IJG libjpeg for this call...

farindk:
Yes, seems that we have to cast it.

kmilos:
@bradh There are also multiple conversion warnings in the uncompressed codec (I haven't been building so far):
```
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc: In member function ‘virtual Error Box_uncC::write(StreamWriter&) const’:
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc:345:9: warning: conversion from ‘int’ to ‘uint8_t’ {aka ‘unsigned char’} may change value [-Wconversion]
  345 |   flags |= (m_components_little_endian ? 0x80 : 0);
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc:346:9: warning: conversion from ‘int’ to ‘uint8_t’ {aka ‘unsigned char’} may change value [-Wconversion]
  346 |   flags |= (m_block_pad_lsb ? 0x40 : 0);
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc:347:9: warning: conversion from ‘int’ to ‘uint8_t’ {aka ‘unsigned char’} may change value [-Wconversion]
  347 |   flags |= (m_block_little_endian ? 0x20 : 0);
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc:348:9: warning: conversion from ‘int’ to ‘uint8_t’ {aka ‘unsigned char’} may change value [-Wconversion]
  348 |   flags |= (m_block_reversed ? 0x10 : 0);
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc:349:9: warning: conversion from ‘int’ to ‘uint8_t’ {aka ‘unsigned char’} may change value [-Wconversion]
  349 |   flags |= (m_pad_unknown ? 0x08 : 0);
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc: In function ‘Error get_heif_chroma_uncompressed(std::shared_ptr<Box_uncC>&, std::shared_ptr<Box_cmpd>&, heif_chroma*, heif_colorspace*)’:
/home/runner/work/libheif/libheif/libheif/uncompressed_image.cc:474:18: warning: conversion from ‘int’ to ‘uint16_t’ {aka ‘short unsigned int’} may change value [-Wconversion]
  474 |     componentSet |= (1 << component_type);
      |     ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

bradh:
We can safely add casts for those. That said, it does seem like a pretty basic false positive if the compiler can't tell that 0x80 or 0 will be unchanged when assigned to a uint8_t.

kmilos:
> That said, it does seem like a pretty basic false positive if the compiler can't tell that 0x80 or 0 will be unchanged when assigned to a uint8_t.

True - that was w/ gcc 9.4 used by mingw CI action on older Ubuntu, and 13.2 here doesn't seem to complain, so it might not be necessary...

PR #952, opened 2023-09-04
add a few (void) to please some compilers

omar-polo:
Spotted building a user of libheif (imlib2) that uses `-Wstrict-prototypes`.


farindk:
Thank you. Good catch.

PR #955, opened 2023-09-10
uncompressed: update to reflect FDIS

bradh:
Key changes are in the cmpd and uncC boxes to increase component counts. This breaks all the test data, which has been regenerated from gpac.


farindk:
Thanks

PR #958, opened 2023-09-11
Update .gitignore for VS Code

kmilos:


homm:
This is not related to the library. You can use global gitignores for your environment.

farindk:
It'll also affect myself, but I admit that putting the IDE specific things into a global `.gitignore` file seems to be the better option. I would leave the `third-party/*` things in the file, because this is not IDE related, but the place where the dependencies will be built.

PR #976, opened 2023-10-09
CMake: bump min required, use imported targets

### LINKED ISSUE #975: Use CMake imported targets
Moving @kmilos comment from #973 into a issue of its own.

> maybe it's time to bump CMake minimum required version and simplifyby using imported targets ZLIB::ZLIB and PNG::PNG...

Do you know what minimum cmake version is required for that? I think the current upper limit might be cmake v3.16.3 (Ubuntu 20.04).

> 3.1 for ZLIB, 3.5 for PNG

kmilos:
Resolves #975


farindk:
Great. Thank you.

PR #978, opened 2023-10-10
Remove deprecated OpenJPEG struct member

kmilos:
See https://github.com/uclouvain/openjpeg/pull/1383


farindk:
Right, makes sense.

PR #980, opened 2023-10-11
cmake: do not export codec libraries to pkg-config when `PLUGIN_LOADING_SUPPORTED_AND_ENABLED` set

pszemus:
When built by CMake with `--preset=release-noplugins` all `WITH_{CODEC}_PLUGIN` variables are set to `ON` (via `CMakePresets.json`) and, as a result, `libheif.pc` file  does not mention detected build-in backends in `Requires.private` field.

Eg:
`cmake -S . -B build --preset=release-noplugins -DWITH_DAV1D=ON` output:
```
[...]
Not compiling 'x265' backend
Not compiling 'libde265' backend
Compiling 'dav1d' as built-in backend
Not compiling 'aomdec' backend
Not compiling 'aomenc' backend
Not compiling 'svtenc' backend
Not compiling 'rav1e' backend
Not compiling 'jpegdec' backend
Not compiling 'jpegenc' backend
Not compiling 'j2kdec' backend
Not compiling 'j2kenc' backend
Not compiling 'kvazaar' backend
Not compiling 'ffmpegdec' backend
Compiling in 'libsharpyuv'
```

`libheif.pc`:
```
[...]
Requires:
Requires.private: libsharpyuv
Libs: -L${libdir} -lheif
Libs.private: 
Cflags: -I${includedir}
```

`dav1d` is detected but not includes in `Requires.private` :disappointed: 

This PR makes use of `PLUGIN_LOADING_SUPPORTED_AND_ENABLED` variable when filling the `Requires.private` pkg-config field. 
Then the `libheif.pc` looks like this:
```
[...]
Requires:
Requires.private: dav1d libsharpyuv
Libs: -L${libdir} -lheif
Libs.private: 
Cflags: -I${includedir}
```


CMakeLists.txt
```diff
@@ -222,31 +222,33 @@ if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
 else()
     set(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
 endif()
-if (LIBDE265_FOUND AND NOT WITH_LIBDE265_PLUGIN)
+if (LIBDE265_FOUND AND (NOT PLUGIN_LOADING_SUPPORTED_AND_ENABLED OR NOT WITH_LIBDE265_PLUGIN))
     list(APPEND REQUIRES_PRIVATE "libde265")
 endif()
-if (X265_FOUND AND NOT WITH_X265_PLUGIN)
+if (X265_FOUND AND NOT (NOT PLUGIN_LOADING_SUPPORTED_AND_ENABLED OR WITH_X265_PLUGIN))
```
farindk, line 228: Looks like there is a NOT at the wrong place here.

pszemus: You're right. Copy-paste error. Will fix in a moment.

farindk:
Thanks. Looks ok. For my taste, I would find it easier to read when it is written as
````
if (X265_FOUND AND NOT (PLUGIN_LOADING_SUPPORTED_AND_ENABLED AND WITH_X265_PLUGIN))
````

I.e. `!(A && B)` instead of ` (!A || !B)`.

PR #982, opened 2023-10-12
add support for URI metadata type

bradh:
The version 2 and 3 of the `infe` box has an `item_type` field that has special case meaning for values of `mime` and `uri `.

The box definition and parsing handles those cases, and the C API exposes the `content_type` that is present for the special case of `mime`.

This PR exposes the `item_uri_type` that is present for the special case of `uri `. This is used to indicate special metadata types for the GIMI profile.

The API and implementation for `heif_image_handle_get_metadata_item_uri_type()`  is essentially the same as `heif_image_handle_get_metadata_content_type()`.

This PR also makes use of that value (if present) in the heif_info example.


farindk:
Thank you

farindk:
What is a GIMI profile?

PR #987, opened 2023-10-15
Fix building plugins

antonio-rojas:
Currently, plugins are never built because cmake doesn't search for the necessary dependencies


antonio-rojas:
Nevermind, I misunderstood the cmake options

Issue #985, opened 2023-10-14
rename heif-convert to heif-dec


farindk:
I would like to rename `heif-convert` to `heif-dec` to have a simpler and more consistent naming. Of course, we have to keep a symlink from `heif-convert` for backwards compatibility.

Are there any concerns about this? Things to take care of?

fancycode:
For Ubuntu / Debian packaging I could do this internally through [dh_link](https://manpages.debian.org/testing/debhelper/dh_link.1.html). But I'm fine with you doing this when building.

PR #996, opened 2023-10-17
Rename heif-convert to heif-dec (planned for v1.18.0)

farindk:
Renames the executable `heif-convert` to `heif-dec` and create a symlink from `heif-convert` to `heif-dec` (#985).

Could anyone working on Windows and macOS please verify that this works?
Especially generating the symlink with the correct directories and executable suffix (100d1725269d9af42b6a803d5784eb804b46ea63)?


kmilos:
> Could anyone working on Windows and macOS please verify that this works?

Does not seem to work under MSYS2:
```
CMake Error: failed to create symbolic link 'C:/msys64/opt/libheif/bin/heif-convert.exe': The system
 cannot find the path specified.
```

kmilos:
> Better with the above change?

Better in a sense that it now finds it, but still fails, now with:
```
CMake Error: failed to create symbolic link 'C:/msys64/opt/libheif/bin/heif-convert.exe': A required
 privilege is not held by the client.
```

farindk:
A web search told me that Windows requires administrator privileges to create symlinks. (The reason is that symlinks were only introduced with Windows Vista and since it would open security risks when using older software, creating symlinks has been restricted to administrators.)

Now, what should we do for Windows? Copy the binary instead?

silverbacknet:
Another option some software uses, since static (no plugin) builds can be pretty large, is a stub that simply forwards the entire command environment to the real exe, and returns whatever it does.

With plug-in being the default, that might be overkill for one scenario.

farindk:
@silverbacknet Right, good idea. But in our case, the binary is small (~56 kB). The libheif dll is always separate. It is only the `heif-convert` tool that would have to be copied.

fancycode:
@farindk please make sure to change the manpage to reference the new command name (currently still mentions `heif-convert`).

CMakeLists.txt
```diff
@@ -260,6 +260,7 @@ configure_file(libheif.pc.in ${CMAKE_CURRENT_BINARY_DIR}/libheif.pc @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/libheif.pc
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
 
+
```
kmilos, line 263: Please remove non-change from diff

examples/CMakeLists.txt
```diff
@@ -19,19 +19,22 @@ install(TARGETS heif-info RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
...
         common.h)
-target_link_libraries(heif-convert PRIVATE heif)
-install(TARGETS heif-convert RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
-install(FILES heif-convert.1 DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
+target_link_libraries(heif-dec PRIVATE heif)
+install(TARGETS heif-dec RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(FILES heif-dec.1 DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
+
+# create symbolic link from the old name `heif-convert` to `heif-dec`
+install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/heif-dec${CMAKE_EXECUTABLE_SUFFIX} ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/heif-convert${CMAKE_EXECUTABLE_SUFFIX})")
```
kmilos, line 37:
```suggestion
if(NOT WIN32)
  install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/heif-dec${CMAKE_EXECUTABLE_SUFFIX} ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/heif-convert${CMAKE_EXECUTABLE_SUFFIX})")
else()
  install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/heif-dec${CMAKE_EXECUTABLE_SUFFIX} ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/heif-convert${CMAKE_EXECUTABLE_SUFFIX})")
endif()
```

PR #997, opened 2023-10-17
cmake: Fix OS X compatibility_version based on libtool legacy

### LINKED ISSUE #992: Binary-versioning regression
The change from autotools (libtool in particular) in libheif-1.15.1 to cmake retained the same SONAME of the library (we call it "install_name" on OS X), which seems correct as it appears only ABIs were added rather than removed or changed. But the compatibility_version got decremented (was 17.0.0 now 1.0.0 when I upgrade libheif to 1.17.0). It is not allowed for compatibility_version to be decreased for any specific install_name, or it will cause a runtime failure for any program that was linked against that install_name that had had a higher compatibility_version.

Libtool handles all these library versioning details automatically, via `-version-info`, but cmake does not know anything about this. Assuming I am reading the public APIs correctly, that symbols were added but no symbols were removed or changed, it's correct that install_name stays the same, but compatibility_version should be incremented.

dmacks:
Sets a darwin linker value that must be monotonic, which libtool automatically had been incrementing but cmake automatically sets to 1.0.0. Fixes  #992

### DIFF
--- a/libheif/CMakeLists.txt
+++ b/libheif/CMakeLists.txt
@@ -100,6 +100,15 @@ set_target_properties(heif
         VERSION ${PROJECT_VERSION}
         SOVERSION ${PROJECT_VERSION_MAJOR})
 
+# compatibility_version defaults to 1.0.0 and is never allowed to be
+# decreased for any specific SONAME. Libtool in the libheif-1.15.1
+# release had set it to 17.0.0, so force it higher here.
+if (APPLE)
+    set_target_properties(heif
+            PROPERTIES
+	    LINK_FLAGS "-Wl,-compatibility_version,18.0.0")
+endif ()
+
 target_compile_definitions(heif
         PUBLIC
         LIBHEIF_EXPORTS


libheif/CMakeLists.txt
```diff
@@ -100,6 +100,15 @@ set_target_properties(heif
         VERSION ${PROJECT_VERSION}
         SOVERSION ${PROJECT_VERSION_MAJOR})
 
+# compatibility_version defaults to 1.0.0 and is never allowed to be
+# decreased for any specific SONAME. Libtool in the libheif-1.15.1
+# release had set it to 17.0.0, so force it higher here.
+if (APPLE)
+    set_target_properties(heif
+            PROPERTIES
+	    LINK_FLAGS "-Wl,-compatibility_version,18.0.0")
```
kmilos, line 109: You don't want to set this manually.

farindk:
Thanks, I'll further modify this to be easier to maintain.

PR #1004, opened 2023-10-20
Update README.md

arjendekorte:
Using cmake-presets version 3 requires CMake >= 3.21


farindk:
Thanks. Could we change the text slightly to "`--preset` argument needs CMake >= 3.21" ?

arjendekorte:
Agreed. It took me a while to figure out what the problem was, since the CMake error isn't the most helpful to find what is wrong, therefor it might be useful to mention this specifically.

farindk:
Thanks

PR #1012, opened 2023-10-26
fuzzer: The box_size_without_header must be positive

dukesook:
The fuzzer is throwing the following error:
/home/runner/work/libheif/libheif/libheif/box.cc:669:27: runtime error: implicit conversion from type 'int64_t' (aka 'long') of value -8 (64-bit, signed) to type 'size_t' (aka 'unsigned long') changed the value to 18446744073709551608 (64-bit, unsigned)

The BitstreamRange constructor expects a size_t, but an int64_t is being passed in (box.cc line 669). This is a problem when the value is negative.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -657,7 +657,7 @@ Error Box::read(BitstreamRange& range, std::shared_ptr<Box>* result)
 
   // Box size may not be larger than remaining bytes in parent box.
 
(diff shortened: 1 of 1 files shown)


farindk:
I thought the problem was already fixed with 316291a9d374da768c8b629d2f04ed18015a3608.
Does this still occur? Is there an input to reproduce this?

dukesook:
Yes, you're correct. This issue has already been resolved. I'll close the PR.

PR #1013, opened 2023-10-26
examples: correct configuration check for uncompressed codec in heif-enc

bradh:
The option is `WITH_UNCOMPRESSED_CODEC` but we still had `ENABLE_UNCOMPRESSED_ENCODER`.

The original change was https://github.com/strukturag/libheif/commit/9895fdf37558c2efdbaeeb5e29504be03e88e8f8#diff-30575ea3ee10e63b2610466a76c0391eb3a6a3b1984b80ddc668421fa095929c

I think this part may have been accidentally reverted during the JPEG 2000 merge.


farindk:
Yes, seems to have been accidentally reverted.

PR #1015, opened 2023-10-29
nclx: use default values that match sRGB

kleisauke:
Since commit 062b5284216748925cdbf67f9eb8a12bbda503e1, libheif saves images with the default NCLX profile `6,6,6` (Rec. 601 aka NTSC) which is pretty far from sRGB and leads to incorrect colors in most viewers. Additionally, the `colr` HEIF box wasn't updated to reflect this.

Commit 673fb03227790cb5184c9c9aaa5c40775cb3afbf resolves the latter issue by guaranteeing the update of `target_nclx_profile` following the color conversion, thereby ensuring the propagation of any defaults specified in `ColorConversionPipeline::convert_image()`.

Commit 3bc75720f9eb686e2a674a710961422b9920d2e3 resolves the former issue by ensuring NCLX profiles defaults to `1,13,1`  (Rec. 709 with IEC 61966-2-1 transfer function).

Context: https://github.com/libvips/libvips/pull/3731

Thanks to @DarthSim for discovering this.

### DIFF
--- a/libheif/color-conversion/colorconversion.cc
+++ b/libheif/color-conversion/colorconversion.cc
@@ -446,8 +446,7 @@ std::shared_ptr<HeifPixelImage> ColorConversionPipeline::convert_image(const std
 
     // --- pass the color profiles to the new image
 
-    auto output_nclx = std::make_shared<color_profile_nclx>();
-    *output_nclx = step.output_state.nclx_profile;
+    auto output_nclx = std::make_shared<color_profile_nclx>(step.output_state.nclx_profile);
     out->set_color_profile_nclx(output_nclx);
     out->set_color_profile_icc(in->get_color_profile_icc());
 
@@ -546,7 +545,7 @@ std::shared_ptr<HeifPixelImage> convert_colorspace(const std::shared_ptr<HeifPix
     output_state.nclx_profile = *target_profile;
   }
 
-  // If some output nclx values are unspecified, set the to the same as the input.
+  // If some output nclx values are unspecified, set them to the same as the input.
 
   if (output_state.nclx_profile.get_matrix_coefficients() == heif_matrix_coefficients_unspecified) {
     output_state.nclx_profile.set_matrix_coefficients(input_state.nclx_profile.get_matrix_coefficients());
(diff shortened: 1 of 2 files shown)


novomesk:
I agree that NCLX with unspecified 2/2/2 should not be interpreted as 6/6/6.

I understand that default interpretation should be sRGB but I believe it should be 1/13/6 and not 1/13/1. (BTW I also use 1/13/1 in my software but it must be explicitly set that way).

I think it is OK to leave initial NCLX values to 2/2/6 - matrix coefficients is good to define, but primaries and transfer function could be defined via ICC.

libheif/context.cc
```diff
@@ -2465,6 +2465,7 @@ Error HeifContext::encode_image_as_hevc(const std::shared_ptr<HeifPixelImage>& i
     if (!src_image) {
       return Error(heif_error_Unsupported_feature, heif_suberror_Unsupported_color_conversion);
     }
+    *target_nclx_profile = *src_image->get_color_profile_nclx();
```
farindk, line 2468: I don't think this is the correct way to fix this. We have an NCLX of the input image and a desired output NCLX from the `heif_encoding_options`. If both differ, the input will be transformed to match the output options.
Here, you are overwriting the target NCLX (the one that will be written to the file) with the input NCLX even if it was converted to another one (specified in the encoding options).

kleisauke: IIUC, this only uses the NCLX parameters of the input if some of the output NCLX parameters are unspecified, otherwise it would still preserve the output NCLX parameters as-is.

So, for example, when saving with `options.output_nclx_profile == nullptr` (default) it would both remap these unspecified values:
https://github.com/strukturag/libheif/blob/673fb03227790cb5184c9c9aaa5c40775cb3afbf/libheif/nclx.cc#L315-L316

To the ones specified within the input NCLX profile here:
https://github.com/strukturag/libheif/blob/673fb03227790cb5184c9c9aaa5c40775cb3afbf/libheif/color-conversion/colorconversion.cc#L544-L560

This updated output NCLX profile would be propagated to the result of `convert_colorspace()` here:
https://github.com/strukturag/libheif/blob/673fb03227790cb5184c9c9aaa5c40775cb3afbf/libheif/color-conversion/colorconversion.cc#L447-L450

farindk:
I've implemented the handling of unspecified values in a different way that avoids overwriting the `target_nclx_profile` in de41052649fb53d72f8af24d8af370fd8a531c8f.
Combined with your changes of the default values, I think this should solve the problem.

libheif/context.cc
kleisauke: You're right; this fix was incorrect.

kleisauke:
Thanks! I can confirm that PR #1017 resolves this.

PR #1017, opened 2023-10-30
set NCLX defaults to match sRGB and change handling of unspecified NCLX values

farindk:
This is an alternative implementation of #1015.

### DIFF
--- a/examples/heif_enc.cc
+++ b/examples/heif_enc.cc
@@ -64,9 +64,9 @@ int metadata_compression = 0;
 const char* encoderId = nullptr;
 std::string chroma_downsampling;
 
-uint16_t nclx_matrix_coefficients = 6;
-uint16_t nclx_colour_primaries = 2;
-uint16_t nclx_transfer_characteristic = 2;
+uint16_t nclx_matrix_coefficients = 1;
+uint16_t nclx_colour_primaries = 1;
+uint16_t nclx_transfer_characteristic = 13;
 int nclx_full_range = true;
 
 std::string property_pitm_description;
--- a/examples/heif_enc.cc
+++ b/examples/heif_enc.cc
@@ -64,9 +64,9 @@ int metadata_compression = 0;
 const char* encoderId = nullptr;
 std::string chroma_downsampling;
 
-uint16_t nclx_matrix_coefficients = 1;
 uint16_t nclx_colour_primaries = 1;
 uint16_t nclx_transfer_characteristic = 13;
+uint16_t nclx_matrix_coefficients = 5;
 int nclx_full_range = true;
 
 std::string property_pitm_description;
--- a/libheif/nclx.cc
+++ b/libheif/nclx.cc
@@ -339,6 +339,22 @@ void color_profile_nclx::set_from_heif_color_profile_nclx(const struct heif_colo
 }
 
 
+void color_profile_nclx::replace_undefined_values_with_defaults()
+{
+  if (m_matrix_coefficients == heif_matrix_coefficients_unspecified) {
+    m_matrix_coefficients = heif_matrix_coefficients_ITU_R_BT_709_5;
+  }
+
+  if (m_colour_primaries == heif_color_primaries_unspecified) {
+    m_colour_primaries = heif_color_primaries_ITU_R_BT_709_5;
+  }
+
+  if (m_transfer_characteristics == heif_color_primaries_unspecified) {
+    m_transfer_characteristics = heif_transfer_characteristic_IEC_61966_2_1;
+  }
+}
+
+
 Error Box_colr::parse(BitstreamRange& range)
 {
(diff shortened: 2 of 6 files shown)


examples/heif_enc.cc
```diff
@@ -64,9 +64,9 @@ int metadata_compression = 0;
 const char* encoderId = nullptr;
 std::string chroma_downsampling;
 
-uint16_t nclx_matrix_coefficients = 6;
-uint16_t nclx_colour_primaries = 2;
-uint16_t nclx_transfer_characteristic = 2;
+uint16_t nclx_matrix_coefficients = 1;
```
kmilos, line 67: Should be 5 for sRGB

libheif/nclx.cc
```diff
@@ -339,6 +339,22 @@ void color_profile_nclx::set_from_heif_color_profile_nclx(const struct heif_colo
...
+{
+  if (m_matrix_coefficients == heif_matrix_coefficients_unspecified) {
+    m_matrix_coefficients = heif_matrix_coefficients_ITU_R_BT_709_5;
+  }
+
+  if (m_colour_primaries == heif_color_primaries_unspecified) {
+    m_colour_primaries = heif_color_primaries_ITU_R_BT_709_5;
+  }
+
+  if (m_transfer_characteristics == heif_color_primaries_unspecified) {
```
kleisauke, line 352:
```suggestion
  if (m_transfer_characteristics == heif_transfer_characteristic_unspecified) {
```

libheif/nclx.cc
```diff
@@ -312,9 +312,9 @@ void color_profile_nclx::free_nclx_color_profile(struct heif_color_profile_nclx*
 
 void color_profile_nclx::set_default()
 {
-  m_colour_primaries = 2;
-  m_transfer_characteristics = 2;
-  m_matrix_coefficients = 6;
+  m_colour_primaries = 1;
+  m_transfer_characteristics = 13;
+  m_matrix_coefficients = 1;
```
kmilos: 5 for sRGB

libheif/nclx.cc
```diff
@@ -339,6 +339,22 @@ void color_profile_nclx::set_from_heif_color_profile_nclx(const struct heif_colo
 }
 
 
+void color_profile_nclx::replace_undefined_values_with_defaults()
+{
+  if (m_matrix_coefficients == heif_matrix_coefficients_unspecified) {
+    m_matrix_coefficients = heif_matrix_coefficients_ITU_R_BT_709_5;
```
kmilos, line 345: `heif_matrix_coefficients_ITU_R_BT_470_6_System_B_G` (i.e. 5) for sRGB

farindk:
Are there any references what the values should be for sRGB? I found this random piece of code: https://observablehq.com/@eeeps/color-space-determiner that also seems to favor 1,13,1.

kmilos:
> I found this random piece of code: https://observablehq.com/@eeeps/color-space-determiner that also seems to favor 1,13,1.

Plain wrong. See https://www.itu.int/rec/T-REC-H.273

farindk:
Accoding to H273, color_primaries=1. I think we agree here.
It also says to use transfer_characteristics=13.
Then, H273 says to use matrix_coefficients=0 of SRGB. Well ok, but that's not what we're looking for.
How do you derive that matrix_coefficients should be 5 or 1 ?

kmilos:
> Then, H273 says to use matrix_coefficients=0 of SRGB. Well ok, but that's not what we're looking for.
> How do you derive that matrix_coefficients should be 5 or 1 ?

IEC 61966-2-1 mentions [sYCC ](https://en.wikipedia.org/wiki/SRGB#sYCC) representation that specifically uses matrix coeffs 5 (601).

However, note that the same wikipedia page also mentions sRGB conversion to XYZ using the 709 coeffs earlier...

Also note that the JPEG/JFIF specification uses 601 matrix coeffs.

farindk:
Ok, that makes sense. Then we'll go for `matrix_coefficients=5`.

farindk:
> For reference, libavif seems to default to `13/1/6`: https://github.com/AOMediaCodec/libavif/blob/f9dd31fea43f748ce1a6e49be4cd9e68b916ebe6/apps/avifenc.c#L2269-L2276 https://github.com/AOMediaCodec/libavif/blob/f9dd31fea43f748ce1a6e49be4cd9e68b916ebe6/apps/avifenc.c#L1460

Yes. But since matrix_coefficients 5 and 6 are exactly the same, this does not matter.

kmilos:
> For reference, libavif seems to default to `13/1/6`

Yeah, I think this was because either Chrome or Safari only implemented that combo and not (IMHO more canonical) `13/1/5` in addition...

farindk:
Thanks, @kleisauke , @kmilos

PR #922, opened 2023-07-11
migrate box implementation out of box.h/box.cc

bradh:
The `box.h` and `box.cc` files are pretty large, and box.h is likely to get larger as documentation is added.

This PR is a commit stack migrating various boxes out to codec (or colour) specific implementation.

Diffstat:
```
 CMakeLists.txt |    4 +
 avif.cc        |  190 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 avif.h         |   98 +++++++++++++++++++++++++++++++++
 box.cc         |  853 +-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 box.h          |  363 -----------------------------------------------------------------------------------------------------------------------------
 context.cc     |    1 
 file.cc        |    2 
 file.h         |    3 +
 hevc.cc        |  288 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 hevc.h         |   67 +++++++++++++++++++++++
 jpeg.cc        |   53 ++++++++++++++++++
 jpeg.h         |   52 +++++++++++++++++
 nclx.cc        |  248 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 nclx.h         |  113 ++++++++++++++++++++++++++++++++++++++
 pixelimage.h   |    2 
 vvc.cc         |  128 ++++++++++++++++++++++++++++++++++++++++++++
 vvc.h          |   98 +++++++++++++++++++++++++++++++++
 17 files changed, 1349 insertions(+), 1214 deletions(-)
```

I had hoped for more, but it is a start.

PR #1019, opened 2023-10-30
Sanitise jpgc box size before memory allocation

lovell:
~~I think the work for https://github.com/strukturag/libheif/pull/922 may have removed some existing security checks around memory allocation.~~

EDIT: The `jpgC` logic was introduced in commit https://github.com/strukturag/libheif/commit/3e82d5199aec196ba59369d5a7834072d0a51bd1 (and #922 moved it).

This PR adds sanity checking for `jpgC` box sizes and a 32 byte test file discovered via fuzz testing that will attempt to allocate ~9 exabytes and therefore segfault when parsed using v1.17.x. I've not checked other box types, but it's possible this sanity check may need to be applied in a more general manner to other `parse` functions.


lovell:
It looks like this was fixed via commit https://github.com/strukturag/libheif/commit/8ff82278f98b03a0764fefa31393e4cab29dc675

farindk:
> It looks like this was fixed via commit [8ff8227](https://github.com/strukturag/libheif/commit/8ff82278f98b03a0764fefa31393e4cab29dc675)

Yes. Sorry for the double work. I also saw this is a fuzzer output.

farindk:
> it's possible this sanity check may need to be applied in a more general manner to other parse functions.

Right, this all started because of the changed handling of boxes with zero size field (undefined size). That should probably be handled in a more general way. I've opened #1021 for this.

Issue #1002, opened 2023-10-19
Add API to add custom `infe` items


farindk:
Some applications might want to safe non-standardized data into the HEIF file. To enable this use case, there should be an API function to write (and read) custom data items.

a.    Adding an optional item_name
b.    Specifying the construction method (store in mdat or idat)
c.    An option to associate the item with 0, 1, or multiple other infe items. 
d.    For items of type ‘uri ‘, specify the item_uri_type

dukesook:
Similar to #954

farindk:
API is available now in `heif_items.h` (look in `develop-v1.18.0` branch).

> b. Specifying the construction method (store in mdat or idat)

This is still missing. Why do we need this?

bradh:
> > b. Specifying the construction method (store in mdat or idat)
> 
> This is still missing. Why do we need this?

One case is to trade off which items are needed if you are putting in items that are useful for search and discovery or preview. So instead of having to read the entire file, you can just read the first part (maybe 1M, maybe 128K - per MIAF progressive mode) and have what you need. That can save you a lot of transfer / latency in a cloud case (e.g. pulling from an S3 bucket or similar object store directly into the browser).

farindk:
> > > b. Specifying the construction method (store in mdat or idat)
> > 
> > 
> > This is still missing. Why do we need this?
> 
> One case is to trade off which items are needed if you are putting in items that are useful for search and discovery or preview.

Ok, good point, but isn't that more an issue of how to sort the data in the file? The construction method does not determine whether the data will be near the start or at the end of the file. Usually an `mdat` is at the end of the file, but the data can be ordered in any way inside it. There can be several `mdat`s or even data behind it.

Does that mean that instead of a construction we would like to have some kind of `priority` that says how early in the file the data should appear?

bradh:
I think we always want to build the data with `ftyp` -> `meta` -> `mdat` (and since its just an offset into the file, multiple `mdat` can always be collapsed). That will never be worse than putting `mdat` first, since you can't make sense of the `mdat` content until you have `meta`.

Priority could be a useful, and I think that is worth more thought prior to implementation.

As an alternative, we can probably come up with some "best guess" heuristics in the absence of priority, but maybe `idat` vs `mdat` is enough.

[...]

PR #1027, opened 2023-11-04
wip: item API suggestions

bradh:
The unit test currently fails - see https://github.com/strukturag/libheif/issues/1002#issuecomment-1793265864

### DIFF
--- a/libheif/heif_items.cc
+++ b/libheif/heif_items.cc
@@ -37,10 +37,15 @@ int heif_context_get_number_of_items(const struct heif_context* ctx)
   return (int) ctx->context->get_heif_file()->get_number_of_items();
 }
 
+
 int heif_context_get_list_of_item_IDs(const struct heif_context* ctx,
                                       heif_item_id* ID_array,
                                       int count)
 {
+  if (!ID_array) {
+    return 0;
+  }
+
   auto ids = ctx->context->get_heif_file()->get_item_IDs();
   for (int i = 0; i < (int) ids.size(); i++) {
     if (i == count) {
@@ -52,6 +57,7 @@ int heif_context_get_list_of_item_IDs(const struct heif_context* ctx,
 
   return (int) ids.size();
 }
+
 
 uint32_t heif_context_get_item_type(const struct heif_context* ctx, heif_item_id item_id)
 {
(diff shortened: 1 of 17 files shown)


libheif/heif_items.cc
```diff
@@ -32,30 +32,30 @@
 
 // ------------------------- reading -------------------------
 
-int heif_context_get_number_of_items(const struct heif_context* ctx)
+size_t heif_context_get_number_of_items(const struct heif_context* ctx)
```
farindk, line 35: I'd like to avoid `size_t` as much as possible in the interface because that one bit of extra range is not needed here and `size_t` makes detecting errors much more complicated:
[See Scott Meyer](http://www.aristeia.com/Papers/C++ReportColumns/sep95.pdf)

I'm probably fine with `size_t` when specifying the size of memory areas that may be so large to hit the 2GB boundary on 32bit systems.

bradh: I'm not sure I agree with the analysis. Even in the quoted article, there are plenty of other cases where signed int won't save you (e.g. the off-by-one case).

Internally we're just casting this to `size_t` anyway, and a common case is to pass in the `std::vector` size, so I'm casting to `int` to match the API, then its being cast back...

Possibly we can use `size_t` and then just check the number isn't greater than some limit, similar to the way we do security checks.

bradh: To get this moving, I've removed the `size_t` changes.

The unit test still fails.

libheif/heif_items.cc
```diff
@@ -216,6 +216,7 @@ struct heif_error heif_context_add_item(struct heif_context* ctx,
 
   if (result && out_item_id) {
     *out_item_id = result.value;
+    ctx->context->get_heif_file()->set_primary_item_id(*out_item_id);
```
farindk, line 219: But that will overwrite any other primary item id that was set before whenever a new item is added, right?

farindk: And only if `out_item_id` is not NULL, which is quite unexpected.

farindk:
The core issue is probably that libheif was designed to write standard image files and not to build arbitrary files (like your new test) from scratch. This probably has to be refactored from the ground up to get a clean solution.
The new add-item interfaces are ok for adding additional boxes to image files, but one cannot expect yet to build files from scratch.

bradh:
> The core issue is probably that libheif was designed to write standard image files and not to build arbitrary files (like your new test) from scratch. This probably has to be refactored from the ground up to get a clean solution. The new add-item interfaces are ok for adding additional boxes to image files, but one cannot expect yet to build files from scratch.

I think that is a valid design choice.  Will close this.

Issue #1003, opened 2023-10-19
Add `taic`, `itai` timestamp boxes


farindk:
This is work in progress, but ultimatively add ‘taic’ & ‘itai’ boxes which provide support for nano-precision timestamps.

farindk:
`taic` uses a `float32` value. Where is it defined how that value is stored in the file (little-endian / big-endian)?
Have you considered using an integer value for this in the standard?

farindk:
Another point: I'd like to remove `status_bits` from  `heif_property_set_tai_timestamp()`. If I read the spec correctly, `timestamp_validity` can be derived from `TAI_timestamp` (see my comments to the specs).
`synchronization_status` should be an extra parameter. It is the responsibility of the library to put this into the correct bits. That's exactly what should be transparent to the user.

Am I correct with the first assumption? Can we derive `timestamp_validity` from `TAI_timestamp==0xFFFF....FFFF` ?

bradh:
> `taic` uses a `float32` value. Where is it defined how that value is stored in the file (little-endian / big-endian)? 

It should (currently) be `double(32)` which is a big-endian value (see ISO 14496-1:2010 Section 8.2.2 - "most significant byte first"). 

Apparently there is a new part (-34) that will clarify the syntax, by codifying all the inconsistencies as normative. That will make double an obsolete alias and introduce `float(32)`. See https://www.mpeg.org/structure/systems/ (https://www.mpeg.org/wp-content/uploads/mpeg_meetings/143_Geneva/w22951.zip for the draft)

> Have you considered using an integer value for this in the standard?

Seems reasonable to me.

PR #1028, opened 2023-11-04
Heif Properties: Timestamp boxes

dukesook:
Feedback is always appreciated.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -3087,3 +3087,74 @@ Error Box_udes::write(StreamWriter& writer) const
   prepend_header(writer, box_start);
   return Error::Ok;
 }
+
+
+
+std::string Box_taic::dump(Indent& indent) const {
+  std::ostringstream sstr;
+  sstr << Box::dump(indent);
+  sstr << indent << "time_uncertainty: " << m_time_uncertainty << "\n";
+  sstr << indent << "correction_offset: " << m_correction_offset << "\n";
+  sstr << indent << "clock_drift_rate: " << m_clock_drift_rate << "\n";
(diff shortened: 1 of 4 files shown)


farindk:
Thanks for the PR.
One comment:  since you are merging the master branch into your PR, it makes the PR very hard to read and see what actually changed. Usually, you should keep your PR branch clean with only your changes. That makes it much easier to read, approve, and merge into the master branch without danger to break things there.

Would it be possible that you untangle this PR into a new, clean PR? You could use cherry-picking of your commits to do this.

dukesook:
Okay, I removed the master branch from the PR so my changes are more clear. 

Referring to the API in heif_properties.h: 
Each timestamp parameter shall be interpreted as "unknown" when set to a certain value. I'm trying to hide this detail from the user with a simple nullptr to indicate that the value is "unknown". However, I'm worried that all the pointers overcomplicate the API functions. 

I have an MP4Box.js branch that can read the itai & taic boxes: https://github.com/gpac/mp4box.js/pull/366

dukesook:
I had to git-cherry pick [fix reading images with unspecified jpgC box](https://github.com/strukturag/libheif/commit/8ff82278f98b03a0764fefa31393e4cab29dc675) for the fuzzer to work.

farindk:
At least for `heif_property_get_clock_info`, it may be better to introduce a struct like `heif_tai_clock_info` that collects all fields. This could even be versioned if we want to be able to support future extensions. This eliminates the long list of parameters and pointers.
Then, we could use constants like `heif_tai_clock_correction_offset = 0x7fff...` to make it easier for the user.

dukesook:
resolves #1003

PR #1033, opened 2023-11-12
avc: implement avcC box parsing

bradh:
There doesn't appear to be much use of AVC - the only test file I'm aware of is https://github.com/nokiatech/heif_conformance/blob/master/conformance_files/multilayer004.heic

However I'd eventually like to support it, and this is an initial step.


bradh:
It looks like this is failing fuzzer checks. Is there any way I can get the repro for it?

farindk:
Thanks

PR #1036, opened 2023-11-14
kvaazar: set chroma in kvazaar_query_input_colorspace2

DarthSim:
Hey there 👋

Currently, `kvazaar_query_input_colorspace2` preserves input `chroma`, and encoding of an RGBA input fails because of this. This PR makes `kvazaar_query_input_colorspace2` rewrite `chroma` if it is incompatible with `heif_colorspace_YCbCr`.


farindk:
Correct. Thank you !

PR #1037, opened 2023-11-14
kvazaar: ensure input pucture dimensions are divisible by 8

DarthSim:
Hey there 👋

Kvazaar requires dimensions of the input `kvz_picture` to be divisible by 8. Otherwise, encoding fails and aborts the whole program. Unfortunately, Kvazaar's API doesn't do this for us.

### DIFF
--- a/libheif/plugins/encoder_kvazaar.cc
+++ b/libheif/plugins/encoder_kvazaar.cc
@@ -510,7 +510,17 @@ static struct heif_error kvazaar_encode_image(void* encoder_raw, const struct he
   }
 */
 
-  kvz_picture* pic = api->picture_alloc_csp(kvzChroma, config->width, config->height);
+  uint32_t pic_width = config->width;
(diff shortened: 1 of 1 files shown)


farindk:
The better way to do that is probably to implement `query_encoded_size` in the encoder plugin (version 3). I can have a look at that.

Or we may need both. A quick test gave me this result
````
heif-enc: image.c:59: kvz_image_alloc: Assertion `(width % 2) == 0' failed.
````
So I assume `query_encoded_size` should round up to multiples of 2 and then maybe we have to allocate images with multiples of 8 to match some memory alignment issues?

farindk:
I have implemented this in a different way (see above). Please check whether this is also ok on your data.

DarthSim:
It works, thanks! 👍

farindk:
Thanks for testing and your initial patch.

Issue #62, opened 2018-08-15
Saving in RGB colorspace (was: "heif-enc lossless isn't lossless")


kliberty:
I used `heif-enc` to convert a png "losslessly" then converted it back to png with `heif-convert` and the results where not lossless. I attached the original png bellow as well as the difference between them (levels increased dramatically).
```
heif-enc -L tree.png
heif-convert tree.heic tree.png
```
![ash_tree_-_geograph org uk_-_590710](https://user-images.githubusercontent.com/8992170/44141670-d46cc212-a04b-11e8-9f3e-c5b18d43594b.png)
![png-heic-png](https://user-images.githubusercontent.com/8992170/44141671-d47c8fd0-a04b-11e8-9a83-af8e1bc9cc2e.png)

farindk:
"Lossless" refers to the HEVC compression settings. When you are encoding from a PNG, there is a color-space conversion involved. PNGs are stored in RGB color-space, while HEIF is in YUV. Hence, you get some loss because of the color-space conversion.

TODO: 'heif-enc' should print a warning about this.

farindk:
Yes, this is definitely the way to go. However, I am still unsure how to correctly indicate the colour space of the image data. Strangely, I didn't see anything about this in the HEIF and HEVC standards, and I also did not come across any RGB coded HEIF images so far. Maybe the coding of RGB images was overlooked?

farindk:
A warning is now printed when an RGB input image is compressed to YUV in lossless mode (6bd1a7e).

dbtsai:
Out of my curiosity, does iOS/OSX render `colr` box? 

I know JPEG is using YUV internally, but in `ImageMagick`, JPEG will aways render as RGB. Do you know why it works like this?

On the plus side, it gives the right mae metric when I use `compare` to compare the original png file.

kliberty:
@dbtsai Sorry, but what do you mean "will always render as RGB" and why what is? JPEG actually has the ability to store RGB data for high quality images (JPEGs out of a DSLR for example).

farindk:
[...]

PR #1039, opened 2023-11-18
Default to `chroma=444`/`matrix_coefficients=0` in lossless mode

kleisauke:
Resolves: #62.

### DIFF
--- a/examples/heif_enc.cc
+++ b/examples/heif_enc.cc
@@ -168,9 +168,9 @@ void show_help(const char* argv0)
             << "  -e, --encoder ID            select encoder to use (the IDs can be listed with --list-encoders)\n"
             << "      --plugin-directory DIR  load all codec plugins in the directory\n"
             << "  -E, --even-size   [deprecated] crop images to even width and height (odd sizes are not decoded correctly by some software)\n"
-            << "  --matrix_coefficients     nclx profile: color conversion matrix coefficients, default=6 (see h.273)\n"
-            << "  --colour_primaries        nclx profile: color primaries (see h.273)\n"
-            << "  --transfer_characteristic nclx profile: transfer characteristics (see h.273)\n"
+            << "  --colour_primaries        nclx profile: color primaries, default: 1 (see h.273)\n"
+            << "  --transfer_characteristic nclx profile: transfer characteristics, default: 13 (see h.273)\n"
+            << "  --matrix_coefficients     nclx profile: color conversion matrix coefficients, default: 6 for lossy or 0 for lossless (see h.273)\n"
             << "  --full_range_flag         nclx profile: full range flag, default: 1\n"
             << "  --enable-two-colr-boxes   will write both an ICC and an nclx color profile if both are present\n"
             << "  --premultiplied-alpha     input image has premultiplied alpha\n"
@@ -178,13 +178,7 @@ void show_help(const char* argv0)
             << "  -C,--chroma-downsampling ALGO   force chroma downsampling algorithm (nn = nearest-neighbor / average / sharp-yuv)\n"
             << "                                  (sharp-yuv makes edges look sharper when using YUV420 with bilinear chroma upsampling)\n"
             << "  --benchmark               measure encoding time, PSNR, and output file size\n"
-            << "  --pitm-description TEXT   (EXPERIMENTAL) set user description for primary image\n"
-
-            << "\n"
-            << "Note: to get lossless encoding, you need this set of options:\n"
-            << "  -L                       switch encoder to lossless mode\n"
-            << "  -p chroma=444            switch off chroma subsampling\n"
-            << "  --matrix_coefficients=0  encode in RGB color-space\n";
+            << "  --pitm-description TEXT   (EXPERIMENTAL) set user description for primary image\n";
 }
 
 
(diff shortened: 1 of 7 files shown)


bigcat88:
And if encoding goes in YCbCr mode, why should be matrix_coefficients set to 0 automatically?
The same for L(greyscale) and LA(greyscale with alpha) modes.

kleisauke:
> And if encoding goes in YCbCr mode, why should be matrix_coefficients set to 0 automatically?
> The same for L(greyscale) and LA(greyscale with alpha) modes.

I'm not entirely familiar with these pixel formats. The latest revision of this PR makes `matrix_coefficients=0` default in lossless mode instead of forcing it, allowing users to still use other matrix coefficients.

For reference, `avifenc` fails when chroma is not 4:4:4 or the matrix coefficients is not identity (`x/y/0`) in lossless mode.
https://github.com/AOMediaCodec/libavif/blob/484984aa0365b71d211a7f6612f30787a1c88230/apps/avifenc.c#L1949-L1956
https://github.com/AOMediaCodec/libavif/blob/484984aa0365b71d211a7f6612f30787a1c88230/apps/avifenc.c#L1971-L1986
https://github.com/AOMediaCodec/libavif/wiki/CICP#common-cicp-triples

bigcat88:
`matrix_coefficients=0` should not be default for lossless mode, as not all colour spaces require it, it is only for RGB(A).
With the subsampling chroma is the same situation, if image was already in 4:2:0 or 4:2:2 for example, you do not want to set chroma=444 when encoding it in lossless mode.

farindk:
We have to differentiate two things here. From the codec side of view (h265 or av1), the codec can be switched to lossless no matter what input color space. I.e. we could have YCbCr 4:2:0 and encode that lossless from the view of the codec. That's what is switched on with the lossless flag(s) in `heif-enc`. However, that is not necessarily lossless from the user's point of view because there can be color conversions and chroma changes. Hence the need for matrix_coefficients=0 and chroma-4:4:4.

This PR is mixing the two concepts. I would like to avoid this. However, I'm fine to redefine the parameters such that "-p lossless=true" is the codec setting, but "--lossless" is not a mere shortcut for that, but also sets chroma and matrix as noted at the bottom of the help:
````
Note: to get lossless encoding, you need this set of options:
  -L                       switch encoder to lossless mode
  -p chroma=444            switch off chroma subsampling
  --matrix_coefficients=0  encode in RGB color-space
````

However, that could be handled in the `heif-enc` without introducing changes to the library behaviour.

**Comment by farindk** [comment]
I've implemented this differently in `heif-enc` in this commit: ee6990776ddbf508e500ae1f9b6c66ec658b4cde
Do you think that is ok?
I'm still a bit worried as this forces also YCbCr input to be encoded in RGB. Maybe we would have to check the input color space and set the matrix coefficients accordingly.

**Comment by farindk** [comment]
I've improved this further in 94c3b21. When the input is YCbCr, `-L` automatically sets the output NCLX to the input and chooses the matching chroma format such that the encoding is also lossless for all YCbCr formats.

**Comment by bigcat88** [comment]
> I've improved this further in [94c3b21](https://github.com/strukturag/libheif/commit/94c3b21d8df09b69c2c1752752e1810fd902d7d7). When the input is YCbCr, `-L` automatically sets the output NCLX to the input and chooses the matching chroma format such that the encoding is also lossless for all YCbCr formats.

wow, that's really cool

**Comment by kleisauke** [comment]
Thanks for the explanation! Indeed, this logic should probably not be included at library level. I'll abandon this PR in favor of the commits mentioned above. :thumbsup:

libvips always encodes in `heif_colorspace_RGB`, so I'll investigate if we could set `matrix_coefficients=0` by default in lossless mode. This would help to resolve this `FIXME`:
https://github.com/libvips/libvips/blob/c37b9750fdd44c97461f32da7a4b17b628f6f0e8/test/test-suite/test_foreign.py#L1416-L1418

Note that `heif-enc` has a similiar issue with commit 94c3b21d8df09b69c2c1752752e1810fd902d7d7 when using `-L`, it looks like `nclx_matrix_coefficients = 0` has no effect.
https://github.com/strukturag/libheif/blob/94c3b21d8df09b69c2c1752752e1810fd902d7d7/examples/heif_enc.cc#L797

Since it's conditionally excluded here:
https://github.com/strukturag/libheif/blob/94c3b21d8df09b69c2c1752752e1810fd902d7d7/examples/heif_enc.cc#L845-L846

Reproducer:
```console
# heif-enc
$ curl -LO https://wsrv.nl/transparency_demo.png
$ heif-enc -L transparency_demo.png -o x.avif
$ heif-convert x.avif output.png
File contains 1 image
Written to output.png
$ vips subtract transparency_demo.png output.png x.v
$ vips abs x.v x2.v
$ vips max x2.v
1.000000

# vips
# subsample_mode=off is not needed after PR https://github.com/libvips/libvips/pull/3760
$ vips copy transparency_demo.png x.avif[lossless,subsample_mode=off]
$ vips copy x.avif output.png
$ vips subtract transparency_demo.png output.png x.v
$ vips abs x.v x2.v
$ vips max x2.v
1.000000
```

farindk:
@kleisauke Yes, the bug for RGB input is a regression from the second commit. Thanks for reporting this. Should be fixed now.

PR #1040, opened 2023-11-21
Fix installing Gnome files if examples are build.

fancycode:
Regression was introduced in 060b55c93ef2338998bf8bcc542421d9f5fd99fd.

Noticed this while working on the packaging of 1.17.4.

### DIFF
--- a/gnome/CMakeLists.txt
+++ b/gnome/CMakeLists.txt
@@ -1,3 +1,3 @@
-if(WITH_EXAMPLES AND TARGET PNG::PNG)
+if(TARGET heif-thumbnailer)
   install(FILES heif.thumbnailer DESTINATION ${CMAKE_INSTALL_DATADIR}/thumbnailers)
 endif()


fancycode:
cc @kmilos

gnome/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+find_package(PNG)
 if(WITH_EXAMPLES AND TARGET PNG::PNG)
```
kmilos, line 2: Hm, I thought the previous `find_package(PNG)` in the examples directory would be sufficient...

Maybe it's best (i.e. clearer) if we change it to just `if(TARGET heif-thumbnailer)` instead? Is that one define in this scope?

fancycode: I don't know much about cmake, but would assume only stuff defined in parent folders is available in subfolders. The `examples` is a sibling of `gnome` so things defined there are probably not available here?

kmilos: For variables, I think that is the case, but I was under the impression the targets were known globally wherever they were defined...

kmilos: Would you mind trying it out anyway?

farindk: How about `if (TARGET heif-thumbnailer)` to test directly if the binary exists than testing it indirectly?

fancycode: This works :+1: 
Interesting though that the `TARGET PNG::PNG` does not work - or are external targets handled differently?

fancycode: No problem, thanks for your various contributions!

fancycode:
@farindk feel free to merge when CI is green

PR #1047, opened 2023-11-27
Fix build warnings and errors in Visual Studio 2022

fishmonger42:
Made the implementation of getopt_internal match the declaration
Fix warning about the unsafe use of int as bool
Remove unnecessary cast (SvtAv1FixedBuf) causing compiler error

### DIFF
--- a/libheif/plugins/encoder_svt.cc
+++ b/libheif/plugins/encoder_svt.cc
@@ -674,7 +674,7 @@ struct heif_error svt_encode_image(void* encoder_raw, const struct heif_image* i
   svt_config.logical_processors = encoder->threads;
 
   // disable 2-pass
-  svt_config.rc_stats_buffer = (SvtAv1FixedBuf) {nullptr, 0};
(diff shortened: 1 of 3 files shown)


libheif/plugins/encoder_svt.cc
```diff
@@ -674,7 +674,7 @@ struct heif_error svt_encode_image(void* encoder_raw, const struct heif_image* i
   svt_config.logical_processors = encoder->threads;
 
   // disable 2-pass
-  svt_config.rc_stats_buffer = (SvtAv1FixedBuf) {nullptr, 0};
+  svt_config.rc_stats_buffer = {nullptr, 0};
```
bradh, line 677: @fishmonger42 This part already got applied (see https://github.com/strukturag/libheif/pull/1059) and is probably the cause of the merge conflicts. Possibly you just need to drop it.

farindk:
Thanks. I've fixed the merge conflict.

PR #1053, opened 2023-12-01
idat: protect against trying to output negative size

bradh:
This one showed up in box fuzzing.

I wasn't sure about creating a new unit test - can drop that part if you would prefer.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -2854,7 +2854,11 @@ std::string Box_idat::dump(Indent& indent) const
   std::ostringstream sstr;
   sstr << Box::dump(indent);
 
-  sstr << indent << "number of data bytes: " << get_box_size() - get_header_size() << "\n";
+  if (get_box_size() >= get_header_size()) {
+    sstr << indent << "number of data bytes: " << get_box_size() - get_header_size() << "\n";
+  } else {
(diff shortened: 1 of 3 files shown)


libheif/box.cc
```diff
@@ -2854,7 +2854,9 @@ std::string Box_idat::dump(Indent& indent) const
   std::ostringstream sstr;
   sstr << Box::dump(indent);
 
-  sstr << indent << "number of data bytes: " << get_box_size() - get_header_size() << "\n";
+  if (get_box_size() > get_header_size()) {
+    sstr << indent << "number of data bytes: " << get_box_size() - get_header_size() << "\n";
+  }
```
farindk, line 2859: Could we add something like
````
else {
  sstr << indent << "number of data bytes is invalid\n";
}
````

farindk: Maybe `>=` ? Size 0 might be a valid edge case.

bradh: > Maybe `>=` ? Size 0 might be a valid edge case.

I checked 14496-12:2022 Section 8.11.11 and didn't see any prohibition, so assume that 0 is valid. Have updated the code.

bradh: Have updated the PR to include these.

farindk:
Thanks

PR #1056, opened 2023-12-02
examples: update heif-info manual page

bradh:
This updates the documentation to reflect the new version option. It also fixes a duplication in the descriptive part for help (was `-help` and `--help`, instead of `-h` and `--help`).


farindk:
Thanks

PR #1059, opened 2023-12-08
svt: fix c99 extension warning

### LINKED ISSUE #956: Usage of non-standard explicit type conversion syntax
https://github.com/strukturag/libheif/blob/d6e233595ca283eecd649d74f7b35762957ec4a6/libheif/plugins/encoder_svt.cc#L675

error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax

This seems to be throwing an error with Visual C++ 2022.

I hit the error trying to enable SVT-AV1 in https://github.com/conda-forge/libheif-feedstock/pull/9

bradh:
Resolves #956

See #956 for the MSVC and clang warning (that becomes an error)

### DIFF
--- a/libheif/plugins/encoder_svt.cc
+++ b/libheif/plugins/encoder_svt.cc
@@ -674,7 +674,7 @@ struct heif_error svt_encode_image(void* encoder_raw, const struct heif_image* i
   svt_config.logical_processors = encoder->threads;
 
   // disable 2-pass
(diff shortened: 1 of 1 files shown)


farindk:
thanks. Will cherry-pick this to master. All fixes like this can go into `master`. In `develop` we only have to put the API changes.

PR #1062, opened 2023-12-10
remove obsolete pkg_config variables

### LINKED ISSUE #758: libheif.pc shows builtin encoders/decoders even if built as plugins
The `libheif.pc` has fields like `builtin_h265_decoder=yes` even if the libde265 plugin was built as plugin, i.e. is not builtin.

Are these `builtin_*` fields really used given that there are more possible decoders / encoders than what is currently checked?

bradh:
Resolves #758


farindk:
Seems everyone is fine with this.

PR #1071, opened 2023-12-21
openjpeg: add support for setting lossless mode

bradh:
Lossless (reversible) is the default in OpenJPEG, and previously we silently ignored setting that on or off.

This moves the compression parameters into the encoder context, and allows setting it.

In my tests, reversible was actually smaller (and matched the pre-PR size). Possibly that is just a function of the test image input though.


farindk:
Thanks. I've also split this into two commits.

Issue #420, opened 2021-01-07
Where can user find jpeg2000 compatible with heif container?


Jamaika1:
https://jpeg.org/jpeg2000/index.html
_Part 16 specifies the carriage of JPEG 2000 codestreams in ISO/IEC 23008-12, commonly referred to as HEIF. A revision is underway to support more flexible wrapping of all JPEG 2000 codestreams, including HTJ2K._

bradh:
libheif can now write this. So you can create whatever you need.

Jamaika1:
Thanks for the info. It's been while since I created libheif.
As far as I know openjpeg doesn't create htj2k. Is libheif compatible with htj2k? I don't know. Give examples of commands.

bradh:
OpenJPEG supports HT (part 15) for decoding as of 2.5.0. There is a discussion topic about the proposed part 16 update in #1054.

Also, your response comes across as demanding, and will point out I was only trying to help with the original question about jpeg2000 source data.

bradh:
You can also test with https://github.com/bradh/libheif/tree/openjph which implements the Part 16 encoding using OpenJPH.

Jamaika1:
Super. I just created a libheif encoder with openjph in C++11. heif(j2c) is created, but I am missing the function j2c/j2k -> heif(j2c/j2k) -> j2c/j2k.

PR #1072, opened 2023-12-21
openjph: HT-J2K encoder

bradh:
This is an early-exposure draft of an OpenJPH based High Throughput JPEG-2000 encoder.

Thanks to @aous72 for OpenJPH and for debugging assistance. Also thanks to @Jamaika1 for suggesting this at #420.

An OpenJPH-based decoder is practical, but is not yet written. In the meantime, the OpenJPEG decoder will work.

I'm not sure about the additional codec type (`heif_compression_HTJ2K`). It seems like the JPEG community think that HT-J2K is really (a variation of) JPEG 2000, but if we want to pick a working encoder for the user, it seems like it would be better to have a different value.

All input appreciated. 

Note: This will likely have merge conflicts with #1071, which I'll resolve after that gets merged.


libheif/context.cc
```diff
@@ -2304,6 +2304,15 @@ Error HeifContext::encode_image(const std::shared_ptr<HeifPixelImage>& pixel_ima
     }
       break;
 
+    case heif_compression_HTJ2K: {
```
farindk, line 2307: Minor thing: maybe this `case` can be simply combined with the `heif_compression_JPEG2000` above.

farindk:
Thank you, this is a great addition.
I also think that adding a new compression format enum `heif_compression_HTJ2K` makes sense, since (at least for the moment) not all J2K decoders will support HTJ2K and thus it would be difficult to choose the correct plugin.
It might still not be that easy, as, according to #1054, there is no indication in the container whether the coded stream actually uses HT-J2K.
Fortunately, OpenJPEG can decode both, and we can simply add HTJ2K into the list of supported formats for it, but we also might need some code in core libheif to detect whether the stream is HTJ2K.

bradh:
> It might still not be that easy, as, according to #1054, there is no indication in the container whether the coded stream actually uses HT-J2K.

It would help to not have to parse the `CAP` segment. I've flagged that on the HTJ2K discord channel to see if could be possible.

libheif/context.cc
bradh: Will see if we need to add something extra for the HT case, and combine if not.

farindk:
Another reason to have a separate `heif_compression_HTJ2K` might be that we currently cannot process any encoder parameters before the plugin is selected. And the plugin is only selected based on the compression format.
I.e. if all J2K encoders would support HTJ2K, a simple "-p htj2k=true" would work, but since OpenJPEG doesn't encode HTJ2K yet, there cannot be a "-p htj2k=true".

Second option: we would put the burden on the user when the codec is selected. He has to know that "-e OpenJPEG" is ordinary J2K und "-e OpenJPH" is HTJ2K.
Encoders that support both might have a "-p htj2k" parameters.

Third option: we require that all JPEG2000 plugins have an ht2k parameter option and if it is set to a value not supported by the plugin, it fails with an error. Sounds like a silly idea, but we can have similar cases for other codecs, e.g. a h265 codec that cannot encode monochrome images. (Currently we handle this by the plugin encoding dummy color planes.)

Fourth option: when selecting the encoder, we pass a compression-format specific options struct to libheif. For Jpeg2000, this would include the HTJ2K flag and would be used for plugin selection. However, this may contradict the explicit user choice through "-e".

My subjective summary: option 1 sounds like the easy solution. Maybe in the future, when all encoders support everything, the two enums can be synonyms.
My unexpected second place goes to options 3 and 4, because we might need a more flexible plugin selection anyways.

bradh:
> My unexpected second place goes to options 3 and 4, because we might need a more flexible plugin selection anyways.

Maybe this is worth deeper consideration. I did some investigation into hardware based encoding and decoding (e.g. with NVIDIA GPUs). 

There are quite a few dimensions to whether the hardware can handle the task, beyond just the codec. [NVIDIA documentation for the decoder](https://docs.nvidia.com/video-technologies/video-codec-sdk/12.1/nvdec-video-decoder-api-prog-guide/index.html) and [encoder](https://docs.nvidia.com/video-technologies/video-codec-sdk/12.1/nvenc-application-note/index.html) shows differences in:

- bit depth
- supported profiles
- maximum size
- chroma

Running the tests code shows additional variation in MB count, minimum size and output surface format.

This is a tricky tradeoff, because we'd like to use the fastest codec that supports the configuration the user has, but that isn't really exposed in the plugin API.

libheif/context.cc
bradh: Have combined it.

bradh:
I have cleaned up the code, and think it is now ready for review.

The one thing that doesn't work yet is the quality setting. That is still to be implemented in OpenJPH.

Also, there is no OpenJPH decoder implementation yet. I do plan to do that, no schedule at this stage.

farindk:
Thank you.

PR #1075, opened 2023-12-21
docs: minor doxygen update

bradh:
No code changes, just a tidy up of existing entries.


farindk:
Thanks. I've added that the use of `bit_depth=24` or `=30` is deprecated.

PR #1077, opened 2023-12-22
Set kvazaar config threads to 0 if ENABLE_MULTITHREADING_SUPPORT is false

meowtec:
The background:

I am trying to build libheif to wasm with kvazaar.
The building is successful but it run with error from kvazaar:
```
--threads=auto value set to 4.
pthread_create failed!
Could not initialize threadqueue.
```

It seems that kvazaar doesn't provide a build option for disable threads, the only way is settings `config->threads` to `0` at runtime.


farindk:
Thanks

PR #1080, opened 2023-12-25
colorconversion: add support for promoting < 8 bit planar to 8 bit planar

bradh:
This is a partial fix for color conversion from planar RGB with bit depth less than 8 bit into 8 bit. It extends an existing operation (`Op_to_sdr_planes`) to handle this case.

This is enough to get my uncompressed codec rework with `heif-convert` to generate JPEG. However it is not sufficient for PNG.

### DIFF
--- a/libheif/color-conversion/hdr_sdr.cc
+++ b/libheif/color-conversion/hdr_sdr.cc
@@ -182,8 +182,27 @@ Op_to_sdr_planes::convert_colorspace(const std::shared_ptr<const HeifPixelImage>
             int in = p_in[y * stride_in + x];
             p_out[y * stride_out + x] = (uint8_t) (in >> shift); // TODO: I think no rounding here, but am not sure.
           }
-      }
-      else {
+      } else if (input_bits < 8) {
+        int width = input->get_width(channel);
+        int height = input->get_height(channel);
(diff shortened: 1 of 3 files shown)


libheif/color-conversion/hdr_sdr.cc
```diff
@@ -182,8 +182,27 @@ Op_to_sdr_planes::convert_colorspace(const std::shared_ptr<const HeifPixelImage>
...
+        int stride_in;
+        const uint8_t* p_in = input->get_plane(channel, &stride_in);
+
+        int stride_out;
+        uint8_t* p_out = outimg->get_plane(channel, &stride_out);
+
+        for (int y = 0; y < height; y++)
+          for (int x = 0; x < width; x++) {
+            int in = p_in[y * stride_in + x];
+            p_out[y * stride_out + x] = (uint8_t) (in << shift); // TODO: I think no rounding here, but am not sure.
```
farindk, line 203: The usual technique is to use `(in << shift) | (in >> (8-shift)`. This puts the MSBs into the LSBs that were shifted in. The advantage is that full white (e.g. 0x3F in 6 bit) also gets mapped to (0xFF in 8 bits) instead of some grey.

farindk:
I have changed the convert such that also inputs with <4 bits can be converted with the full output value range.

PR #1081, opened 2023-12-25
refactor uncompressed decode

### LINKED ISSUE #904: uncompressed decode for component order is incorrect

bradh:
This is a pretty large refactoring of the uncompressed codec.

There are two commits. The first splits the uncompressed boxes out into a separate file. That part is basically mechanical.

The second reworks how the uncompressed decode logic works. Previously we tried to pull the right bits into the image. This implementation is closer to the source - we step through the data and push into the image planes. There are multiple decoder implementations, and a factory method to choose which decoder to use. The choices at this point are oriented around interleave, but other options are possible.

There are no changes to the encode side.

There are now many more tests. We now support YCbCr 4:2:2 and 4:2:0 subsampling, and correctly handle component interleave. We also handle component alignment (for at least some cases).

Resolves #904

What is not yet supported:

- block_size != 0 (along with associated flags)
- components little endian
- high bit depth (except == 16)
- multi-Y interleave 
- 4:1:1 subsampling
- palette images
- component formats other than unsigned integer
- the "version == 1" `uncC` short-cut format

Even with all of that, this is a better basis than what we currently have.


farindk:
Thanks.

Issue #1084, opened 2023-12-31
Uncompressed codec enabled by default even if experimental?


davcohen:
Hi, 
I just discovered that there is CVE-2023-49460 + CVE-2023-49464 impacting libheif1 when the library is compiled with the `WITH_UNCOMPRESSED_CODEC` option and that the vulnerable function is used. So I asked myself, should this experimental option be enabled by default?

There is difference between the Default CMakeLists.txt option (https://github.com/strukturag/libheif/blob/3ebec35955aeff0a24ec132a6a208e60d51ea11d/CMakeLists.txt#L182) and the CMakePresets's default configuration (https://github.com/strukturag/libheif/blob/3ebec35955aeff0a24ec132a6a208e60d51ea11d/CMakePresets.json#L32).

If it was by default disabled (as it is in the CMakeLists.txt file), it should be the same for the `default` presets otherwise it is not coherent and the experimental option is enabled by default for all builds. 

Only the `develop` preset should enable it IMO.
The same for all experimental modes IMO to make the default library file more secure.

WDYT?

bradh:
I am sympathetic to your concern, but I think there could be some misunderstandings in this.

If you do a "really default" build, with something like "cmake ..", then you should not get the uncompressed codec. The reason for that is nothing to do with the CVEs, but rather that the standard (ISO/IEC 23001-17) is not finished yet. 

If you use a preset, you can get different results. How are you using the `--preset` option? In particular, why did you choose `default` (which is not documented)? Why are you concerned about this particular preset? Just the name?

I don't think the CVEs and experimental status are particularly related. We have reported vulnerabilities and fixes for other functions that are not experimental, just like those that are experimental. Changing the setting on a preset (irrespective of what it is named) won't make things more secure. Equally, changing it to `"OFF"` won't make it less secure, so you could choose to provide a PR if that would help you.

farindk:
I agree that the default preset should match the default cmake options. In consequence, this might mean that the whole `default` preset is actually useless. Maybe it would be better to abandon that preset and simply always define all options for every other preset.

bradh:
If the name is what is confusing people, perhaps call it `base`.

davcohen:
Thanks for the answers and indeed you are right, sorry for that. I thought CMAKE used the default preset when building it with cmake, which was wrong. I got confused with the name of the presets.
It might be good to change his name to avoid confusion but basically, my concern was more about the default build which I thought included an experimental feature.

[...]

PR #1087, opened 2024-01-02
Cleaner presets for CMake

davcohen:
In this PR, there is a new `default` preset in CMakePresets.json reflecting the CMakeLists.txt default plugin options and a `base` preset.

The change should not have any consequences in the builds except the ones building using the default preset (`cmake --preset=default ..`).

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -8,8 +8,42 @@
   "configurePresets": [
     {
       "name": "default",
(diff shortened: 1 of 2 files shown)


davcohen:
Related to #1084

farindk:
Thanks. I'll take this one step further and remove the `base` preset as it has no function is only a source of errors.

PR #1091, opened 2024-01-06
add support for iden derived images

bradh:
This adds support for `iden` derived images.

For simple cases, see https://github.com/nokiatech/heif_conformance/blob/master/conformance_files/C008.heic and https://github.com/nokiatech/heif_conformance/blob/master/conformance_files/C014.heic

I could see that as a pretty simple edit - where you don't want to redo the encoding, just to rotate / crop.

A more interesting case is https://github.com/nokiatech/heif_conformance/blob/master/conformance_files/C023.heic where four source images are cropped (`clap`) and then incorporated into a grid.

https://github.com/nokiatech/heif_conformance/blob/master/conformance_files/C039.heic is a bit trickier - it does two-stage `iden`. I could maybe see that as a progressive edit.

This PR supports all of those cases.

There is a potential issue around colour handling and alpha. We have code like https://github.com/strukturag/libheif/blob/master/libheif/context.cc#L836-L874 and https://github.com/strukturag/libheif/blob/master/libheif/context.cc#L1058-L1113 that does special handling for the `grid` case. Possibly `iden` needs to be added to that. I don't have any examples though, and am not confident about the changes.

### DIFF
--- a/go/heif/heif.go
+++ b/go/heif/heif.go
@@ -301,6 +301,8 @@ const (
 
 	SuberrorInvalidRegionData = C.heif_suberror_Invalid_region_data
 
+	SuberrorMissingIrefReference = C.heif_suberror_missing_iref_reference
+
 	SuberrorWrongTileImagePixelDepth = C.heif_suberror_Wrong_tile_image_pixel_depth
 
 	SuberrorUnknownNCLXColorPrimaries = C.heif_suberror_Unknown_NCLX_color_primaries
--- a/libheif/error.cc
+++ b/libheif/error.cc
@@ -158,6 +158,8 @@ const char* Error::get_error_string(heif_suberror_code err)
       return "Unknown NCLX matrix coefficients";
     case heif_suberror_Invalid_region_data:
       return "Invalid region item data";
+    case heif_suberror_missing_iref_reference:
+      return "'iref' box does not include a referenced item";
     case heif_suberror_Invalid_J2K_codestream:
       return "Invalid JPEG 2000 codestream";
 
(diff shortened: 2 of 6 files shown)


bradh:
@farindk Looks like the fuzzer died for an unrelated issue. Can you restart that?

bradh:
Disregard. I need to do more work on this.

fancycode:
> @farindk Looks like the fuzzer died for an unrelated issue. Can you restart that?

I restarted the CI job but it still fails when cloning `x265`:
```
#9 [4/8] RUN hg clone     --branch stable     http://hg.videolan.org/x265     x265
#9 sha256:fa22b89abc7931afe44e5c35377c72aec6d22b5d949a4b0b6748d9ab67dc26ef
#9 60.66 abort: HTTP Error 504: Gateway Time-out
#9 ERROR: process "/bin/sh -c hg clone     --branch stable     http://hg.videolan.org/x265     x265" did not complete successfully: exit code: 255
```

bradh:
I have cleaned up this, and think it is now ready.

The PR at #1092 is (essentially) the same. It passed CI except for one issue with OSX, which seems unrelated.

farindk:
I think this worked before and was broken some time ago.
This much simpler patch also makes `iden` images work again:
````
diff --git a/libheif/context.cc b/libheif/context.cc
index 6b6866d..b2f99fa 100644
--- a/libheif/context.cc
+++ b/libheif/context.cc
@@ -879,7 +879,7 @@ Error HeifContext::interpret_heif_file()
   for (heif_item_id id : image_IDs) {
     std::string item_type = m_heif_file->get_item_type(id);
     // skip region annotations, handled next
-    if (item_type == "rgan") {
+    if (item_type == "rgan" || item_type == "iden") {
       continue;
     }
     std::string content_type = m_heif_file->get_content_type(id);
````

I'll think about it, because I think the logic in that block should be rewritten anyways.

farindk:
I used the alternative, simpler fix.
Thanks anyway to have identified the regression.

Issue #1095, opened 2024-01-07
[macOS] WITH_FFMPEG_DECODER -> Undefined symbol


bigcat88:
Building libheif with brew with modified formula:

`
-DWITH_FFMPEG_DECODER=ON
-DWITH_FFMPEG_DECODER_PLUGIN=ON
`

```
[100%] Linking CXX shared module libheif-ffmpegdec.so
cd /tmp/libheif-20240107-59963-jb96bl/libheif-1.17.6/build/libheif/plugins && /opt/homebrew/Cellar/cmake/3.28.1/bin/cmake -E cmake_link_script CMakeFiles/heif-ffmpegdec.dir/link.txt --verbose=1
/opt/homebrew/Library/Homebrew/shims/mac/super/clang++ -O3 -DNDEBUG -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk -bundle -Wl,-headerpad_max_install_names -o libheif-ffmpegdec.so "CMakeFiles/heif-ffmpegdec.dir/decoder_ffmpeg.cc.o" "CMakeFiles/heif-ffmpegdec.dir/__/error.cc.o" "CMakeFiles/heif-ffmpegdec.dir/__/heif_plugin.cc.o"  -Wl,-rpath,/tmp/libheif-20240107-59963-jb96bl/libheif-1.17.6/build/libheif /opt/homebrew/lib/libavcodec.dylib ../libheif.1.17.6.dylib 
ld: Undefined symbols:
  _av_version_info, referenced from:
      ffmpeg_plugin_name() in decoder_ffmpeg.cc.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libheif/plugins/libheif-ffmpegdec.so] Error 1
```

Who have a clue and can point me where should I look?

_P.S: This is part of my tests/home_research of the libheif plugin  system to understand what is missing in order to start using it in the python pypi infrastructure..._

bigcat88:
https://github.com/strukturag/libheif/blob/98b987fa7317a42eb1ef6d35aaa9c881f8a5153b/libheif/plugins/decoder_ffmpeg.cc#L76-L82


Commenting out call to `av_version_info()` (by replacing it with `"Unknown"`) leads to successful compilation and working(_probably as I see_) ffmpeg decoder.

bigcat88:
[...]

PR #1097, opened 2024-01-07
ffmpeg: added reference to "avutil"

bigcat88:
Closes: #1095


farindk:
Right, that makes sense. Thank you.

PR #1099, opened 2024-01-10
plugins: fix return value for X_get_parameter_string()

bradh:
In the encoder plugins, we have code that looks like:
https://github.com/strukturag/libheif/blob/master/libheif/plugins/encoder_aom.cc#L648-L678

The problem is that if we return `heif_error_unsupported_parameter` even if it is valid.

That doesn't show up in `heif-enc` example, because we ignore the return value. However it doesn't look reasonable.

There is another problem in some of the codecs, where we assume the only string parameter is `chroma`. 

This PR addresses both issues across the codecs that have the problem (pretty much anything that supports a string parameter, given the way the plugins have developed).

Note that there are likely similar problems in other methods, but I wanted to keep this PR simple.

### DIFF
--- a/libheif/plugins/encoder_aom.cc
+++ b/libheif/plugins/encoder_aom.cc
@@ -660,6 +660,7 @@ struct heif_error aom_get_parameter_string(void* encoder_raw, const char* name,
         assert(false);
         return heif_error_invalid_parameter_value;
     }
+    return heif_error_ok;
   }
   else if (strcmp(name, kParam_tune) == 0) {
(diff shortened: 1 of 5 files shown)


bradh:
Looks like something is wrong with CIFuzz - failing x265 installation again.

bradh:
Probably needs configuration change to move to `https://bitbucket.org/multicoreware/x265_git/src/stable/` - they don't appear to be on mercurial any more.

farindk:
Right. Thank you.

PR #1104, opened 2024-01-11
Fix potential memory leak in file heif_convert.cc

xiaoxiaoafeifei:
Fix potential memory leak in file heif_convert.cc


farindk:
Thank you.

PR #1110, opened 2024-01-16
jpeg2000: enhance parsing of JPEG 2000 codestream main header

bradh:
This updates the existing SIZ parsing, and adds CAP segment parsing.

Also adds unit tests, and an optional test coverage check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -393,6 +393,24 @@ endif (DOXYGEN_FOUND)
 
 # --- Testing
 
+option(ENABLE_COVERAGE "" OFF)
+if(ENABLE_COVERAGE)
+  # set compiler flags
+  set(CMAKE_CXX_FLAGS "-O0 -coverage")
+
+  # find required tools
+  find_program(LCOV lcov REQUIRED)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
kmilos, line 1: Take this out to a sperate PR?

bradh: Do you mean the coverage part? That could be dropped if its not wanted. I just found it useful when writing the tests for the parsing logic.

farindk: I'll just merge both together... Not necessary to spend the time just to split that.

kmilos: Yes, just this CMakeLists.txt change - it's kinda independent of this J2K work, and can be useful on its own, no?

kmilos: > I'll just merge both together... Not necessary to spend the time just to split that.

At least separate commits then please. Easier to follow though git blame, eventually revert etc.

farindk:
Thanks, Brad. I've split it into two commits and fixed the uin16_t datatype.

PR #1112, opened 2024-01-16
jpeg2000: add missing test code for JPEG 2000 encoding

bradh:
This was in #1071, but appears to have been dropped during the split of that PR.


farindk:
Yes, I missed to add it to the commit...

PR #1113, opened 2024-01-17
openjpeg: fix memory leaks

bradh:
These show when running the unit tests or heif-enc under valgrind.


farindk:
BTW: I've made good experience with wrapping C-pointers in a `shared_ptr` with the deallocation function as custom deleted. That eliminates all cleanup duties in the error cases.

PR #1121, opened 2024-02-02
Add experimental gain map support

DichenZhang1:
This feature is hidden behind the WITH_EXPERIMENTAL_GAIN_MAP compile flag.

    See https://helpx.adobe.com/camera-raw/using/gain-map.html for an introduction
    to gain maps. Note that in this change, gain maps are not stored as auxiliary
    items as proposed by Adobe, but as a hidden input to a new 'tmap' (tone mapped)
    derived item.

    An 'altr' (alternatives group) box is also added to tell the decoder that it
    may show the tone mapped image (preferrably) or the base image as a fall back.
    The primary item id still points to the base image for backward compatibility.


DichenZhang1:
Hi there!

Can someone help to review this pull request? This is adding gain map support for HEIC and AVIF images. 

Thank you!

farindk:
Thank you for the contribution. I generally looks good.

The `GainMapMetadata` is currently a private data class, but also exposed in the API.
I guess that this data should also be public.
In that case, we should rename this to follow the usual naming scheme (like `heif_gain_map_metadata`), make it a POD struct and move it to a public header. We can add a new header file (e.g. `heif_hdr.h`) where we collect HDR related things so that `heif.h` is not getting too long.

farindk:
Is this standardized anywhere? Could you point me to the specification?

The Adobe document says:

> Gain Maps are currently undergoing standardization in ISO/TC 42 Photography. Furthermore, various
> technical committees have ongoing efforts to standardize Gain Map storage in different file formats.
> Consequently, the original Gain Map storage proposals have been removed from this document to avoid
> confusion and conflicting information.

Does that mean that there is no final spec yet? We can include it as "experimental" anyway, but have to make it very clear that the spec is not finalized yet.

Do you have any kind of test program? Maybe a patched `heif_convert` that I could use to test this?

DichenZhang1:
Hi Dirk,

Thank you for looking into this!

I've changed the `GainMapMetadata` name to `heif_gain_map_metadata` and made it public. The spec is not finalized yet, and I'll keep monitoring the updates. For verification, I've added some code in `heif_enc` and `heif_info` and they can be a closed loop for encoding and decoding. But we only have a pseudo gain map metadata (all values are zeros) and ideally this should be an input from user.

Thank you!

libheif/heif.h
```diff
@@ -966,6 +969,19 @@ struct heif_error heif_context_get_image_handle(struct heif_context* ctx,
                                                 heif_item_id id,
                                                 struct heif_image_handle**);
 
+#if WITH_EXPERIMENTAL_GAIN_MAP
+// Get id of the gain map image of the HEIF file. If no gain map image is available, this
+// method will return heif_suberror_No_item_data.
+LIBHEIF_API
+struct heif_error heif_context_get_gain_map_image_ID(struct heif_context* ctx, heif_item_id* id);
```
farindk, line 976: I assume there can be several gain maps in the file. We should add the color image ID as another parameter.

farindk: Same for the two functions below.

libheif/context.cc
```diff
@@ -517,6 +517,18 @@ static bool item_type_is_image(const std::string& item_type, const std::string&
           item_type == "mski");
 }
 
+#if WITH_EXPERIMENTAL_GAIN_MAP
+static bool item_type_is_gain_map_image(const std::string& item_type, const std::string& item_name) {
+  return  (item_name == "GMap" &&
```
farindk, line 522: It appears that images are not using a `GMap` `item_name`. (see my comment in the discussion.)

libheif/heif.h
DichenZhang1: I believe this is intended that we only support one gain map at most, and that gain map is for the primary image. Assume if we have multiple color images and multiple gain maps, and it is likely that users accidentally links one gain map to another gain map, and we can't detect that case.

But I agree that  the standard is still in draft stage and this is not clearly defined. I think we can hold this for a little while and still mark this gain map support as "experimental".

PR #1132, opened 2024-02-26
uncompressed: support version 1 boxes

bradh:
This adds support for parsing uncompressed heif files using version 1 of the `uncC` box, which is a shortcut for simple RGB, RGBA and ABGR files at 8bpp. 

I looked at trying to synthesize the extra box (`cmpd`) at parse time. That didn't work out, because the item property association then didn't line up. Instead, the extra boxes are produced when the item properties are requested.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -2193,7 +2193,14 @@ Error Box_ipco::get_properties_for_item_ID(uint32_t itemID,
     }
 
     if (assoc.property_index > 0) {
-      out_properties.push_back(allProperties[assoc.property_index - 1]);
+      std::shared_ptr<Box> prop = allProperties[assoc.property_index - 1];
+      out_properties.push_back(prop);
+      std::shared_ptr<std::vector<std::shared_ptr<Box>>> implied_props = prop->get_implied_boxes();
+      if (implied_props) {
+        for(auto& implied_prop: *implied_props) {
+          out_properties.push_back(implied_prop);
+        }
+      }
     }
   }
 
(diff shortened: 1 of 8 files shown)


farindk:
I'm not sure if we should go with the 'implied boxes' concept. When we add the planned low-level boxes interface to modify existing files and copy over raw boxes, the 'implied boxes' will always be a special case that has to be taken care of.

Two alternative proposals:
- the uncompressed codec looks for both variants and then internally generates the implied boxes. They are never visible outside the uncompressed codec.
- we do not generate implied boxes at all, but you have an abstract interface to extract the data so that it is completely transparent to the codec whether it is coming from the minimized form or not.

bradh:
We should definitely look at what is coming for the metabox V1 proposal - see 
https://www.mpeg.org/wp-content/uploads/mpeg_meetings/146_Rennes/w23988.zip

On the second alternative, I considered whether we might use the visitor pattern. So each property would be called with the image data and transform it if needed. However that looked like quite a lot of redesign, and would also be a bit difficult to invert for the "which tiles to correspond to a ROI?" case.

farindk:
> On the second alternative, I considered whether we might use the visitor pattern. So each property would be called with the image data and transform it if needed. However that looked like quite a lot of redesign, and would also be a bit difficult to invert for the "which tiles to correspond to a ROI?" case.

Yes, that sounds like too much for me. I was thinking about something much simpler. Could be just a large data-structure that holds all information from the `uncC` box, irrespective how it was read.
I have something slightly similar in the `Box_cmin` where there is a `RelativeIntrinsicMatrix`, which is the data as it is stored in the box and an `AbsoluteIntrinsicMatrix`, which is the data the library works with.
In a similar way, you could have a type (e.g. `UncompressedConfiguration`) that holds all the information independent from any box implementation, and fill this either from the separate boxes or the minimized box.

bradh:
See #1198 for an alternative approach.

PR #1136, opened 2024-02-28
explicitly state default quality setting in --help output

felix-red-panda:
minor change so users can easily determine the default quality setting


farindk:
Users can easily see the default quality using the `-P` option that lists all parameters.

farindk:
Sorry, I think that it does not make sense to use the same quality default for every codec. However, I have added an issue #1200 to investigate this further.

PR #1141, opened 2024-03-10
encoder_kvazaar.cc: Fix some memory leaks

hlewin:
This fixes the following memory leaks:
- the `data` chunks are not freed in all cases after being appended to the encoder output
- `append_chunk_data` throwing an exception (std::bad_alloc) would not free any of the kvazaar-structures

At least the successful encoding case should be leak-free now: The address sanitizer does not report any leaks.
I did not test the exceptional cases - using unique_ptr this should be no different than the successful case.


farindk:
Thank you

Issue #1147, opened 2024-03-18
Files with shared auxiliary images fail to parse


leo-barnes:
Filing issue to document email thread with @farindk.

ImageMagick is rejecting some prototype files that I have been creating with the following error:
identify: Invalid input: Unspecified: Too many auxiliary image references (2.0) `file_with_shared_alpha.heic' @ error/heic.c/IsHEIFSuccess/142.

The error is coming from libheif rather than ImageMagick itself. The file in question has two main images (in an 'altr' group) that share an alpha auxiliary image. To me this should be perfectly valid and makes a lot of sense once you start creating files containing 'altr' groups, but I can see why this would be unexpected for simpler files.

The conclusion from the thread is that there is nothing in the spec that explicitly forbids sharing the same auxiliary image between multiple main images and that the restriction in libheif should probably be relaxed. ISOBMFF says that all irefs of a specific type from a source item should be collected into a single ItemReferenceBox, so having multiple irefs is not really an option.
[file_with_shared_alpha.heic.zip](https://github.com/strukturag/libheif/files/14641161/file_with_shared_alpha.heic.zip)

bradh:
It looks like we just need to add the auxilary image to each of the parents. Might be some complexity with loop detection, but will check that.

One question before I look to PR this: would it make more sense to have the `altr` group (in this test file, item 4) as the target? Is that at least a possibility that should be covered / supported?

leo-barnes:
The same issue might also be present for shared metadata and shared thumbnails. Creating a file with shared EXIF payloads seems to result in the following error from ImageMagick:
`identify: Invalid input: Unspecified: Metadata not correctly assigned to image (2.0)`
(Assuming I didn't mess up creating the file)

bradh:
@leo-barnes Can you share those additional test files?

leo-barnes:
Attached a file that shares alpha, thumbnail and XMP between the two main images in an `altr` group.
[file_with_shared_everything.heic.zip](https://github.com/strukturag/libheif/files/14665966/file_with_shared_everything.heic.zip)

bradh:
I've got it to a "less broken" stage:

```
$ ./build/examples/heif-info ./file_with_shared_everything.heic 
MIME type: image/heic
main brand: heic
compatible brands: heic, mif1

image: 512x384 (id=1), primary
  colorspace: YCbCr, 4:2:0
  bit depth: 8
  thumbnail: 128x96
  color profile: no
  alpha channel: yes 
[...]

PR #1148, opened 2024-03-19
libheif: support for shared auxl alpha images

bradh:
Initial work (for discussion / review) on support for shared alpha images.

Relates #1147 

I've shown some commented out code that returns an error if the alpha requires scaling (which is valid in HEIF). Until we support that, the options are silently ignoring the alpha plane, or not supporting the image at all.

We do have some scaling code, but my attempt to apply it failed - its designed for an image, not an image plane, and the refactoring got away.

Fixes / suggestions appreciated.

### DIFF
--- a/libheif/context.cc
+++ b/libheif/context.cc
@@ -573,8 +573,6 @@ Error HeifContext::interpret_heif_file()
 
 
   // --- read through properties for each image and extract image resolutions
-  // Note: this has to be executed before assigning the auxiliary images below because we will only
-  // merge the alpha image with the main image when their resolutions are the same.
 
   for (auto& pair : m_all_images) {
     auto& image = pair.second;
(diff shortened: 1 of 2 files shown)


bradh:
> I see the same issue for `cdsc`:

Yes (and a bit lower for the region case too). Not too hard to fix.

> Presumably the same holds for `thmb` references.

This (and depth images) will require a bit more refactoring.

bradh:
Looks like I need to address the fuzzing issue....

bradh:
This should now fully address #1147 

Also cleans up some unused members that made the problem look more complicated.

leo-barnes:
Hi @bradh!

Do you have a time-line for when this will land?

bradh:
> Do you have a time-line for when this will land?

I think it is complete but cannot progress it beyond this point.

farindk:
Yes, I know I'm the bottleneck. I'm currently completely blocked by other projects. Will have a look as soon as I can spend time on this.

leo-barnes:
Got it, no worries. Was just wondering what the plan was and whether there was anything spec or info-wise that was blocking you.

farindk:
No worries at all. I didn't think that for a single second.
Just wanted to let you know that I'm fine, healthy and happy. Just working on something different and urgent at the moment.

farindk:
Thanks, this is great.

One thing: I'm not sure if libheif should automatically scale the alpha channel. It is not clear what scaling filter should be used and whether the user wants this. At least, we might want to enable this as a decoding option. But we'll probably have to postpone this since currently, even libheif assumes that the alpha channel has the same resolution.

PR #1154, opened 2024-03-20
doc: explain quality a bit more for heif-convert

### LINKED ISSUE #1150: Add info to Man page; Quality; png
Using the command `heif-convert` with the quality option `-q [0...100]` only has an effect when the output type is jpg.
The option has no effect for `.png` and `.y4m`.

It would be cool if that could be reflected in the man pages.

bradh:
Minor update suggested at #1150 and an extra cut-n-paste fix.

Resolves #1150.

A full rework of the man page can come after #985


farindk:
thanks

PR #1156, opened 2024-03-26
Add uuid property

dukesook:
I'm interested in adding uuid property getters/setters to enable user-defined image metadata.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -392,12 +392,14 @@ Error Box::parse(BitstreamRange& range)
   }
   else {
     uint64_t content_size = get_box_size() - get_header_size();
-    if (range.prepare_read(content_size)) {
-      if (content_size > MAX_BOX_SIZE) {
-        return Error(heif_error_Invalid_input,
-                     heif_suberror_Invalid_box_size);
-      }
-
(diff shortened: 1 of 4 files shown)


dukesook:
I wasn't sure if adding a new Box_uuid class to box.h was a good idea. Instead, I extended the Box class to handle the uuid functionality.

Any feedback would be appreciated.

farindk:
I wouldn't mix this with the base `Box` class, but introduce a catch-all `Box_other` class that keeps the data for everything that libheif does not (yet) understand. I wouldn't even make it uuid specific, but also for other (4cc) boxes that we do not parse yet.

farindk:
I have implemented this in a slightly different way. It is now not limited to 'uuid' boxes anymore, but one can add all sorts of boxes with raw data, and also read the raw data of boxes that are not understood by libheif.

Only caveat at this time: we cannot read the raw data of boxes that were successfully parsed by libheif. This should be added, preferably by storing a pointer to the source data in the Box. But a clean implementation needs more time.

farindk:
@dukesook Please review and if you think that this is ok, I'll close this PR.

dukesook:
This is a great implementation, thanks for doing it. Go ahead and close this PR.

PR #1158, opened 2024-04-02
libheif: check compatible brands in the ftyp box

bradh:
Keeps the existing `heif_check_filetype` behaviour since that is public API. Adds a new function to check both the major version and the compatible_brands list. Switches the heif-convert example to use that new function.

I have put an arbitrary 512 byte limit on how long we expect the `ftyp` box could be. That seems reasonable, but could be higher or lower.

The goal of this is to support files that are both video and images (sharing data), where the major_brand is something like `mp41` but `heic` is listed in the `compatible_brands[]`.

### DIFF
--- a/examples/heif_convert.cc
+++ b/examples/heif_convert.cc
@@ -380,20 +380,26 @@ int main(int argc, char** argv)
   // TODO: check, whether reading from named pipes works at all.
 
   std::ifstream istr(input_filename.c_str(), std::ios_base::binary);
-  uint8_t length[4];
-  istr.read((char*) length, 4);
+  std::vector<uint8_t> length(4);
+  istr.read((char*) length.data(), length.size());
   uint32_t box_size = (length[0] << 24) + (length[1] << 16) + (length[2] << 8) + (length[3]);
   if ((box_size < 16) || (box_size > 512)) {
-    fprintf(stderr, "Input file does not appear to start with a valid box length\n");
+    fprintf(stderr, "Input file does not appear to start with a valid box length.");
+    if (box_size == 0xFFD8FFE0) {
+      fprintf(stderr, " Possibly could be a JPEG file instead.\n");
+    } else {
+      fprintf(stderr, "\n");
+    }
     return 1;
   }
 
-  std::vector<uint8_t> ftyp_bytes(box_size - 4);
-  istr.read((char*) ftyp_bytes.data(), ftyp_bytes.size());
+  std::vector<uint8_t> ftyp_bytes(box_size);
+  std::copy(length.begin(), length.end(), ftyp_bytes.begin());
+  istr.read((char*) ftyp_bytes.data() + 4, ftyp_bytes.size() - 4);
 
-  enum heif_filetype_result filetype_check = heif_check_filetype_full(ftyp_bytes.data(), (int)ftyp_bytes.size());
-  if (filetype_check != heif_filetype_yes_supported) {
-    fprintf(stderr, "Input file is not a supported format\n");
+  heif_error filetype_check = heif_has_compatible_filetype(ftyp_bytes.data(), (int)ftyp_bytes.size());
+  if (filetype_check.code != heif_error_Ok) {
+    fprintf(stderr, "Input file is not a supported format. %s\n", filetype_check.message);
     return 1;
   }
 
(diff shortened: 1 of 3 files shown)


examples/heif_convert.cc
```diff
@@ -380,27 +380,23 @@ int main(int argc, char** argv)
...
-  istr.read((char*) magic, 12);
-
-  if (heif_check_jpeg_filetype(magic, 12)) {
-    fprintf(stderr, "Input file '%s' is a JPEG image\n", input_filename.c_str());
+  uint8_t length[4];
+  istr.read((char*) length, 4);
+  uint32_t box_size = (length[0] << 24) + (length[1] << 16) + (length[2] << 8) + (length[3]);
+  if ((box_size < 16) || (box_size > 512)) {
+    fprintf(stderr, "Input file does not appear to start with a valid box length\n");
     return 1;
```
farindk, line 388: I deliberately checked for JPEG here because there were several "bug reports" that claimed that a heic file could not be decoded when it was in fact a JPEG file.

libheif/heif.cc
```diff
@@ -145,6 +145,50 @@ heif_filetype_result heif_check_filetype(const uint8_t* data, int len)
...
+  for (int offset = 12; offset < len; offset += 4) {
+    heif_brand2 compatible_brand = heif_fourcc_to_brand((char*) (data + offset));
+    it = supported_brands.find(compatible_brand);
+    if (it != supported_brands.end()) {
+      return heif_filetype_yes_supported;
+    }
+  }
+
+  return heif_filetype_yes_unsupported;
+}
```
farindk, line 189: Should probably return `heif_filetype_no` at the end.

libheif/heif.cc
```diff
@@ -145,6 +145,50 @@ heif_filetype_result heif_check_filetype(const uint8_t* data, int len)
 }
 
 
+heif_filetype_result heif_check_filetype_full(const uint8_t* data, int len)
+{
```
farindk, line 149: Proposal for a better name: `heif_is_compatible_filetype` ("is"/"has", I'm not sure).

In order to make using the function easier for the user, the input data should just be the first `len` bytes from the file.
Maybe return a `heif_error` that can tell exactly why it is not compatible or when the `len` is too small to cover the whole ftyp box.

Maybe, we can also reuse `heif_has_compatible_brand` in some way.

examples/heif_convert.cc
bradh: OK. That was a problem when I didn't have the whole file. However I can put it back in as part of passing the full number of bytes.

libheif/heif.cc
bradh: I think the idea of returning an `heif_error` will probably cover this - we can say whether the box structure / signature looked bad, or it didn't contain a compatible brand, or whether it is OK.

libheif/heif.cc
bradh: > Proposal for a better name: `heif_is_compatible_filetype` ("is"/"has", I'm not sure).

I will do this. Probably "has" but will check how it reads in the caller code.

> In order to make using the function easier for the user, the input data should just be the first `len` bytes from the file. Maybe return a `heif_error` that can tell exactly why it is not compatible or when the `len` is too small to cover the whole ftyp box.

Will do. I was trying to work around the streaming read case, but in retrospect it probably breaks in other places.

> Maybe, we can also reuse `heif_list_compatible_brands` in some way.

Will check this. On first look, it seems that we don't really need the list, but could check the same underlying vector to avoid hard-coding the `supported_brands` set.

farindk: > Maybe, we can also reuse heif_list_compatible_brands in some way.

The idea was that using `heif_list_compatible_brands`, your function could be simply calling that and checking whether a supported brand is in the returned list. Thus no more need for all the file handling or checking for incorrect parameters.

bradh: Ah, I get it now. There would still need to be a bit more checking if you want the JPEG sanity check, but it would do most of the work. 

Would adding the JPEG check to `heif_list_compatible_brands` (i.e. returning a different error for JPEG signature) likely be acceptable?

bradh: With the current header length check, it can never match JPEG. However I can do a check for the first four bytes matching `0xFF, 0xD8, 0xFF, 0xE0` if it would not be a valid length.

libheif/heif.cc
bradh: Fixed.

libheif/heif.cc
bradh: Should now be fixed.

examples/heif_convert.cc
bradh: Adapted to do a partial check based on the first four bytes.

PR #1173, opened 2024-05-20
Replace `pwd` with $(pwd) for command substitution

wantehchang:


wantehchang:
@farindk Dirk: Please review and merge. Thanks!

PR #1175, opened 2024-05-21
Reuse matching ispe box if one exists.

dukesook:
Reusing identical properties will help clean up the ipco box. It becomes cumbersome when there are many images like this:
image

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -753,6 +753,10 @@ int Box::find_or_append_child_box(const std::shared_ptr<Box>& box)
 
 bool Box::operator==(const Box& other) const
 {
+  if (this->get_short_type() != other.get_short_type()) {
+    return false;
+  }
+
   StreamWriter writer1;
   StreamWriter writer2;
 
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -742,7 +742,7 @@ std::vector<std::shared_ptr<Box>> Box::get_child_boxes(uint32_t short_type) cons
 
 int Box::find_or_append_child_box(const std::shared_ptr<Box>& box)
 {
-  for (size_t i = 0; i < m_children.size(); i++) {
+  for (int i = 0; i < (int) m_children.size(); i++) {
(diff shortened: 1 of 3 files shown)


bradh:
I like the concept, but it seems like we should do it for more than just `ispe`.  For example, adding `pixi` would also make it simpler, and `hvcC` would likely save significant space.

I haven't got all the way through the process, but maybe we can always make the box, and do an operator `==` implementation on the box class to check if its the same. The complexity would be to check equivalence, but having it in the box implementation would encapsulate it for changes (e.g. if a new version got added, we wouldn't forget to update the check over in file.cc. 

If that worked out, we could change:

```
int index = m_ipco_box->append_child_box(ispe_clap_pixi_whatever);
```
to be something like:
```
int index = m_ipco_box->find_or_append_child_box(ispe_clap_pixi_whatever);
```

Thoughts?

farindk:
This absolutely makes sense. I also think that there should be an `operator==` to do this check for all box types. Maybe the equivalence could be tested directly on the binary data. Then we have it automatically for every box and there is no need to maintain the operator== for each class.

farindk:
@dukesook Looks good. Maybe you could test `short_type` for equivalence before writing out both boxes to increase efficiency.

farindk:
Thank you. That looks good. I'll have a closer look and merge this in the coming days.

dukesook:
I updated the PR with the provided feedback.

The Box class now:
 1. Overrides the == operator
 2. Provides an equal() function to work with shared_ptrs
 3. Provides a generic find_or_append_child_box() function to prevent redundant boxes. (This works great with pixi & ispe, but the hvcC box is added first and later populated with values)

bradh:
LGTM. I think it would be worth unit testing, so sent a PR for that.

farindk:
Thank you. This is great to have.

farindk:
I've made the `operator==` virtual so that we can still add more efficient equality comparisons.

PR #1180, opened 2024-05-31
Encode Grid

### LINKED ISSUE #648: Writing Gridded Imagery
I want to write a heif that is gridded. In other words, the image is divided into an arbitrary number of tiles. 

Does libheif support this feature? If not, where would be an ideal place to implement it?

### LINKED ISSUE #1176: How to create a grid HEIF
I saw the code do ImageGrid::write() but I don't know how to use it. I think this is related with Box_iref, iloc and iinf box.
Is there spec or handbook to do this? I have 23008-12 and 14496-12 spec but don't understand how this works..
Could somebody give me some hint?

And is iref_box can reference to single image encoding, not grid?

dukesook:
LIBHEIF_API
struct heif_error heif_context_encode_grid()

Resolves #1176, resolves #648

### DIFF
--- a/libheif/context.cc
+++ b/libheif/context.cc
@@ -2314,6 +2314,66 @@ Error HeifContext::encode_image(const std::shared_ptr<HeifPixelImage>& pixel_ima
   return error;
 }
 
+Error HeifContext::encode_grid_image(const std::vector<std::shared_ptr<HeifPixelImage>>& pixel_images,
+                                     uint16_t rows,
+                                     uint16_t columns,
+                                     struct heif_encoder* encoder,
+                                     const struct heif_encoding_options& options,
+                                     enum heif_image_input_class input_class,
+                                     std::shared_ptr<Image>& out_grid_image)
+{
+  Error error;
+
+  // TODO: the hdlr box is not the right place for comments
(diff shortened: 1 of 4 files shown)


farindk:
Thank you. This is nice to have.

In the future, we should have another way to build the grid image step by step by adding tiles one by one, but this is a nice convenience function that can still come in handy.

dukesook:
Great, thanks for merging.

I also wanted a mechanism for encoding grids that are internal to the codec (for example, the uncompressed codec). But I wasn't sure what the best way to do this was.

farindk:
> I also wanted a mechanism for encoding grids that are internal to the codec (for example, the uncompressed codec). But I wasn't sure what the best way to do this was.

Interesting question.
@bradh: Would it be possible to write uncompressed image tiles one by one? Does that pose any problems when the data is `zlib` compressed?

bradh:
It should be possible to write tiles one-by-one. There is a way to say "each of these extents needs to be done separately", and to specify the range.  That opens up the possibility to recycle the bits in one tile that is used to fill many grid spots (e.g. when using a map, and there is a lot of blue ocean; or when making black fill tiles for an image that is rotated to be north-up).

PR #1185, opened 2024-06-11
Fix some bugs in ffmpeg decoder plugin

roticv:
* memory leak fix
* heap buffer overflow fix

These issues were caught by ASAN


libheif/plugins/decoder_ffmpeg.cc
```diff
@@ -177,7 +177,11 @@ static struct heif_error ffmpeg_v1_push_data(void* decoder_raw, const void* data
       NalUnit* nal_unit = new NalUnit();
       nal_unit->set_data(cdata + ptr, nal_size);
 
+      NalUnit* old_nal_unit = decoder->NalMap[nal_unit->unit_type()];
```
farindk, line 180: Why not delete it immediately?
You also don't need to test for NULL. `delete` handles this automatically.

roticv: Makes sense. Will update

roticv: Fixed in ed27e0d0fa0ea797b94785459787352388a87437

farindk:
Thank you

PR #1186, opened 2024-06-12
tests: add unit test for box equality operations

bradh:
I originally proposed this against the work that Devon did, but its still valid with the recent rework.


farindk:
looks good

farindk:
I did a wrong click... Thanks

PR #1193, opened 2024-06-25
[ffmpeg plugin] Add more chroma subsampling support

roticv:
In this patch, support for mono, 422, 444 are added.

In terms of implementation:
* The check whether a pixel format is supported is moved into ffmpeg_get_chroma_format
* chroma_width and chroma_height needs to be computed separately for 422, 444 as they are different from 420

For testing: I created mono, YUV422, YUV422 samples with heif-enc and ran with ffmpeg plugin with ASAN. I validated that chroma_height and chroma_width are correctly set as they can cause heap-read-out-of-bound if not set correctly.

### DIFF
--- a/libheif/plugins/decoder_ffmpeg.cc
+++ b/libheif/plugins/decoder_ffmpeg.cc
@@ -187,6 +187,60 @@ static struct heif_error ffmpeg_v1_push_data(void* decoder_raw, const void* data
   return heif_error_success;
 }
 
+static heif_chroma ffmpeg_get_chroma_format(enum AVPixelFormat pix_fmt) {
+    if (pix_fmt == AV_PIX_FMT_GRAY8)
+    {
+        return heif_chroma_monochrome;
+    }
+    else if ((pix_fmt == AV_PIX_FMT_YUV420P) || (pix_fmt == AV_PIX_FMT_YUVJ420P) ||
+        (pix_fmt == AV_PIX_FMT_YUV420P10LE))
+    {
+        return heif_chroma_420;
+    }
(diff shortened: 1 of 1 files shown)


libheif/plugins/decoder_ffmpeg.cc
```diff
@@ -187,6 +187,60 @@ static struct heif_error ffmpeg_v1_push_data(void* decoder_raw, const void* data
...
+
+static int ffmpeg_get_chroma_width(const AVFrame* frame, heif_channel channel, heif_chroma chroma)
+{
+    if (channel == heif_channel_Y)
+    {
+        return frame->width;
+    }
+    else if (chroma == heif_chroma_420 || chroma == heif_chroma_422)
+    {
+        return (frame->width) / 2;
```
farindk, line 220: Do we need a (w+1)/2 here ?

libheif/plugins/decoder_ffmpeg.cc
```diff
@@ -187,6 +187,60 @@ static struct heif_error ffmpeg_v1_push_data(void* decoder_raw, const void* data
...
+
+static int ffmpeg_get_chroma_height(const AVFrame* frame, heif_channel channel, heif_chroma chroma)
+{
+    if (channel == heif_channel_Y)
+    {
+        return frame->height;
+    }
+    else if (chroma == heif_chroma_420)
+    {
+        return (frame->height) / 2;
```
farindk, line 236: Also (h+1)/2 ?

farindk:
Thanks, I've merged it and  I've added to chroma rounding in the above commit.

PR #1194, opened 2024-06-25
initial work on generic compression decoding

bradh:
This is from the draft Amendment 2 for ISO/IEC 23001-17.

It re-uses and generalises the existing zlib support (from header compression), and adds an optional dependency on Brotli.

It also picks up a few extra Go and JS error codes that I noted when adding mine. Those could be a separate commit if preferred.

Feedback on the approach and internal API changes would be appreciated.

I still have some work to do on this:

  - [ ] decode row compression
  - [ ] decode pixel compression (needs test data)
  - [ ] icbr byte ranges (needs test data)
  - [ ] some unit tests

There is no encoding support for this yet, and we'd likely need some kind of configuration option to specify that. It seems a bit premature given where the draft is up to.


Jamaika1:
Only applies to png and jpg file decoder. Tiff isn't taken into account.

bradh:
> Only applies to png and jpg file decoder. Tiff isn't taken into account.

I think I have no idea what you are referring to. This is intended to decode zlib/deflate/brotli compressed HEIF. What are you expecting to happen?

Jamaika1:
It's true. I don't know everything. zlib and deflate was previously added for png. Brotli? I guess only jpegxl uses brotli. libtiff I see that the entire file can be added to libheif. It has lot of compression algorithms. (lerc,zstd,webp,libdeflate). I don't know what it's tested on.

bradh:
> It's true. I don't know everything. zlib and deflate was previously added for png. Brotli? I guess only jpegxl uses brotli. libtiff I see that the entire file can be added to libheif. It has lot of compression algorithms. (lerc,zstd,webp,libdeflate). I don't know what it's tested on.

I think you're confused. We have the uncompressed codec in HEIF. That is defined in ISO/IEC 23001-17:2024.

There is an upcoming amendment to ISO/IEC 23001-17:2024 to allow (lossless) compression with zlib, deflate and brotli, intended for formats like floating point and signed integer, or high bit depth. So similar to what TIFF does.

silverbacknet:
> Only applies to png and jpg file decoder. Tiff isn't taken into account.

FWIW, a separate libtiff decoder was also submitted today, that seems to be exactly what you were looking for: #1205 It's a limited proof of concept that could be extended in the future, but it works for your basic 8-bit grey and 24/32-bit rgb/a color images for now, in any format libtiff can decode. I assume it'll be a few more iterations before it's merged into the mainline.

bradh:
Reworked and targets master on #1208

PR #1195, opened 2024-06-25
Update to SVT-AV1 v2.1.1

wantehchang:
Remove the no-dec build option because the decoder portion of SVT-AV1 was removed in v2.1.1.


farindk:
Thanks

PR #1198, opened 2024-06-26
uncompressed: support version 1 boxes

bradh:
This solves the same problem as https://github.com/strukturag/libheif/pull/1132, but aligned to the "the uncompressed codec looks for both variants and then internally generates the implied boxes. They are never visible outside the uncompressed codec."

It doesn't actually generate the boxes, it just works with and without the `cmpd` box.

### DIFF
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -2748,14 +2748,14 @@ static void set_default_options(heif_encoding_options& options)
   options.color_conversion_options.preferred_chroma_upsampling_algorithm = heif_chroma_upsampling_bilinear;
   options.color_conversion_options.only_use_preferred_chroma_algorithm = false;
 
-  options.prefer_minimised = false;
+  options.prefer_uncC_short_form = true;
 }
 
 static void copy_options(heif_encoding_options& options, const heif_encoding_options& input_options)
 {
   switch (input_options.version) {
     case 7:
-      options.prefer_minimised = input_options.prefer_minimised;
+      options.prefer_uncC_short_form = input_options.prefer_uncC_short_form;
       // fallthrough
     case 6:
       options.color_conversion_options = input_options.color_conversion_options;
--- a/libheif/heif.cc
+++ b/libheif/heif.cc
@@ -2734,7 +2734,7 @@ int heif_encoder_has_default(struct heif_encoder* encoder,
 
 static void set_default_options(heif_encoding_options& options)
 {
-  options.version = 6;
+  options.version = 7;
 
   options.save_alpha_channel = true;
   options.macOS_compatibility_workaround = false;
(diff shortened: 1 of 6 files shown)


libheif/heif.cc
```diff
@@ -2747,11 +2747,16 @@ static void set_default_options(heif_encoding_options& options)
   options.color_conversion_options.preferred_chroma_downsampling_algorithm = heif_chroma_downsampling_average;
   options.color_conversion_options.preferred_chroma_upsampling_algorithm = heif_chroma_upsampling_bilinear;
   options.color_conversion_options.only_use_preferred_chroma_algorithm = false;
+
+  options.prefer_minimised = false;
```
farindk, line 2751: Copy from #1132:


Is there any case where a user would want the longer form of the uncC box, if it is equivalent?
If there is no such case, we could drop the option.

If we decide to keep the option, it's better in my experience to have a very specific option (e.g. "force_uncC_long_form") and then have a function "set_force_long_forms_everywhere" that sets all those specific options. If we have a single catch-all option, it will get very messy if we want to enable it just for some box types, later on.

Moreover, I understand that the default 'false' is the backwards-compatible setting. However, if I see this correctly, nothing will break when we make 'true' the default and now write the short form box instead.

bradh: I think I meant to default to "make backwards compatible", but to have a switch function as you suggest.  Some consumers will be on old versions for a while.

My original thinking was that this flag would be generally applicable. So it could be used to enable minified boxes everywhere, given the concept would be to optimise for size over compatibility. I don't have a problem with multiple option flags that could be set seperately or together as needed.

Will add that.

bradh: I have implemented this. However the interaction of flags might be somewhat tricky if we flip the defaults.

At this point the default is false for both options, and the short form `uncC` will be created if either flag is set. If we change the default to be true, both would need to be false in order to write the long form.

farindk:
Thanks for the change.

Ready from your side?
I'll probably remove the `prefer_minimised_boxes_everywhere` option, because it doesn't work to have this in the options struct as they can contradict each other. When we have more similar options later on, we can add a
````
heif_encoder_options_set_minimised_boxes(bool);
````
To switch all those options at the same time.

I'd also prefer to set `prefer_uncC_short_form = true;`, because that is no breaking change for anyone and currently there are probably very few users of the codec anyway. It's better to have a nice default in the future.

bradh:
I think I've now fixed the option to match.

Do you want me to redo this against master?

farindk:
> Do you want me to redo this against master?

Not needed, I can merge it into master.

bradh:
> > Do you want me to redo this against master?
> 
> Not needed, I can merge it into master.

Thanks.

farindk:
Rebased and merged into master.
Thank you.

PR #1204, opened 2024-06-28
Limit total size of image instead of width and height individually.

fancycode:
This allows processing panorama images with could be more than 32k pixels wide but a lot less than 32k pixexls high (e.g. 40000 x 5000 pixels). This fits in the memory limit defined by the previous 32k x 32k pixels limit but would be rejected without this PR.


farindk:
Thank you, that makes sense.
But now we get an integer overflow somewhere:
````
file_fuzzer: /src/libheif/fuzzing/file_fuzzer.cc:41: void TestDecodeImage(struct heif_context *, const struct heif_image_handle *, size_t): Assertion `height >= 0' failed.
````

fancycode:
> Thank you, that makes sense. But now we get an integer overflow somewhere:
> 
> ```
> file_fuzzer: /src/libheif/fuzzing/file_fuzzer.cc:41: void TestDecodeImage(struct heif_context *, const struct heif_image_handle *, size_t): Assertion `height >= 0' failed.
> ```

Fixed (and added fuzzing input to corpus). The size check is now implemented in a central location so it doesn't need to be modified in lots of places if it needs to be changed in the future.

Problem was caused by a fuzzing input that had width `0` and height `4294967103`. This then overflowed the height which was cast to `int` in https://github.com/strukturag/libheif/blob/ddb541afdbe8dadfada4536077b8e4af1bb34b5e/libheif/context.cc#L610

farindk:
Thank you

PR #1207, opened 2024-06-28
uncompressed: add sanity limit check on number of tile rows and cols

### LINKED ISSUE #1206: Integer overflow in uncompressed_box.cc
@bradh I got a fuzzer hitting an integer overflow:

````
libheif/codecs/uncompressed_box.cc:263:38: runtime error: unsigned integer overflow: 4294967295 + 1 cannot be represented in type 'unsigned int'
````

https://github.com/strukturag/libheif/blob/9d4a5107c469c76fec7146b896819bc215a66fbe/libheif/codecs/uncompressed_box.cc#L263-265

Probably, it is enough to output an error when the number of tiles exceeds a sensible maximum.

bradh:
Resolves #1206 

Includes unit test.

This does not make the limit configurable, but that could be added.

Should we have the same limit for the number of image grid rows and columns?  That already has a short-form and long-form option, but no limits on the long form case:

https://github.com/strukturag/libheif/blob/7a8c58e4be8b3eb77d82330a407a7f5e6da15a44/libheif/context.cc#L189-L212

### DIFF
--- a/libheif/codecs/uncompressed_box.cc
+++ b/libheif/codecs/uncompressed_box.cc
@@ -27,6 +27,7 @@
 #include <cassert>
 
 #include "libheif/heif.h"
+#include "security_limits.h"
 #include "uncompressed.h"
 #include "uncompressed_box.h"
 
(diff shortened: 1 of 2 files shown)


farindk:
We might be able to omit that limit for grid images (total image size) because with the planned ROI decoding, it would be perfectly fine to have a huge image above the security limit if each tile is below the limit.

farindk:
Image Grid has a maximum of 256 tile rows/columns, stored in uint16.

bradh:
> Image Grid has a maximum of 256 tile rows/columns, stored in uint16.

Ah, indeed. Mixed up the image size and the tile limits.

farindk:
I suggest that we limit the total number of tiles instead of width/height separately, as was recently implemented in https://github.com/strukturag/libheif/pull/1204

bradh:
> I suggest that we limit the total number of tiles instead of width/height separately, as was recently implemented in #1204

Sounds reasonable.

> I.e. `MAX_NUMBER_OF_TILES` = 32768*32768;

Is this the limit value you want to use? Essentially allowing number of tiles to equal number of pixels?

farindk:
> > I.e. `MAX_NUMBER_OF_TILES` = 32768*32768;
> 
> Is this the limit value you want to use? Essentially allowing number of tiles to equal number of pixels?

I just used your `MAX_TILE_GRID_ROW_OR_COLUMN` value. The question is probably whether we need any tile limit at all. AFAIK, there is no memory overhead in uncompressed codec for tiles and at some point, the image size limit kicks in anyway.

Thus: do we need a tile limit?

bradh:
OK. So the plan is to limit number of uncompressed tile rows and tile columns to less than what can be represented in uint32_t (i.e. just avoiding the overflow in #1206), and not limiting grid. Then for really huge images, the user only has to increase the maximum image size limits, and we don't need to add API to change the tile size limits.

Does that seem reasonable?

farindk:
Yes. And for huge images, the user does not even have to increase the limit if he is accessing the image only in small ROIs that are below the limit. (Through the ROI decoding API that I'm going to add.)

bradh:
Have rebased. Should now match the intent above.

farindk:
Looks good. Thank you.

PR #1208, opened 2024-06-28
uncompressed: generic compression decode support

bradh:
This is from the draft Amendment 2 for ISO/IEC 23001-17. It re-uses and generalises the existing zlib support (from header compression), and adds an optional dependency on Brotli.

This version replaces #1194. 

Feedback on the approach and internal API changes would be appreciated.

I still have some work to do on this:

  - [x] decode row compression
  - [x] decode pixel compression (needs test data)
  - [x] icbr byte ranges (needs test data)
  - [x] some unit tests

There is no encoding support for this yet, and we'd likely need some kind of configuration option to specify that. It seems a bit premature given where the draft is up to.


farindk:
Looks good to me.

Let me know whether I should merge this now or wait until you have added the missing parts.

bradh:
Not yet. I think I'd like to handle the decompression earlier (in the same way we handle compressed metadata).

This approach was fine for the case where the whole file is compressed, but turns out to be hard for the `icbr` ranges case, since the codec isn't really working at the `mdat` extent range level.

bradh:
I think this is going to fail CIFuzz until https://github.com/google/oss-fuzz/pull/12165 is merged

libheif/CMakeLists.txt
```diff
@@ -133,12 +134,22 @@ else ()
     message("Not compiling 'libsharpyuv'")
 endif ()
 
-if (WITH_DEFLATE_HEADER_COMPRESSION)
+if (WITH_DEFLATE_HEADER_COMPRESSION OR WITH_UNCOMPRESSED_CODEC)
     find_package(ZLIB REQUIRED)
     target_link_libraries(heif PRIVATE ZLIB::ZLIB)
+    target_compile_definitions(heif PRIVATE WITH_ZLIB_COMPRESSION=1)
```
farindk, line 140: Could we rename this to `HAVE_ZLIB` to be consistent with `HAVE_BROTLI` ?
We could even make zlib optional like brotli.

libheif/codecs/uncompressed_box.cc
```diff
@@ -361,3 +361,91 @@ Error Box_uncC::write(StreamWriter& writer) const
 
   return Error::Ok;
 }
+
+
+Error Box_cmpC::parse(BitstreamRange& range)
+{
+  parse_full_box_header(range);
```
farindk, line 368: We could add a version check like here: https://github.com/strukturag/libheif/commit/2de879aadb52c7d2f661c0067faa6b853330a9f1

libheif/compression.h
```diff
@@ -17,17 +17,28 @@
...
 
 #include <vector>
 #include <cinttypes>
 #include <cstddef>
 
-#if WITH_DEFLATE_HEADER_COMPRESSION
+#include <error.h>
+
+#if WITH_ZLIB_COMPRESSION
 std::vector<uint8_t> deflate(const uint8_t* input, size_t size);
```
farindk, line 30: Maybe `deflate_zlib` for consistency.

libheif/compression.h
```diff
@@ -17,17 +17,28 @@
...
 
-std::vector<uint8_t> inflate(const std::vector<uint8_t>&);
+Error inflate_zlib(const std::vector<uint8_t>&compressed_input, std::vector<uint8_t> *output);
+
+Error inflate_deflate(const std::vector<uint8_t>&, std::vector<uint8_t> *output);
```
farindk, line 34: Funny name. What does this do?

libheif/compression.h
farindk, line 30: Maybe `deflate_zlib` for consistency.

libheif/compression.h
bradh: zlib can be run in "zlib" (RFC 1951) or "deflate" (RFC 1950) header modes. There doesn't appear to be a way to say "just deal with both" on decode (i.e. when inflating compressed data).

So we had deflate() and inflate(). I extended the API to handle the inflate() options for both deflate compression and zlib compression, based on the configured compression from the `cmpC` box. That got put into the name - its the inflate process for the deflate compression mode.

Possibly inflate_defl(), to match the 4CC code? The downside to that is that the compression / decompression code is really independent from the 23001-17 Amd 2 process. inflate_rfc1950() but that is a bit obscure too.

libheif/codecs/uncompressed_box.cc
```diff
@@ -361,3 +361,91 @@ Error Box_uncC::write(StreamWriter& writer) const
...
+
+  prepend_header(writer, box_start);
+
+  return Error::Ok;
+}
+
+
+Error Box_icbr::parse(BitstreamRange& range)
+{
+  parse_full_box_header(range);
```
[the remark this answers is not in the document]
bradh: Fixed.

libheif/codecs/uncompressed_box.cc
bradh: Fixed.

libheif/CMakeLists.txt
bradh: Fixed.

libheif/compression.h
farindk: Yes, I would also prefer `compress_deflate`, `decompress_deflate`, `decompress_brotli` ...

farindk: I'll do the renaming after the merge.

bradh: Thanks Dirk.

PR #1211, opened 2024-06-30
Other image pixel data-types

farindk:
@bradh
These are the changes to add new image data-types.
The concept for the public API is this:
- we have more (custom) `heif_channel`s, with the numbers `heif_channel_other_first` to `heif_channel_other_last`. If we later have specific channel names, we can also introduce `heif_channel` IDs for that, but I think it is difficult because, e.g., IR could be different kinds of IR.
- there is a new `heif_channel_datatype` enum to differentiate between integer, float, complex. The actual number of bits is stored in the existing bit-depth value. For complex, this is the total bit-depth, e.g. 128 for 2x64 bit floats. (I'm not completely sure yet whether this makes sense or it would be better to use the bit-depth per component. This is somehow different to how 'interleaved' is handled, but on the other hand, I can imaging that there are complex data-types like complex with 64bit magnitude and 16bit phase, thus a total bit-depth might make more sense).
- These special channels can be stored together with the visual data in an image. If we have an image that has no visual data, we set its colorspace to `heif_colorspace_nonvisual`.

This is WIP, comments welcome.

You should be able to add planes with these new datatypes to a `HeifPixelImage` now, but expect that not everything is working yet and public API is incomplete.
I'm probably also going to rename 'plane' to 'channel' to have a consistent naming scheme (see `heif_channel`).
Hence, the name to add a channel is now `HeifPixelImage::add_channel`, replacing the older `HeifPixelImage::add_plane`.


bradh:
> * we have more (custom) `heif_channel`s, with the numbers `heif_channel_other_first` to `heif_channel_other_last`.

I think this approach is OK. I would like to see a much bigger range though (a few thousand maybe), because hyperspectral data often has hundreds of bands. See https://www.usgs.gov/centers/eros/science/usgs-eros-archive-earth-observing-one-eo-1-hyperion for an older example that I'm planning to do as a sample (since we have NITF data for it), and there is commercial imagery claiming 500+ bands.

It would probably be useful to add a label (or maybe an identifier and a human readable label) for each band. I may have missed that in the API.

farindk:
Extended it to 3000 channels. Do you think that is future proof?

I do not completely foresee the consequences it has when we increase the maximum. AFAIK, the enum type is an integer type guaranteed to hold all values specified in it. I.e. if we increase the maximum, the integer type may change and break backwards compatibility. However, I think we should be safe until 32bit signed.

I've also thought about making the channel ID an `uint32_t`, but that would require changing/duplicating all API. I would like to avoid that at least for the moment.

> It would probably be useful to add a label (or maybe an identifier and a human readable label) for each band. I may have missed that in the API.

Definitely. I just omitted that until we have some input on how these could look like. In a previous draft of 23001-17, there was a small list of special IDs, but it has been removed, as far as I can see. It is probably also not the best approach. We could also adopt a naming scheme that is already in use (maybe you have some info about that). Or, we could define our own channel metadata. That would need some investigation about the requirements first. (Maybe the sensor sensitivity distribution over the frequency spectrum?)

PR #1219, opened 2024-07-04
CI: Enable more codecs when compiling on Windows.

fancycode:


fancycode:
@farindk this improves build coverage on Windows and fixes the always failing arm64 builds there due to the missing x265 in vcpkg.

farindk:
thank you

PR #1221, opened 2024-07-04
aom: Use all available CPUs for encoding by default.

### LINKED ISSUE #1212: Number of threads for AOM encoder limited
The AOM encoder currently defaults to 4 threads and limits to 16 threads:
https://github.com/strukturag/libheif/blob/ddb541afdbe8dadfada4536077b8e4af1bb34b5e/libheif/plugins/encoder_aom.cc#L237-L246

Is there a reason for exactly these values or could we default to "number of cpu threads" and limit to a higher value for systems with more than 16 cores?

CC @wantehchang

fancycode:
Fixes #1212

### DIFF
--- a/libheif/plugins/encoder_aom.cc
+++ b/libheif/plugins/encoder_aom.cc
@@ -26,6 +26,7 @@
 #include <cassert>
 #include <vector>
 #include <string>
+#include <thread>
 #include <memory>
 #include "encoder_aom.h"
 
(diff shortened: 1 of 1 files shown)


libheif/plugins/encoder_aom.cc
```diff
@@ -237,7 +238,12 @@ static void aom_init_parameters()
   p->version = 2;
   p->name = kParam_threads;
   p->type = heif_encoder_parameter_type_integer;
-  p->integer.default_value = 4;
+  unsigned int threads = std::thread::hardware_concurrency();
+  if (threads == 0) {
+    // Could not autodetect, use previous default value.
+    threads = 4;
+  }
+  p->integer.default_value = threads;
```
wantehchang, line 246: We should make sure `p->integer.default_value <= p->integer.maximum`. A computer may have more than 64 processor cores.

farindk: Thank you, good catch. I have corrected this.

wantehchang: Great! The new code looks good to me.

PR #1222, opened 2024-07-04
Fix filename in fuzzing corpus so it gets parsed during regression testing

fancycode:
Follow-up to #1204


fancycode:
See https://github.com/strukturag/libheif/commit/29bc4c32514b1fc720b29dc1b0ac6cea4f7bf2b5#commitcomment-143843141

libheif/context.cc
```diff
@@ -609,6 +609,17 @@ Error HeifContext::interpret_heif_file()
...
+          std::stringstream sstr;
+          sstr << "Image size " << width << "x" << height << " exceeds the maximum image size "
+                << m_maximum_image_size_limit << "\n";
+
+          return Error(heif_error_Memory_allocation_error,
+                        heif_suberror_Security_limit_exceeded,
+                        sstr.str());
+        }
+
         image->set_resolution(width, height);
```
farindk, line 623: Simpler fix would be to set the argument type for `set_resolution()` to `uint32_t`. It is `uint32_t` internally anyway.

fancycode: Ok, but it's returned as `int` in the getters and I didn't want to introduce any potential other errors by changing the type in various places.
https://github.com/strukturag/libheif/blob/e42b01f3b46eb02545b0ddeff4368d3dd3bdb430/libheif/context.h#L117-L119

Also it's returned as `int` in the public API:
https://github.com/strukturag/libheif/blob/e42b01f3b46eb02545b0ddeff4368d3dd3bdb430/libheif/api/libheif/heif.cc#L804
https://github.com/strukturag/libheif/blob/e42b01f3b46eb02545b0ddeff4368d3dd3bdb430/libheif/api/libheif/heif.cc#L815

farindk: Ok, makes sense.

PR #1227, opened 2024-07-07
Added patch to check that overlay's offsets are valid

flyyee:


libheif/pixelimage.cc
```diff
@@ -829,6 +829,26 @@ Error HeifPixelImage::overlay(std::shared_ptr<HeifPixelImage>& overlay, int dx,
                    "Overlay image outside of left or top canvas border");
     }
 
+    // verify that the destination points are within the bounds of the image's dimensions
+    if (out_x0 < 0 ||
+        out_x0 > out_w ||
+        out_y0 < 0 ||
+        out_y0 > out_h) {
```
farindk, line 836: I think all the checks against out_w, out_h, in_w, in_h should be >=.

farindk:
Thank you for the report and fix.

PR #1230, opened 2024-07-09
compression: add (internal) api documentation for compression

bradh:
This is a follow-up to the recent work on compression / decompression, adding some internal documentation on those.

It isn't going to show up in the Doxygen since its not public API. Depending on editor, it may show up as a hint. For example, in VS Code, it appears on hover:

image

### DIFF
--- a/libheif/compression.h
+++ b/libheif/compression.h
@@ -27,18 +27,68 @@
 #include <error.h>
 
 #if HAVE_ZLIB
+/**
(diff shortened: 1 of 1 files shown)


bradh:
See #1231 for a proposed fix on the fuzzer failure.

farindk:
> For example, in VS Code, it appears on hover:

Thanks. Also shows in CLion.

PR #1232, opened 2024-07-10
fix heif_dec no-colons option for windows

dnd29:
In Windows, file paths may contain colons, so replacing colons in file paths is not appropriate. Therefore, I propose modifying the code to only replace colons in auxType.

Proposed Changes
Modify the code to ensure that only colons in auxType are replaced, leaving file paths with colons intact.


farindk:
Thanks, makes sense.

PR #1236, opened 2024-07-10
Fix libbrotlidec pkg-config requires

kmilos:
See https://github.com/strukturag/libheif/commit/c0e4d242f157ed6386cf0bfcb85374cec80af837#r144071870


farindk:
Thanks

PR #1240, opened 2024-07-11
appveyor: Build multithreaded.

fancycode:


fancycode:
Doesn't make a difference as the machines in the `free` plan have one cpu:
https://github.com/appveyor/ci/issues/216#issuecomment-86497583 

But should be added anyway in case this changes in the future.

PR #1243, opened 2024-07-16
rename property raw data implementation to match header

bradh:
I saw link time failures with this function.

Looks like the rename only happened in the header.


farindk:
Right, thank you.

Issue #1244, opened 2024-07-16
heif_item_get_property_raw_data() should include the UUID for uuid boxes


bradh:
`heif_item_get_property_raw_data()` appears to include the flags / version, but not the extended type field (i.e. the UUID for the `uuid` box case).

Without the UUID, its hard to tell which UUID box is being returned.

bradh:
@farindk There are a couple of ways we could do this:

 - just prepend the UUID (i.e. the 16 bytes from the extended_type)
 - have a separate return value for the UUID
 - have a different method to get the UUID

I had originally planned to do the first. However it does mean we are mixing header and body at this low level.

Do you have a preference?

farindk:
The cleanest approach is probably to add an `heif_item_get_property_uuid_type()`. That leaves all the header parsing to the library.

The first option would, as you said, mix header and content. I see that we are also doing that with the `version` and `flags` from the `FullBox` header, but since we do not know that it is a `FullBox`, the library cannot parse the `FullBox` fields.

The approach from #1245 is also complicated for the user, because he first has to check with `heif_item_get_property_type()` whether it is an uuid box and depending on that interpret the content (e.g. the camera intrinsic box data would look different depending on whether it was stored as an uuid-box, as Apple is currently doing, or as a `cmin` box).

farindk:
I forgot to tell the main reason why we should have a separate function to extract the uuid type:
it would match how the other functions work. For example `heif_item_add_raw_property()` also has a separate uuid parameter. Thus, we could pass the data received by `heif_item_get_property_raw_data()` directly into that.

bradh:
> The approach from #1245 is also complicated for the user, because he first has to check with `heif_item_get_property_type()` whether it is an uuid box and depending on that interpret the content (e.g. the camera intrinsic box data would look different depending on whether it was stored as an uuid-box, as Apple is currently doing, or as a `cmin` box).

I think it is OK for the user that is looking for a specific box. One thing that would help is to be able to ask for the raw data for "uuid box with this UUID value". Right now, I ask for all of them and check the first 16 bytes.

https://github.com/OSGeo/gdal/compare/master...bradh:gdal:geoheif#diff-50def72c0349b456f1946c49ca32d3c9b7718be3390c194a338a1d7eae0188fbR728

However that would be less general.

bradh:
I implemented the separate function in #1246.

bradh:
Resolved as of #1246

PR #1245, opened 2024-07-17
include extended_type in raw data output

bradh:
This implements the first option of https://github.com/strukturag/libheif/issues/1244#issuecomment-2232051757


bradh:
Replaced by #1246

PR #1246, opened 2024-07-19
libheif: add API to get UUID for user extension boxes

bradh:
This is an alternative approach to solving #1244, adding a function to get the UUID (`extended_type` field).

Also includes documentation, a unit test, and some refactoring of shared code.

### DIFF
--- a/libheif/api/libheif/heif_properties.cc
+++ b/libheif/api/libheif/heif_properties.cc
@@ -334,16 +334,11 @@ struct heif_error heif_item_add_raw_property(const struct heif_context* context,
   return heif_error_success;
 }
 
-
-struct heif_error heif_item_get_property_raw_size(const struct heif_context* context,
-                                                  heif_item_id itemId,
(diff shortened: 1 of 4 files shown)


libheif/api/libheif/heif_properties.cc
```diff
@@ -409,3 +411,29 @@ struct heif_error heif_item_get_property_raw_data(const struct heif_context* con
 
   return heif_error_success;
 }
+
+struct heif_error heif_item_get_property_extended_type(const struct heif_context* context,
```
farindk, line 415: I know that it is called `extended_type` in the standard. But what do you think about calling it `heif_item_get_property_uuid_type`? I think that makes it clearer to the use who does not know the standard by heart that this belongs together to `uuid` boxes.

bradh: I can do that. I considered putting uuid in the name, so happy with the suggestion.

farindk:
Thank you.

PR #1248, opened 2024-07-20
examples: initial TIFF encoder support

bradh:
Inspired by the TIFF decoder (and a cold wet morning), I've implemented a very basic TIFF encoder.

There are a lot more things we can do with TIFF over PNG/JPEG/Y4M. For example, it could do multiple images per file (instead of splitting them), different compression types and levels, different chroma or interleave formats etc. It could also do the various sample types (e.g. signed, float, etc), which is part of the reason for wanting it. None of that is done yet, partly because it will complicate the Encoder API and command line options. 

I wonder if we should have only limited TIFF support for `heif-dec`, or whether to extend the `Encoder` API. Related, we could have HEIF as an output format (e.g. so you can pull in HEIC and output AVIF, or JPEG-2000, or uncompressed, or some other combination). Possibly that should be a different tool.

Thoughts?

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -129,8 +129,11 @@ endif ()
 
 find_package(TIFF)
 if (TIFF_FOUND)
+    target_link_libraries(heif-dec PRIVATE TIFF::TIFF)
     target_link_libraries(heif-enc PRIVATE TIFF::TIFF)
+    target_sources(heif-dec PRIVATE encoder_tiff.cc encoder_tiff.h)
     target_sources(heif-enc PRIVATE decoder_tiff.cc decoder_tiff.h)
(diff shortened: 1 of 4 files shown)


farindk:
> I wonder if we should have only limited TIFF support for heif-dec, or whether to extend the Encoder API.

One way could be to extend the encoder API to support multiple images per file and have an `Encoder_SingleImage` class as the base for the existing formats that will split this up into multiple images.

> Related, we could have HEIF as an output format (e.g. so you can pull in HEIC and output AVIF, or JPEG-2000, or uncompressed, or some other combination). Possibly that should be a different tool.

Sure. But at some point we might have to draw a line what is still the business of libheif and what should be handled by image convertion packages like ImageMagick. On the other hand, adding libheif as input or output format wouldn't be that difficult.

PR #1258, opened 2024-08-02
Add files from tests to fuzzing corpus to improve coverage.

fancycode:


fancycode:
@farindk not sure if the bpp check is fine or if values < 8 should be supported here.

farindk:
The problem is that we never considered <8 bits before and thus 8 bit color conversions use hard-coded constants at all kind of places. Simply replacing (`bitdepth != 8`) with (`bitdepth <= 8`) will lead to wrong conversions,  but `bitdepth != 8` will lead to misclassifying this as HDR. Both are wrong and we should instead disable the color-conversions for `bitdepth < 8` until this is supported.
I'll merge this and add the required tests.

PR #1261, opened 2024-08-08
api: properties build fix

bradh:
Looks like there was a merge problem, leading to this:

https://github.com/strukturag/libheif/actions/runs/10288783492/job/28475023513#step:5:393

This PR cleans that up.

### DIFF
--- a/libheif/api/libheif/heif_properties.cc
+++ b/libheif/api/libheif/heif_properties.cc
@@ -353,6 +353,12 @@ static struct heif_error find_property(const struct heif_context* context,
 
   auto box = properties[propertyId - 1];
   *box_other = std::dynamic_pointer_cast<Box_other>(box);
+
+  // TODO: every Box (not just Box_other) should have a get_raw_data() method.
+  if (*box_other == nullptr) {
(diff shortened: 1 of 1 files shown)


libheif/api/libheif/heif_properties.cc
```diff
@@ -353,6 +353,12 @@ static struct heif_error find_property(const struct heif_context* context,
 
   auto box = properties[propertyId - 1];
   *box_other = std::dynamic_pointer_cast<Box_other>(box);
+
+  // TODO: every Box (not just Box_other) should have a get_raw_data() method.
+  if (*box_other == nullptr) {
+    return {heif_error_Usage_error, heif_suberror_Invalid_property, "this property is not read as a raw box"};
+  }
+
```
farindk, line 361: I think we should not make `find_property()` specific only for `Box_other`.

farindk:
I'll cherry-pick 1ce3dced404d5768295542addbb530913fd2d9ff into the master branch, as this also fixes the merge error and make sure that `find_property` is not specific to `Box_other`.

Issue #1267, opened 2024-08-12
uncompressed codec gives wrong colors for any YCbCr and RGB input


farindk:
@bradh While working on #1242, I noticed that the uncompressed codec gives me wrong colors for basically all inputs I try. I tried YCbCr input from JPEGs and RGB from 8bit PNGs.

First, I thought this is because of my heavy refactoring, but I get the same output with v1.18.2.

I'm not attaching any test image here, because it fails with all color images I tried. If you need an image, let me know.

farindk:
RGB data without alpha seems to be ok. But PNGs with alpha do not work correctly.

bradh:
Is this on the encode side? Like `heif-enc --uncompressed something.png -o uncompressed.heif` ?

farindk:
Difficult to say whether it's on the encoder side or decoder side. I can only do an `heif-enc` / `heif-dec` loop and the above picture comes out as
![out](https://github.com/user-attachments/assets/a03cf161-be7c-4d93-93e6-ea815fe5777a)

````
$ ./heif-enc with-alpha-512x513.png --uncompressed -o out.heif
$ ./heif-dec out.heif -o out.png
````

bradh:
I can reproduce. The PNG + alpha problem looks like its on the decoder side, since gpac shows the encoded result correctly. The YCbCr problem looks like at least partly on the encoder side.

bradh:
Should be resolved once both those PRs are merged.

farindk:
Alpha output is still not working because `heif_image_handle_has_alpha_channel()` is not returning `true` for RGBA images. We need a way to determine this without actually decoding the image.

farindk:
Alpha output of uncompressed RGBA works now. Also `heif-info` shows that there is an alpha channel.

PR #1268, opened 2024-08-13
uncompressed: fix decoding of images with rgba profile

bradh:
This is the RGB with alpha part of #1267.

I also noted that the version 1 case wasn't displaying the bit depth correctly, which is resolved by this.

There is still an issue where heif-info shows "alpha channel: no", because its looking for an aux image, not at the actual image channels.

@farindk Can you verify that it works for your test cases?


farindk:
The colors  are correct now (red, blue swap), but the PNG output still has no alpha.
The reason for the missing alpha seems to be that `heif_image_handle_has_alpha_channel()` is not returning `true` for RGBA images. We will need a way to determine this without actually decoding the image.

PR #1269, opened 2024-08-13
uncompressed: fix encoding of YCbCr with subsampling

bradh:
This is the YCbCr part of #1267.

I'm actually surprised it encoded at all (rather than segfaulting), however now valgrind clean.

@farindk Can you confirm this works for your test cases?


farindk:
Thank you, that works fine now.

PR #1270, opened 2024-08-13
uncompressed: update generic compression to current DAM 2

bradh:
The big change is that offsets are now relative to the start of the data, not the start of the file. That requires new test data.

The box name also changes.


farindk:
Thanks.

If you have larger changes in the coming days, can you please develop them against the `file-layout` branch. I did a major refactorization there and it is difficult to merge changes into it.

I cannot merge `file-layout` into `master` yet, because it still lacks full functionality.

libheif/codecs/uncompressed_box.cc
```diff
@@ -394,68 +391,164 @@ Error Box_cmpC::write(StreamWriter& writer) const
...
+  int unit_offset_nbbits = 16;
+  int unit_size_nbbits = 8;
+  uint64_t implied_offset = 0;
+  bool can_use_implied_offsets = true;
+  for (CompressedUnitInfo unit_info: m_unit_infos) {
+    if (unit_info.unit_offset != implied_offset) {
+      can_use_implied_offsets = false;
+    }
+    implied_offset += unit_info.unit_size;
+    while (unit_info.unit_offset >= (1ULL << unit_offset_nbbits)) {
```
farindk, line 476: clang-tidy says:
````
uncompressed_box.cc:476:43: error: The result of the left shift is undefined due to shifting by '64', which is greater or equal to the width of type 'unsigned long long' [clang-analyzer-core.UndefinedBinaryOperatorResult,-warnings-as-errors]
    while (unit_info.unit_offset >= (1ULL << unit_offset_nbbits))
````

I think this is a false positive, but you may want to check too.

farindk: Here is the link to the logical reasoning:
https://github.com/strukturag/libheif/actions/runs/10369014420/job/28703823871#step:6:521

farindk: Now I see why this is wrong. The computation of `unit_offset_nbbits` is in a for-loop and leads to the error when it has reached 64bit in a previous iteration.
Please check the computation. Maybe it has to be outside the for-loop.

bradh: I have an idea for a cleaner approach.  Will update.

PR #1271, opened 2024-08-14
uncompressed: simplify compressed units handling

bradh:
This avoids the potential for undefined behaviour, as well as making the implementation a bit simpler.


farindk:
Thank you.

PR #1278, opened 2024-08-21
ci: remove coverity from PR

bradh:
This fails for most users, and is confusing for occasional contributors.


bradh:
The fuzzer failure is legitimate, but unrelated to this change. I'll follow up with a new PR for that.

PR #1282, opened 2024-08-24
encoder_aom.cc: fix memory leaks

kleisauke:
Discovered via fuzz testing.

Resolves: https://github.com/libvips/libvips/issues/4109.


farindk:
Thank you. I have rewritten with a `unique_ptr` with custom destructor such that the image is freed in any case and it is not forgotten when the cost changes in the future.

PR #1286, opened 2024-08-26
tests: enable J2K_HT tests when OpenJPH is found

bradh:


farindk:
The AppVeyor error is unrelated. I have corrected that.

PR #1289, opened 2024-08-27
[ffmpeg plugin] Fix leak during failure

roticv:
When decoding "fails" in ffmpeg decoder, it is possible that NalMap is not cleaned up properly causing a leak.


farindk:
Thank you. I have fixed this in a different way by using a `std::unique_ptr`. That makes sure that the memory is freed in any edge case and also simplifies existing code.

PR #1293, opened 2024-09-01
initial version of NVIDIA hw plugin

bradh:
This is an initial cut of the changes to provide the compression format down to the decoder.

I'm confused by the AVC NAL Unit syntax. I've factored the NALU parsing out of ffmpeg, and it works fine for HEVC. The AVC syntax is different though. Inputs welcome.

I don't know if this code works on AV1 - I don't have compatible hardware for that yet. It does work for JPEG and HEVC. It definitely doesn't work for AVC without that NALU parsing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,6 +229,14 @@ if (WITH_OPENJPH_ENCODER OR WITH_OPENJPH_DECODER)
     find_package(OPENJPH)
 endif()
 
+# nvidia hardware decode
+
+plugin_option(NV_DECODER "NVIDIA Hardware HEVC/AVC/AV1 decoder" OFF OFF)
+if (WITH_NV_DECODER)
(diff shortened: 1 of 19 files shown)


farindk:
My only computer with an nVidia card is an old laptop that only supports H.264. But if there is any example image using H.264, I could have a look at the NALU parsing.

bradh:
I'm currently refactoring the NALU parsing (since it changed a bit in ffmpeg). Hope to put that up in hour or so.

https://github.com/MPEGGroup/FileFormatConformance/blob/main/data/file_features/published/heif/multilayer004.heic is the sample I'm using for H.264. It isn't going to be great quality without the LCEVC support, but should be valid for format checks.

farindk:
I've written an AVC decoding plugin based on OpenH264. At least `multilayer004.heic` decodes.

PR #1294, opened 2024-09-01
plugins: refactor NALU parsing for shared future usage

bradh:
Potentially there is additional code sharing possible.

Also, may be useful to have an abstract NalUnit superclass with HevcNalUnit and AvcNalUnit concrete subclasses.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,20 +163,22 @@ endif ()
 
 # openh264 decoder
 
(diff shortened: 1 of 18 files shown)


bradh:
Well that build didn't go well...

farindk:
I've cherry-picked your commits into `master` because the PR branch had a lot of double commits.

PR #1301, opened 2024-09-08
Refactor TIFF decoding

### LINKED ISSUE #1300: reading TIFF mono files fails
`heif-enc` with a single band file fails with an error like:

```
Could not encode HEIF/AVIF file: Encoder plugin generated an error: Unsupported bit depth: Bit depth not supported by x265
```

The error is misleading though. Its really an invalid pixel image being passed as input. That happens because of this logic:

https://github.com/strukturag/libheif/blob/8cc6c796ac706bdf99c1f6f82f32b7873bbbf19d/examples/decoder_tiff.cc#L307-L321

(and equivalent call for the planar case).

Essentially we're creating a mono image, then adding an interleaved image plane, which doesn't make sense.

I'm currently working on a fix.

bradh:
This is an initial rework of the TIFF decoder support, in preparation for reading other data types and higher bit depths.

Because this code is starting to get tricky, I really wanted unit tests. Also, this is looking like code that would be useful in general - not just for the specific examples. In support of that, I've made it into a library - notionally `libheifio`, but I'm flexible on the name. It doesn't get provided as API yet. It has error handling, not just `exit()` calls.

Resolves #1300 

Also resolves a failure reading RGBA planar  files. 
```
./examples/heif-enc ~/rgba_planar.tif out.heif
munmap_chunk(): invalid pointer
Aborted (core dumped)
```


bradh:
It looks like the library may not be static. That needs to be fixed.

farindk:
Is this still a draft or ready for merging?

bradh:
@farindk if you are OK with the design then I am OK with this as an initial implementation that we can build onto.

farindk:
> @farindk if you are OK with the design then I am OK with this as an initial implementation that we can build onto.

Yes, I can merge this and then I would reorganize the files a bit: move the `heifio` directory next to the `libheif` directory and move the tests into a subdirectory of the top-level `tests` directory. Hope that is ok.

farindk:
Merged with rebase.

PR #1302, opened 2024-09-08
Windows fixes

bradh:
The unistd change is untested, but should at least compile.

The changes for the utilities are `const char*` incompatibility with `char *`.

The C++ 20 is driven by `use of designated initializers requires at least '/std:c++20'` which we use in structure initialisation in a few places. 

The fallthrough will require at least C++ 17, even if we change the initialisers.


bradh:
Not sure what I've done that broke the AVC box checks....

farindk:
@fancycode Can we increase the C++ version to C++20 ? Ubuntu 20.04 uses gcc-9, which does not support full C++20 (https://gcc.gnu.org/projects/cxx-status.html), but it does support designated initializers. The safer choice would be C++17.

farindk:
"Designated initializers" are only used in one place in PR #1296 . If the step to C++20 is not completely safe for older distributions, we could also use C++17 as an intermediate step.

farindk:
> Cygwin will probably also not ship GCC 13 w/ full C++20 support well until next summer ([it was just bumped to GCC 12](https://cygwin.com/pipermail/cygwin-announce/2024-July/011859.html)).

According to https://gcc.gnu.org/projects/cxx-status.html, there is only one C++20 feature that requires GCC 13 (Atomic Compare-and-Exchange with Padding Bits). I don't think that this is a blocker.

farindk:
> > there is only one C++20 feature that requires GCC 13
> 
> A few more unfortunately, but maybe also irrelevant for libheif code:

Yes seems irrelevant. That's just calendar and text formatting in the library.
For the compiler, even gcc-11 should be ok.

farindk:
I've merged most of it by cherry-picking.

The `(char*)` casts have been removed in favor of fixing the Windows `getopt*` implementations.

I've fixed the integer check in the ftyp parsing.

Let's try to set the language standard to C++20. This will also give us `std::cmp_greater` and similar, which should help in reducing the number of integer casts in the safety checks during parsing.

bradh:
Verified it builds ok on windows, at least for the release build.

We may need to update catch to a get the development build working, but I am OK for now.

PR #1321, opened 2024-10-01
extract stream decoding to Decoder classes

farindk:
Move actual bitstream decoding into `Decoder` classes that can decode data irrespective where the data is located in the file.


CMakeLists.txt
```diff
@@ -243,7 +243,7 @@ endif()
 plugin_option(OpenJPEG_ENCODER "OpenJPEG J2K encoder" OFF ON)
 plugin_option(OpenJPEG_DECODER "OpenJPEG J2K decoder" OFF ON)
 if (WITH_OpenJPEG_ENCODER OR WITH_OpenJPEG_DECODER)
-    find_package(OpenJPEG 2)
+    find_package(OpenJPEG)
```
bradh, line 246: I know why you do this (I do too), but probably not wanted in this PR.

farindk: I think I'll have to put that on `.gitignore`. No matter how much care I take, sooner or later it slips through :-)

PR #1324, opened 2024-10-07
fix memory leak when get_compressed_data fails

alex:


bradh:
Looks OK, but maybe a smart pointer with destructor would be cleaner with these multiple paths?

farindk:
Thanks. As @bradh noted, I'll change this to a smart pointer.

PR #1329, opened 2024-10-10
Security: restore previous default limits prior to commit 2374ade

lovell:
Commit https://github.com/strukturag/libheif/commit/2374ade0a24f107d5de8fb6bd5f9ceef04b6c510 allows security limits to be set per context but also removed/altered a couple of the original limits. Removing the `max_iloc_items` limit opens libheif to a denial of service attack. This PR restores the previous defaults.


lovell:
Thanks Dirk, I agree the defaults can be improved, this PR is attempting only to restore the previous defaults to reduce the noise levels we've been experiencing from libvips fuzz tests over the last couple of days.

PR #1330, opened 2024-10-10
aom encoder: expose auto-tiles boolean option (since v3.10.0)

lovell:
Setting this option allows libaom to determine the optimum tile count for the given output dimensions and available threads.

This typically reduces encoding (wall clock) time and/or peak RAM consumption at the cost of slightly increased file sizes. I see a ~30-50% speed up on many-core systems.

```
$ /usr/bin/time -v heif-enc in.jpg -o out.avif
	User time (seconds): 1.71
	System time (seconds): 0.03
	Percent of CPU this job got: 258%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.67
	Maximum resident set size (kbytes): 50432
```
```
$ /usr/bin/time -v heif-enc -p auto-tiles=true in.jpg -o out.tiled.avif
	User time (seconds): 2.23
	System time (seconds): 0.04
	Percent of CPU this job got: 588%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.38
	Maximum resident set size (kbytes): 52864
```
```
 847374 in.jpg
 117500 out.avif
 118148 out.tiled.avif
```

### DIFF
--- a/libheif/plugins/encoder_aom.cc
+++ b/libheif/plugins/encoder_aom.cc
@@ -71,6 +71,7 @@ struct encoder_struct_aom
   int threads;
   bool lossless;
   bool lossless_alpha;
+  bool auto_tiles;
 
 #if defined(HAVE_AOM_CODEC_SET_OPTION)
   std::vector<custom_option> custom_options;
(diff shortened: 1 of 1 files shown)


farindk:
Very nice.

Note that I am currently finalizing writing/reading individual tiles:
https://github.com/strukturag/libheif/wiki/Read-and-Writing-Tiled-Images

This might also be interesting for libvips. Not just because of the parallelization possibilities, but also because of the support for much larger images.

PR #1339, opened 2024-10-15
Timestamp properties - git merge master

dukesook:
Keep the timestamp-properties branch up to date with master.

The bitstream.h started to diverge.


farindk:
What is the current status of the standard? Is it stable?

I could move the timestamp API into the experimental header so that it is in the main branch, but not enabled by default until we have the final standard.

dukesook:
Yes, I think using the experimental header for now is a good idea..

farindk:
I've merged it into master with the TAI function in heif_experimental.

Be aware that I also renamed `WITH_EXPERIMENTAL_FEATURES` to `ENABLE_EXPERIMENTAL_FEATURES` for consistency.

PR #1341, opened 2024-10-16
examples: speculative build fix for AppVeyor

bradh:
Recent builds have failed, example https://ci.appveyor.com/project/strukturag/libheif/builds/50803252/job/9ra01vv6lfwrtjsj

This is a potential (possibly partial) fix.

### DIFF
--- a/examples/heif_enc.cc
+++ b/examples/heif_enc.cc
@@ -37,6 +37,8 @@
 #include <vector>
 #include <string>
 #include <sstream>
+#include <optional>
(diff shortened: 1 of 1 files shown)


bradh:
Partial fix.

bradh:
> The main issue for windows is that the functions to scan the directory do not exist in Windows. Maybe we can replace it with `std::filesystem`.

Looks like best option since we are already  up to C++ 20.

PR #1343, opened 2024-10-17
start on mini box alpha support

bradh:
This is a first cut of the mini box support.

It is enough to support what the (experimental) libavif generates, except for HDR gainmaps, which we don't do yet.

Although it is not yet complete, I'd like to get this part in now (behind a CMake switch) because its hitting merge conflicts.


libheif/api/libheif/heif.h
```diff
@@ -251,6 +251,9 @@ enum heif_suberror_code
 
   heif_suberror_No_avcC_box = 143,
 
+  // we got a mini box, but could not read it properly
+  heif_suberror_No_mini_box = 149,
```
farindk, line 255: Maybe we should rename the error to something else so that it fits the semantics?
Something like `heif_suberror_Invalid_mini_box` ?

libheif/api/libheif/heif.h
```diff
@@ -1966,6 +1980,13 @@ struct heif_decoded_mastering_display_colour_volume
   double min_display_mastering_luminance;
 };
 
+struct heif_ambient_viewing_environment
+{
+  uint32_t ambient_illumination;
+  uint16_t ambient_light_x;
+  uint16_t ambient_light_y;
+};
+
```
farindk, line 1989: Is this struct finalized in the standard? I did not find its definition. If it's not final yet, we should move it to `heif_experimental.h`.

libheif/bitstream.h
```diff
@@ -411,10 +422,21 @@ class BitReader
...
+  void set_start_offset(uint64_t offset)
+  {
+    start_offset = offset;
+  }
+
+  uint64_t get_file_offset() const
+  {
+    return start_offset + (data_length - bytes_remaining - (nextbits_cnt / 8));
+  }
+
```
farindk, line 434: Could you please remove the `set_start_offset()` function from here and simply add the offset in the `mini`-parser? It seems unnatural to me to move this offset variable into the `BitReader` class.

libheif/mini.h
```diff
@@ -0,0 +1,161 @@
...
+#include "box.h"
+
+#include <memory>
+#include <string>
+#include <vector>
+
+class Box_mini : public Box
+{
+public:
+	Box_mini()
```
farindk, line 34: This file and `mini.cc` use a much larger indentation than usual. Can we normalize that to 2 spaces?

farindk:
Thanks for this PR. Looks good, I just have a few minor comments above.

libheif/api/libheif/heif.h
bradh: The property is in ISO/IEC 23008-12:2022 Section 6.5.36 "Ambient viewing Environment". That just points back to ISO/IEC 14496-12:2022, Section 12.1.9 "Ambient viewing environment" for the syntax and semantics.

ISO/IEC 14496-12:2022 has the syntax, but references ISO/IEC 23008-2 for the semantics. The semantics are in ISO/IEC 23008-2:2023 Section D.3.39.

farindk: Thanks for guiding me through this jungle of documents.

Since the `ambient_viewing_environment` and `content_colour_volume` are defined in H.265 without any versioning information, I think we can dare to use it in the API without a version number. If it would be an HEIF property only, I would have proposed to use a version number in the struct.

libheif/api/libheif/heif.h
bradh: Fixed.

libheif/mini.h
bradh: Fixed

farindk:
Thanks

PR #1344, opened 2024-10-17
heif-enc: exit with invalid x265 internal parameters

chemag:
Tested:
```
$ build/examples/heif-enc lena.490x490.png -p x265:colorprim=12 -p x265:foo=bar -o /tmp/foo.heic
Could not encode HEIF/AVIF file: Usage error: Unsupported parameter: Unsupported internal encoder parameter: x265:foo
```

### DIFF
--- a/libheif/plugins/encoder_x265.cc
+++ b/libheif/plugins/encoder_x265.cc
@@ -859,7 +859,17 @@ static struct heif_error x265_encode_image(void* encoder_raw, const struct heif_
     }
     else if (strncmp(p.name.c_str(), "x265:", 5) == 0) {
       std::string x265p = p.name.substr(5);
-      api->param_parse(param, x265p.c_str(), p.value_string.c_str());
+      if (api->param_parse(param, x265p.c_str(), p.value_string.c_str()) < 0) {
+        char error_message[1024];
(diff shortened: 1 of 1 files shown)


libheif/plugins/encoder_x265.cc
```diff
@@ -859,7 +859,17 @@ static struct heif_error x265_encode_image(void* encoder_raw, const struct heif_
...
       std::string x265p = p.name.substr(5);
-      api->param_parse(param, x265p.c_str(), p.value_string.c_str());
+      if (api->param_parse(param, x265p.c_str(), p.value_string.c_str()) < 0) {
+        char error_message[1024];
+        strcpy(error_message, "Unsupported internal encoder parameter: ");
+        strcat(error_message, p.name.c_str());
+        struct heif_error err = {
+          .code = heif_error_Usage_error,
+          .subcode = heif_suberror_Unsupported_parameter,
+          .message = error_message
```
farindk, line 869: We cannot return a pointer to a local char array because it will be destroyed when the function returns. The array has to be stored in the plugin data structure.

I'll change it...

farindk:
Thanks for the contribution.

I've corrected the string memory issue.

PR #1347, opened 2024-10-20
test case for uncompressed decoder advertisement

bradh:
This shows that we are not advertising decoder support for uncompressed.

The base reason is that it checks for the compression type in the list of plugins, and there is no plugin for uncompressed decode. There is one for uncompressed encode.

This doesn't show up in the heif-dec example code, because it has a hard-coded response depending on compile time option for uncompressed.

I can see two ways forward:
 - add a dummy decoder plugin
 - add a special case (depending on compile time option for uncompressed) to `heif_have_decoder_for_format()`.

@farindk your preference?

### DIFF
--- a/tests/uncompressed_decode.cc
+++ b/tests/uncompressed_decode.cc
@@ -165,6 +165,6 @@ TEST_CASE("check image handle no metadata blocks") {
 }
 
 TEST_CASE("check uncompressed is advertised") {
-  REQUIRE(heif_have_decoder_for_format(heif_compression_uncompressed) == true);
+  REQUIRE(heif_have_decoder_for_format(heif_compression_uncompressed) == (int)true);
 }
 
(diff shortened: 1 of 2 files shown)


bradh:
I added the special case for this proposal. Just seemed simpler and easier, not necessarily better.

tests/uncompressed_decode.cc
```diff
@@ -165,6 +165,6 @@ TEST_CASE("check image handle no metadata blocks") {
 }
 
 TEST_CASE("check uncompressed is advertised") {
-  REQUIRE(heif_have_decoder_for_format(heif_compression_uncompressed) == true);
+  REQUIRE(heif_have_decoder_for_format(heif_compression_uncompressed) == (int)true);
```
farindk, line 168: Minor nitpicking: I think we should not check against a specific `true` value, because anything `!= 0`is true.

farindk:
I've changed it to a dummy decoder plugin, because there are other API functions that did not list the uncompressed decoder (e.g. `heif_get_decoder_descriptors()`) and we also need an ID for it to be used in `heif_decoding_options::decoder_id`.

PR #1351, opened 2024-10-29
mini: implement HEVC reading

bradh:
Adds support for HEVC to the existing AV1 mini box decoding.

This also completes most of the ItemPropertyBox reconstruction (less HDR since we don't support it yet). The work on `pixi` box is also outstanding. As part of the reconstruction, there are properties that may or may not be used. The CDAM describes using a FreeSpaceBox, which seems like an easier way to get the associations right.


libheif/box.h
```diff
@@ -412,6 +412,23 @@ class Box_ftyp : public Box
 };
 
 
+class Box_free : public Box
```
farindk, line 415: Would it be possible to reuse the existing `Box_other` as the placeholder box instead of introducing a new `Box_free` class?

bradh: That was my first attempt. 

It didn't work out because some of the properties are marked essential (e.g. imir and irot, plus the codec might end up getting omitted if we support the JPEG case). We have a check for any `Box_other` instances that are marked as essential, since its something we should handle and can't. So a `Box_free` (which is a no-op) seemed better.

The other way to do this would be to use `Box_other` but to check each association against the properties first, and only make it if we need it.

Or we could do both the `free` (either as an explicit `Box_free` box, or as a `Box_other` with `free` parameter), and the check, which would exactly align with the CDAM description.

farindk:
I've moved the expansion of the boxes into the `Box_mini` class to keep `HeifFile` clean. Please do a double check that it still works.

PR #1352, opened 2024-10-29
evc: initial work for EVC

bradh:
Initial work on EVC. No codec yet, just the basic box.


bradh:
I've redone the merge, adopting the new define. Looks like I need to do a bit more work though.

farindk:
Why did you close this PR? It is very good and just was not merged yet because we lack the EVC codec implementation.
If you don't mind, I'd like to keep this open.

bradh:
I had hoped to use the xeve and xevd implementations for the codec, but they don't look like they are ready for production use.  So I've moved on, and didn't really see how this was going to get to a usable state. 

I am OK with re-opening though, in case someone else wants to continue the work.

farindk:
Right, the last commit to xeve/xevd was ten months ago and there are no alternative implementations. Not sure if that will ever take off. But with this PR, we are at least prepared, just in case.

torusrxxx:
xeve/xevd has been officially supported by FFmpeg, should we also add support for EVC via FFmpeg?

bradh:
Possibly we don't see FFmpeg in the same light :-)

I think xevd is getting some work, but I don't see it as a production-ready capability at this point.

farindk:
I haven't seen anyone use EVC yet, anywhere.
Royalty-free "baseline" profile is not much better than H.264, but with a much worse ecosystem. "main" profile isn't royalty free. So why not directly use AV1 instead?
If it's only dead code that has to be maintained, it is not worth it.

torusrxxx:
I mean, we will only support it indirectly via the FFmpeg plugin for now, the maintenance burden is reduced to just support for EVC boxes. This is `libheif`, if even us do not support it, who else? There's no second choice.

farindk:
> This is libheif, if even us do not support it, who else?

No user ever asked for it. Thus, there seems to be no need for it.
What is your use-case?

torusrxxx:
I have zero use case for it, I think it's a problem. Probably I will cherry-pick some commits here into a new branch to make it usable.

Issue #1362, opened 2024-11-01
1.19.0: test suite fails in 3 units


kloczek:
cmake settings:
```console
[tkloczko@pers-jacek x86_64-redhat-linux-gnu]$ cmake -L
CMake Warning:
  No source or binary directory provided.  Both will be assumed to be the
  same as the current working directory, but note that this warning will
  become a fatal error in future CMake releases.


CMake Error: The source directory "/home/tkloczko/rpmbuild/BUILD/libheif-1.19.0/x86_64-redhat-linux-gnu" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
-- Cache values
BROTLI_COMMON_LIB:FILEPATH=BROTLI_COMMON_LIB-NOTFOUND
BROTLI_DEC_INCLUDE_DIR:PATH=BROTLI_DEC_INCLUDE_DIR-NOTFOUND
BROTLI_DEC_LIB:FILEPATH=BROTLI_DEC_LIB-NOTFOUND
BROTLI_ENC_INCLUDE_DIR:PATH=BROTLI_ENC_INCLUDE_DIR-NOTFOUND
BROTLI_ENC_LIB:FILEPATH=BROTLI_ENC_LIB-NOTFOUND
BUILD_SHARED_LIBS:BOOL=ON
BUILD_TESTING:BOOL=ON
CMAKE_BUILD_TYPE:STRING=RelWithDebInfo
CMAKE_COMPILE_WARNING_AS_ERROR:BOOL=ON
CMAKE_INSTALL_PREFIX:PATH=/usr
ENABLE_COVERAGE:BOOL=OFF
ENABLE_EXPERIMENTAL_FEATURES:BOOL=OFF
ENABLE_EXPERIMENTAL_MINI_FORMAT:BOOL=OFF
ENABLE_MULTITHREADING_SUPPORT:BOOL=ON
ENABLE_PARALLEL_TILE_DECODING:BOOL=ON
ENABLE_PLUGIN_LOADING:BOOL=ON
FUZZING_COMPILE_OPTIONS:STRING=-fsanitize=fuzzer,address,shift,integer -fno-sanitize-recover=shift,integer
FUZZING_CXX_COMPILER:STRING=clang++
FUZZING_C_COMPILER:STRING=clang
FUZZING_LINKER_OPTIONS:STRING=
PLUGIN_DIRECTORY:STRING=/usr/lib64/libheif
[...]

PR #1363, opened 2024-11-02
Update CMakeLists.txt

arjendekorte:
The extended_type test will always fail when build with WITH_UNCOMPRESSED_CODEC=ON as HEVC will not be available then.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -47,8 +47,11 @@ endif()
 # --- tests that only access the public API
 
 add_libheif_test(encode)
-add_libheif_test(extended_type)
 add_libheif_test(region)
+
+if (NOT WITH_UNCOMPRESSED_CODEC)
+    add_libheif_test(extended_type)
+endif()
 
 if (WITH_OPENJPH_ENCODER AND SUPPORTS_J2K_HT_ENCODING)
     add_libheif_test(encode_htj2k)


farindk:
The uncompressed codec has nothing to do with the HEVC codec.
The real reason is probably that you compiled with plugins and the test could not find the plugin, similar to what was observed here: #1362

arjendekorte:
It indeed has nothing to do with `WITH_UNCOMPRESSED_CODEC=ON`, but it is a different problem as mentioned in #1362. The problem is the HEVC codec is not available here:
```
[    3s] Preset CMake variables:
[    3s] 
[    3s]   BUILD_TESTING="ON"
[    3s]   CMAKE_COMPILE_WARNING_AS_ERROR="ON"
[    3s]   ENABLE_EXPERIMENTAL_FEATURES="ON"
[    3s]   ENABLE_PLUGIN_LOADING="OFF"
[    3s]   WITH_REDUCED_VISIBILITY="OFF"
[    3s]   WITH_UNCOMPRESSED_CODEC="ON"

[...]

[    3s] === Summary of compiled codecs ===
[    3s] libde265 HEVC decoder           : - not found
[    3s] FFMPEG HEVC decoder (HW acc)    : - disabled
[    3s] x265 HEVC encoder               : - not found
[    3s] Kvazaar HEVC encoder            : - disabled
[    3s] AOM AV1 decoder                 : + built-in
[    3s] AOM AV1 encoder                 : + built-in
[    3s] Dav1d AV1 decoder               : - disabled
[    3s] SVT AV1 encoder                 : - disabled
[    3s] Rav1e AV1 encoder               : - disabled
[    3s] JPEG decoder                    : - disabled
[    3s] JPEG encoder                    : - disabled
[    3s] OpenH264 decoder                : - not found
[    3s] OpenJPEG J2K decoder            : - disabled
[    3s] OpenJPEG J2K encoder            : - disabled
[    3s] OpenJPH HT-J2K encoder          : - disabled
[    3s] uvg266 VVC enc. (experimental)  : - disabled
[    3s] vvenc VVC enc. (experimental)   : - disabled
[    3s] vvdec VVC dec. (experimental)   : - disabled
```
This causes the extended_type test to fail at line 44.

arjendekorte:
I guess the way to do this, is by running this test only if the HEVC is available
```
diff -purN a/tests/CMakeLists.txt b/tests/CMakeLists.txt
--- a/tests/CMakeLists.txt      2024-11-01 11:56:56.000000000 +0100
+++ b/tests/CMakeLists.txt      2024-11-03 22:35:11.344681868 +0100
@@ -47,9 +47,12 @@ endif()
 # --- tests that only access the public API
 
 add_libheif_test(encode)
-add_libheif_test(extended_type)
 add_libheif_test(region)
 
+if ((WITH_X265 OR WITH_KVAZAAR) AND SUPPORTS_HEIC_ENCODING)
+    add_libheif_test(extended_type)
+endif()
+
 if (WITH_OPENJPH_ENCODER AND SUPPORTS_J2K_HT_ENCODING)
     add_libheif_test(encode_htj2k)
 else()
```

farindk:
Yes, that makes sense. I'll think of a change in the spirit of your proposal. (It is a bit more difficult because there are tests that require HEVC and AV1, or some tests would run with any codec. This makes things complex.)

farindk:
At the above commit, these cases should now be reported as "skipped tests".

PR #1369, opened 2024-11-05
tests: update JPEG 2000 and HT tests to Catch2

bradh:


farindk:
FYI: I'll have to revert back to the amalgamated catch2 code because downloading the latest version during building does not work for the packaging people...

bradh:
> FYI: I'll have to revert back to the amalgamated catch2 code because downloading the latest version during building does not work for the packaging people...

Can we just require the package instead?

farindk:
Ubuntu 22.04 is currently at catch2 v2.13.8 (3 years behind).

bradh:
> Ubuntu 22.04 is currently at catch2 v2.13.8 (3 years behind).

https://launchpad.net/~fenics-packages/+archive/ubuntu/fenics will get you back to 22.04.

So in the CMake logic, check if its already available, and only download if not.

Then in CI, apply that PPA (or another, like https://launchpad.net/~savoury1/+archive/ubuntu/build-tools), install the package, and it should be found at build time (hence no download).

fancycode:
> > Ubuntu 22.04 is currently at catch2 v2.13.8 (3 years behind).
> 
> https://launchpad.net/~fenics-packages/+archive/ubuntu/fenics will get you back to 22.04.
> 
> So in the CMake logic, check if its already available, and only download if not.
> 
> Then in CI, apply that PPA (or another, like https://launchpad.net/~savoury1/+archive/ubuntu/build-tools), install the package, and it should be found at build time (hence no download).

The problem is in packaging where you only have access to official repositories, i.e. no PPAs, downloads or git checkouts.

farindk:
This issue is because of packaging, not the CI.
As I understand it, when cmake runs, it's already too late to decide whether a current version is available.
It has to be specified as a build-dependency and if that is not available, the code has to be included. (see #1362)

I'm not sure whether an external PPA would work, but my guess is no, because of security issues.

bradh:
But if you're packaging on 24.04 (for example), then you should have an up-to-date version of Catch2 (3.4.0 in my install), right?

PR #1407, opened 2024-11-21
remove unused header in decoder_libde265.h

roticv:
For code cleaniness, we shouldn't include headers that are not used.


farindk:
This is needed for `MAYBE_UNUSED` and `LIBHEIF_API`.

roticv:
I made some mistake. Will resend commit

PR #1408, opened 2024-11-22
AOM encoder: Expose control over Intra Block Copy (intrabc)

lovell:
The "intra block copy" feature of the AV1 codec can lead to an order of magnitude increase in encoding time when certain conditions/dimensions are met.

This feature, which is switched on by default in libaom but not always triggered, searches for similar areas within the same frame, e.g. repeating text, but at a high CPU cost. My understanding is that it can also prevent the use of (decoding?) loop filters, so useful features like deblocking and deringing are skipped.

There is more context and discussion, including comments from libaom maintainers, at https://github.com/libvips/libvips/issues/2983

/cc @kleisauke who suggested the original patch

### DIFF
--- a/libheif/plugins/encoder_aom.cc
+++ b/libheif/plugins/encoder_aom.cc
@@ -73,6 +73,7 @@ struct encoder_struct_aom
   bool lossless;
   bool lossless_alpha;
   bool auto_tiles;
+  bool intra_block_copy;
 
 #if defined(HAVE_AOM_CODEC_SET_OPTION)
   std::vector<custom_option> custom_options;
@@ -162,6 +163,7 @@ static const char* kParam_alpha_min_q = "alpha-min-q";
 static const char* kParam_alpha_max_q = "alpha-max-q";
 static const char* kParam_lossless_alpha = "lossless-alpha";
 static const char* kParam_auto_tiles = "auto-tiles";
+static const char* kParam_intra_block_copy = "intra-block-copy";
 static const char* kParam_threads = "threads";
 static const char* kParam_realtime = "realtime";
 static const char* kParam_speed = "speed";
@@ -200,7 +202,7 @@ static const char* aom_plugin_name()
 }
 
 
-#define MAX_NPARAMETERS 15
+#define MAX_NPARAMETERS 16
 
 static struct heif_encoder_parameter aom_encoder_params[MAX_NPARAMETERS];
 static const struct heif_encoder_parameter* aom_encoder_parameter_ptrs[MAX_NPARAMETERS + 1];
@@ -372,6 +374,16 @@ static void aom_init_parameters()
   p->boolean.default_value = false;
   p->has_default = true;
   d[i++] = p++;
+
+#if defined(AOM_CTRL_AV1E_SET_ENABLE_INTRABC)
+  assert(i < MAX_NPARAMETERS);
+  p->version = 2;
+  p->name = kParam_intra_block_copy;
+  p->type = heif_encoder_parameter_type_boolean;
+  p->boolean.default_value = true;
+  p->has_default = true;
+  d[i++] = p++;
+#endif
 
   assert(i < MAX_NPARAMETERS + 1);
   d[i++] = nullptr;
@@ -585,6 +597,11 @@ struct heif_error aom_set_parameter_boolean(void* encoder_raw, const char* name,
   } else if (strcmp(name, kParam_auto_tiles) == 0) {
       encoder->auto_tiles = value;
       return heif_error_ok;
+#if defined(AOM_CTRL_AV1E_SET_ENABLE_INTRABC)
+  } else if (strcmp(name, kParam_intra_block_copy) == 0) {
+      encoder->intra_block_copy = value;
+      return heif_error_ok;
+#endif
   }
 
   set_value(kParam_realtime, realtime_mode);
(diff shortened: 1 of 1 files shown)


farindk:
Thanks. Did you see any significant reductions in file size when IntraBC mode is used? If it provides some improvements at least for some types of images (e.g. screenshots), it might make sense to add an option to enable it.

lovell:
Screenshots and subtitles are good examples of the sort of thing I imagine this feature was targeting. Some quick testing suggests up to 10% reduction in file size (but with the 10x slow down) for the libaom-default enabled state vs this PR.

I think that's enough to warrant this being an option, so I'll go ahead and make an update to expose it. Should the default libheif behaviour differ from libaom? Is a consistent encoding time for a given "speed" setting more important than file size?

farindk:
Handling encoding options is difficult because every encoder has its own options and compression vs. bitrate vs. time characteristic. My proposal for the future would be to have options in two level:
- one low level that maps 1:1 to the options of the underlying encoder, and
- high level options that are more meaningful for the user. This could include things like "content type", "speed vs quality".

In that sense, it would be better to have the same defaults in the low level options as the encoder library. In the case above, a "disable-blockcopy" would be more consistent.

If you didn't start yet, I can take over and implement this.

libheif/plugins/encoder_aom.cc
```diff
@@ -373,6 +375,14 @@ static void aom_init_parameters()
   p->has_default = true;
   d[i++] = p++;
 
+  assert(i < MAX_NPARAMETERS);
```
farindk, line 378: Should we add a
`#if defined(AOM_CTRL_AV1E_SET_ENABLE_INTRABC)`
around this block such that the option is only available when supported by our AOM version?

lovell: Good question. None of the other parameters currently do this, and many of those are only available in more recent versions of libaom. I guess we should probably be consistent either way. Adding these guards might make the (debug only) `MAX_NPARAMETERS` checking more complex too.

farindk: Then we should add such a `#if` to the other parameters too unless they are so old that we can assume that they will always be present.

The `assert(i < MAX_NPARAMETERS)` doesn't have to be modified. It is just a quick and dirty way to make sure that we do not write beyond the boundaries of the fixed-size array.

lovell: Thanks Dirk, I've added an `#if` wrapper around the use of this parameter and left existing code as-is for now.

libheif/plugins/encoder_aom.cc
```diff
@@ -162,6 +163,7 @@ static const char* kParam_alpha_min_q = "alpha-min-q";
 static const char* kParam_alpha_max_q = "alpha-max-q";
 static const char* kParam_lossless_alpha = "lossless-alpha";
 static const char* kParam_auto_tiles = "auto-tiles";
+static const char* kParam_intra_block_copy = "intra-block-copy";
```
wantehchang, line 166: It would be good to name this parameter "enable-intrabc" to match the name used in libaom. On the other hand, I am not sure if libheif has tried to make these parameter names match the names used in libaom, so this may not be worth the trouble.

farindk: Yes, I believe this makes sense. Parameters that correspond 1:1 to codec parameters should be identical to reduce the confusion and cognitive load.

libheif/plugins/encoder_aom.cc
wantehchang: I looked into this. `AOM_CTRL_AV1E_SET_ENABLE_INTRABC` first appeared in the `v1.0.0-errata1-avif` tag. (That tag was created for the libavif library, so libaom versions older that that tag don't need to be supported by an AVIF library.)  We can assume `AOM_CTRL_AV1E_SET_ENABLE_INTRABC` is defined and remove the `#if defined(AOM_CTRL_AV1E_SET_ENABLE_INTRABC)` checks.

libheif/plugins/encoder_aom.cc
wantehchang: Thanks for the comment, Dirk. This will require a corresponding change to libvips (see https://github.com/libvips/libvips/pull/4284) to use the new parameter name. During the transitional period libheif can support both the old and new parameter names.

farindk:
I just realized that this explicit option is actually not needed. We have the feature that we can pass arbitrary options to aom like `-p aom:enable-intrabc=0`, which calls `aom_codec_set_option(&codec, "enable-intrabc", "0")` with the same effect.

In this case, I'd still keep the option because it is easier to use when it is listed in the `heif-enc -P` parameters overview.

libheif/plugins/encoder_aom.cc
farindk: @wantehchang I don't understand the reasoning. Why couldn't an aom version before `v1.0.0-errata1-avif` be used together with libheif v1.19?

However, given that the intra-bc flag was added in Dec. 2019, it is probably safe to assume it exists.

wantehchang: Dirk: I could not find libheif's minimum supported version of libaom. That's why I used the `v1.0.0-errata1-avif` tag, which was created for libavif (presumably the first version of libaom ever supported by libavif), to argue that any other AVIF library is likely to need a version of libaom >= `v1.0.0-errata1-avif`.

In my opinion there is no need to support libaom versions < v3.0.0. The latest libaom release is v3.11.0.

Issue #902, opened 2023-06-17
All library functions being exported with BUILD_SHARED_LIBS=Off on MSVC


zenden2k:
When compiling static library with CMake and Visual Studio 2019, all the libheif functions are being exported.
This causes all consumers of this library to have a large list of libheif functions exported from resulting executables.

I see that setting the `BUILD_SHARED_LIBS=Off` option does not result in a macro definition `LIBHEIF_STATIC_BUILD` .
Setting `LIBHEIF_STATIC_BUILD` manually causes compiler errors.

```cpp
#if defined(_MSC_VER) && !defined(LIBHEIF_STATIC_BUILD)
#ifdef LIBHEIF_EXPORTS
#define LIBHEIF_API __declspec(dllexport)
#else
#define LIBHEIF_API __declspec(dllimport)
#endif
#elif defined(HAVE_VISIBILITY) && HAVE_VISIBILITY
#ifdef LIBHEIF_EXPORTS
#define LIBHEIF_API __attribute__((__visibility__("default")))
#else
#define LIBHEIF_API
#endif
#else
#define LIBHEIF_API
#endif
```

rgpublic:
I'm not sure but might be related... I'm also trying to compile libheif statically:

`cmake -DBUILD_SHARED_LIBS=OFF --preset=release-noplugins ..`

Initially, it seems to work (no error) and I get a .a file. But when I try to compile ImageMagick it fails (various missing references). The last version I could compile a static IM binary with is 1.15.2 - I guess due to the autotools still existing there. Sth. seems to be wrong with the CMake static builds...

rgpublic:
Any updates here perhaps? :eyes:

rgpublic:
Just for the record: I've finally gotten to the bottom of this and could for the first successfully compile ImageMagick with a more recent CMake-based libheif (1.19.5). The solution is to add aom and de265 manually to the linker when compiling ImageMagick like so:

`export LIBS='-laom -lde265'`

Interestingly, this was never necessary with the old autotools-based libheif. This could of course also be the solution for other things you might want to compile with a static libheif.

dg0yt:
Duplicates #513.

Issue #513, opened 2021-05-18
Static linking on Windows/MSVC/vcpkg?


Ziriax:
I'm experimenting with static linking and vcpkg, and I get a lot of undefined linker errors when linking the libraries into my final program; it seems `libheif` is trying to reference `libde265` dynamically even when building statically, like:

```
Creating library Release\eXpress.Imaging.Heif.Native.lib and object Release\eXpress.Imaging.Heif.Native.exp
heif.lib(heif_decoder_libde265.cc.obj) : error LNK2019: unresolved external symbol __imp_de265_get_version referenced in function "char const * __cdecl libde265_plugin_name(void)" (?libde265_plugin_name@@YAPEBDXZ)
```

I was able to fix this by modifying the  `libheif/CMakeLists.txt` as follows:

```diff
--- a/libheif/CMakeLists.txt
+++ b/libheif/CMakeLists.txt
@@ -54,10 +54,17 @@ set_target_properties(heif
                           VERSION ${PROJECT_VERSION}
                           SOVERSION ${PROJECT_VERSION_MAJOR})

-target_compile_definitions(heif
-                           PUBLIC
-                               LIBHEIF_EXPORTS
-                               HAVE_VISIBILITY)
+if(BUILD_SHARED_LIBS)
+    target_compile_definitions(heif
+                            PUBLIC
+                                LIBHEIF_EXPORTS
+                                HAVE_VISIBILITY)
+else()
+    target_compile_definitions(heif
+                               PUBLIC
+                                   LIBHEIF_STATIC_BUILD
+                                   LIBDE265_STATIC_BUILD)
+endif()

 if(LIBDE265_FOUND)
     target_compile_definitions(heif PRIVATE HAVE_LIBDE265=1)
``` 

I'm not sure this is the correct way to do this. 

And maybe the `libde265` lib should be modified accordingly?

kasperisager:
I just hit this as well while building a static version of `libheif` using LLVM on Windows.

Tom94:
Same. OP's fix works for me, so for the time being my project depends on a fork of libheif with the fix implemented. (https://github.com/Tom94/tev/pull/253)

dg0yt:
It is 2025, and the problem occurs again in vcpkg.

PR #1415, opened 2024-11-29
Add static flag to libheif.pc

kmilos:
This is just one (small) step in addressing e.g. #513 and #902 for _clients_ of the static library (for which the correct OOTB _build_ of on Windows still needs to be handled).

This is not gated by WIN32 anywhere (i.e. placeholder to be conditionally substituted by CMake), but it doesn't affect other platforms, so might as well hard code it...


farindk:
> Shouldn't this only be set if it is a static build?

Yes, that `Cflags.private` name is confusing.

bradh:
Yes, just as confusing as those....

PR #1416, opened 2024-11-29
Patch to avoid out of bounds assert with a debug build on Windows.

dlemstra:
This patch fixes an out of range assert with a debug build on Windows:

```c++
    _NODISCARD _CONSTEXPR20 const _Ty& operator[](const size_type _Pos) const noexcept /* strengthened */ {
        auto& _My_data = _Mypair._Myval2;
#if _CONTAINER_DEBUG_LEVEL > 0
        _STL_VERIFY(
            _Pos < static_cast<size_type>(_My_data._Mylast - _My_data._Myfirst), "vector subscript out of range");
#endif // _CONTAINER_DEBUG_LEVEL > 0

        return _My_data._Myfirst[_Pos];
    }
```

This happens when `idx + size == indata.size()` and can be avoided by getting the address of the last element and then adding `1` to the pointer.

### DIFF
--- a/libheif/plugins/decoder_openh264.cc
+++ b/libheif/plugins/decoder_openh264.cc
@@ -130,6 +130,7 @@ struct heif_error openh264_decode_image(void* decoder_raw, struct heif_image** o
   std::vector<uint8_t> scdata;
 
   size_t idx = 0;
+  auto end = &indata[0];
   while (idx < indata.size()) {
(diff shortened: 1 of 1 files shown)


farindk:
Thank you. I have simplified it a bit by avoiding the `operator[]` altogether for the end pointer.

PR #1442, opened 2025-01-14
Security: ensure max_size_entity_group limit is enforced

lovell:
Commit https://github.com/strukturag/libheif/commit/41ccc4e82dfd7e0e2f40b27a959ade9ead8304a8 added support for limiting `max_size_entity_group` as part of libheif v0.19.0. This PR adds the necessary return to enforce that limit.


farindk:
Thanks. The error return is definitely missing...

PR #1452, opened 2025-01-26
Added extra CMAKE flags for linking plugins statically

agruzdev:
Currently for statically linked backends there are used only two variables: LIBDE265_INCLUDE_DIR and LIBDE265_LIBRARY, for example. This is not enough for multiconfig build systems, like MSVC, where we need to provide both 'debug' and 'release' libraries.

This commit adds support of optional cmake variables:
* plugin_DEBUG_LIBRARY - if not empty, then target_link_libraries will be called twice:
```
target_link_libraries(heif PRIVATE debug ${${varName}_DEBUG_LIBRARY})
target_link_libraries(heif PRIVATE optimized ${${varName}_LIBRARIES})
```
* plugin_LINK_DIRS - willl be passed to `target_link_directories( )`

This change is backward compatible. In case if the new variables empty, behaviour is not changed.


farindk:
Thanks

PR #1454, opened 2025-01-31
encoder_jpeg.cc: Fix for libjpeg versions < 9d

### LINKED ISSUE #1453: Build issue with (non-turbo) libjpeg < 9d related to plugins/encoder_jpeg.cc
Builds using jpeg-[9a](https://www.ijg.org/files/jpegsrc.v9a.tar.gz),[b](https://www.ijg.org/files/jpegsrc.v9b.tar.gz),[c](https://www.ijg.org/files/jpegsrc.v9c.tar.gz) are failing due to type mismatch invoking `jpeg_mem_dest()`.

Like libjpeg-turbo, the prototypes in non-turbo libjpeg up to jpeg-9c used `unsigned long` (while `LIBJPEG_TURBO_VERSION` is not defined):

```
EXTERN(void) jpeg_mem_dest JPP((j_compress_ptr cinfo,
			       unsigned char ** outbuffer,
			       unsigned long * outsize));
```

The [code in encoder_jpeg.cc](https://github.com/strukturag/libheif/blob/master/libheif/plugins/encoder_jpeg.cc#L363) declares the output size as:

```c
#ifdef LIBJPEG_TURBO_VERSION
  unsigned long outlength = 0;
#else
  size_t outlength = 0;
#endif
```
so this should probably be something like:

```c
#if defined(LIBJPEG_TURBO_VERSION) || (JPEG_LIB_VERSION_MAJOR < 9 || (JPEG_LIB_VERSION_MAJOR == 9 && JPEG_LIB_VERSION_MINOR < 4))
  unsigned long outlength = 0;
#else
  size_t outlength = 0;
#endif
```

jerstlouis:
- Fixes #1453

### DIFF
--- a/libheif/plugins/encoder_jpeg.cc
+++ b/libheif/plugins/encoder_jpeg.cc
@@ -360,7 +360,7 @@ struct heif_error jpeg_encode_image(void* encoder_raw, const struct heif_image* 
   }
 
   uint8_t* outbuffer = nullptr;
-#ifdef LIBJPEG_TURBO_VERSION
+#if defined(LIBJPEG_TURBO_VERSION) || (JPEG_LIB_VERSION_MAJOR < 9 || (JPEG_LIB_VERSION_MAJOR == 9 && JPEG_LIB_VERSION_MINOR < 4))
   unsigned long outlength = 0;
 #else
   size_t outlength = 0;


farindk:
Thank you, @jerstlouis

PR #1457, opened 2025-02-05
encoder_svt: add "tune" param

pszemus:
Add "tune" SVT-AV1 parameter with possible values: 0: VQ, 1: PSNR, 2: SSIM (default: 1)

https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/v2.3.0/Source/API/EbSvtAv1Enc.h?ref_type=tags#L718 (introduced in [v0.9.1](https://gitlab.com/AOMediaCodec/SVT-AV1/-/releases/v0.9.1))


farindk:
Thanks. I will change this to use the names (vq, psnr, ssim) instead of the integer. Should be easier for the user.

PR #1464, opened 2025-02-14
aom: print warning on `aom_codec_set_option` failure

pszemus:
Prints a warning if setting the custom AOM encoder option fails.

I assume we don't want to return an error as described in https://github.com/pszemus/libheif/blob/master/examples/heif_enc.cc#L165 for x265 encoder, but since we have an actual error message here, maybe we want the whole operation (AOM encoder setup) to fail?

### DIFF
--- a/libheif/plugins/encoder_aom.cc
+++ b/libheif/plugins/encoder_aom.cc
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <cstring>
 #include <cassert>
+#include <iostream>
 #include <vector>
 #include <string>
 #include <thread>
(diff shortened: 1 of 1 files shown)


farindk:
I think what I meant with "there is no checking for valid parameters when using the prefix" is that libheif does not know the valid parameters when they are simply passed through. However, if the codec library returns an error, we should properly return this.

I will do the necessary change to this PR.

pszemus:
Thank you @farindk !
Don't you think that every other call to `aom_codec_control` should also be checked for its return value and possibly report an error?

farindk:
Not sure how likely it is that a call to `aom_codec_control` with known parameters fails, but to be on the safe side, I've added error checking to all of them, trying not to clutter the source code too much.

PR #1467, opened 2025-02-18
Build emscripten

dukesook:
I had to make the following two changes in order to successfully run build-emscripten.sh:

1. For some reason, the linker was unable to find `libde265.a` and added that path to the `LIBRARY_LINKER_FLAGS` variable.

2. The compiler threw an "implicit instantiation of undefined template" error regarding the `basic_string` data type.


build-emscripten.sh
```diff
@@ -50,7 +50,7 @@ if [ "$ENABLE_LIBDE265" = "1" ]; then
     fi
     LIBDE265_DIR="$(pwd)/libde265-${LIBDE265_VERSION}"
     CONFIGURE_ARGS_LIBDE265="-DLIBDE265_INCLUDE_DIR=${LIBDE265_DIR} -DLIBDE265_LIBRARY=-L${LIBDE265_DIR}/libde265/.libs"
-    LIBRARY_LINKER_FLAGS="$LIBRARY_LINKER_FLAGS -lde265"
+    LIBRARY_LINKER_FLAGS="$LIBRARY_LINKER_FLAGS -L${LIBDE265_DIR}/libde265/.libs -lde265"
     LIBRARY_INCLUDE_FLAGS="$LIBRARY_INCLUDE_FLAGS -L${LIBDE265_DIR}/libde265/.libs"
```
farindk, line 54: Looks like the `-L` flag was in the wrong line. Please check whether it still works when we remove it from `LIBRARY_INCLUDE_FLAGS` and update the PR accordingly.

farindk: The same issue might be in line 89 with the AOM library.

dukesook: I was still able to compile successfully after removing the libde265 path from `LIBRARY_INCLUDE_FLAGS`. I then just removed `LIBRARY_INCLUDE_FLAGS` altogether since it's no longer being used.

A similar "library not found" error occurred when ENABLE_AOM=1 and the same solution resolved it.

farindk:
Thank you

PR #1468, opened 2025-02-18
Emscripten: Add option for Modern ES6 modules

dukesook:
The `-sEXPORT_ES6` flag adds the following export statement to libheif.js:
```js
export default libheif;
```

This allows libheif.js to be imported with the modern `import` keyword.

I also believe this setting should be on by default since ESM is the standard going forward.


farindk:
Yes, I think this should be explicitly enabled (with `USE_ES6`) because it appears to be not backwards compatible. If I see this correctly, `EXPORT_ES6` also enables `MODULARIZE`, which changes how the functions are exported.

PR #1503, opened 2025-04-06
Add support for handling 'tmap' items

ram-mohan:
Gain map technology offers a way to create an image (henceforth referred
as derived image) from one base input image and a secondary input image
called as a gainmap input image. Reconstruction is done by applying the
gain map to the base image according to iso 21496-1 section 6.

This change provides API to read, write base input image, gain map input
image, gain map metadata. Third party applications can make use of this
API to render only base image or reconstruct the derived image and
render the same.

By default these API is disabled and can be enabled at configure time
using option -DWITH_EXPERIMENTAL_GAIN_MAP=1

TODO: add plugins for encoding, decoding uhdr images

### DIFF
--- a/libheif/context.h
+++ b/libheif/context.h
@@ -47,6 +47,7 @@ class StreamWriter;
 
 class ImageItem;
 
+class ImageMetadata;
 
 // This is a higher-level view than HeifFile.
 // Images are grouped logically into main images and their thumbnails.
@@ -146,6 +147,14 @@ public:
   heif_property_id add_property(heif_item_id targetItem, std::shared_ptr<Box> property, bool essential);
 
   Result<heif_item_id> add_pyramid_group(const std::vector<heif_item_id>& layers);
+
+#if WITH_EXPERIMENTAL_GAIN_MAP
+  Error add_tmap_item(const std::vector<uint8_t>& metadata, heif_item_id& item_id);
+
+  Error link_gain_map(const std::shared_ptr<ImageItem>& primary_image,
+                      const std::shared_ptr<ImageItem>& gain_map_image, const heif_item_id tmap_id);
+#endif
+
 
   // --- region items
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -54,6 +54,7 @@
         "WITH_VVDEC_PLUGIN" : "OFF",
         "WITH_VVENC" : "ON",
         "WITH_VVENC_PLUGIN" : "OFF",
+        "WITH_EXPERIMENTAL_GAIN_MAP" : "OFF",
 
         "WITH_REDUCED_VISIBILITY" : "OFF",
         "WITH_HEADER_COMPRESSION" : "ON",
@@ -110,6 +111,7 @@
         "WITH_VVDEC_PLUGIN" : "ON",
         "WITH_VVENC" : "ON",
         "WITH_VVENC_PLUGIN" : "ON",
+        "WITH_EXPERIMENTAL_GAIN_MAP" : "OFF",
 
         "WITH_REDUCED_VISIBILITY" : "ON",
         "WITH_HEADER_COMPRESSION" : "ON",
@@ -148,6 +150,7 @@
         "WITH_UVG266" : "OFF",
         "WITH_VVDEC" : "OFF",
         "WITH_VVENC" : "OFF",
+        "WITH_EXPERIMENTAL_GAIN_MAP" : "OFF",
 
         "WITH_REDUCED_VISIBILITY" : "ON",
         "WITH_HEADER_COMPRESSION" : "OFF",
--- a/libheif/CMakeLists.txt
+++ b/libheif/CMakeLists.txt
@@ -249,6 +249,10 @@ if (ENABLE_EXPERIMENTAL_MINI_FORMAT)
             mini.cc)
 endif ()
 
+if (WITH_EXPERIMENTAL_GAIN_MAP)
+    target_compile_definitions(heif PUBLIC WITH_EXPERIMENTAL_GAIN_MAP=1)
+endif ()
+
 write_basic_package_version_file(${PROJECT_NAME}-config-version.cmake COMPATIBILITY ExactVersion)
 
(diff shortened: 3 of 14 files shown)


libheif/context.h
```diff
@@ -94,6 +95,14 @@ class HeifContext : public ErrorBuffer
 
   std::shared_ptr<ImageItem> get_primary_image(bool return_error_image);
 
+#if WITH_EXPERIMENTAL_GAIN_MAP
+  std::shared_ptr<ImageItem> get_gain_map_image() { return m_gain_map_image; }
```
farindk, line 99: This only works if there is only a single gain map in the file. The gain map should probably be stored in the `ImageItem` like the alpha image or the depth map.

libheif/context.h
```diff
@@ -94,6 +95,14 @@ class HeifContext : public ErrorBuffer
...
+
+  std::shared_ptr<ImageMetadata> get_gain_map_metadata();
```
farindk, line 101: Same here, the gain map metadata has to be part of `ImageItem` to work correctly when there are several images in a file.

farindk:
Thank you. Note that there is also some work for this in #1121.

I will have a closer look at this when we switched to the v1.20.x branch and can do API additions.

farindk:
Just for reference
https://developer.apple.com/documentation/appkit/applying-apple-hdr-effect-to-your-photos

libheif/context.h
ram-mohan: Thank you for the review, As indicated, i have moved the gainmap info inside ImageItem

libheif/context.h
ram-mohan: Done

Tom94:
Hi there, just wanted to check in on the progress of this. I realize that gain maps can already be decoded as generic auxiliary images, but I don't think it's possible to read gain map metadata without this PR's tmap item support (please let me know if I am wrong) -- that feature alone would already be very useful. Thank you!

fxthomas:
I rebased it on v1.21.2 here: https://github.com/fxthomas/libheif/tree/pr/1503-gain-maps-v1.21.2

A couple of warts in the API so far (I haven't touched most of the original branch, just moved things around when needed following changes in libheif `master`):

- I'm not sure about truncating the `version` field of the `tmap` metadata, it seems to me that it would be more future-proof to just copy the whole structure without modification.
- It looks like having a proper API for handling derived images (`dimg` boxes) might be something to look into for the future, since this is something that is common across several sub-formats.
- There are several hard-coded things that might need some configurable options : the encoded bit depth is hardcoded to 10-bits, and it is currently not possible to store a full ICC profile to the new `tmap` derived image (only `nclx`), unlike some other HEIC images I've found in the wild.

I verified that it is possible to re-encode an input HEIC+gain map file into a new HEIC+gain map file with almost identical data (the "except" is because of re-compression of the original decompressed data) and that both display identically on an HDR display.

Tom94:
> I'm not sure about truncating the version field of the tmap metadata, it seems to me that it would be more future-proof to just copy the whole structure without modification.

To make things worse, I encountered heic images in the wild whose `tmap` metadata doesn't have the `version` field (start directly with the ISO gainmap metadata). And I don't know of a good, spec-compliant way to check for this automatically. (One might consider checking the size for 62 or 142, but the ISO standard permits padding at the end of the metadata which makes that approach ambiguous... works in practice on all images I tried, though.) Long story short: 100% agree with leaving this up to the consumer.

fxthomas:
After checking : the version I have access to (ISO/DIS 21496-1:2024 draft, which I unfortunately cannot link to here because ISO is not public access...) seems to mandate the presence of the version field. In my version this field also contains a struct of 2 big endian int16 values (which matches images generated by iOS devices, for example), but the original PR truncates a single uint8 value (an older revision of the draft?).

Either way all things do point towards leaving interpretation outside of libheif (and not just for `tmap` but for all derived images in general which follow the same principles).

Tom94:
Yes, the two big endian int16 values are always included. What I am saying is that some heic files (notably ones from my iPhone) include an *additional* uint8 version (the one being skipped in this PR), whereas other heic files don't include the uint8 (notably from Hasselblad cameras).

fxthomas:
> some heic files (notably ones from my iPhone) include an additional uint8 version (the one being skipped in this PR)

Ah, I found it elsewhere! ISO/IEC 23008-12 describes the structure of HEIF/HEIC files and the latest version describes how ISO 21496 is supposed to be implemented in those containers.

While I only have a draft (I cannot check if it is still the case in the published version), the version I have has an uint8 version field marked as mandatory, which is then followed by the struct described in ISO 21496. I'm still in favor of just returning the whole data, but this explains why the PR truncated the field originally.

Tom94:
Ooooh, thank you very much for the pointer! (Yet another document to purchase...) That at least settles which of the heic images are compliant.

> I'm still in favor of just returning the whole data, but this explains why the PR truncated the field originally.

FWIW, also still agreeing. Even if only for users to be able to read non-standard metadata.

fxthomas:
Rebased on v1.23.1 : https://github.com/fxthomas/libheif/tree/pr/1503-gain-maps-v1.23.1

It appears to work at first glance but I just did a cursory testing this time.

kleisauke:
FWIW, [libultrahdr v2.0.0](https://github.com/google/libultrahdr/releases/tag/v2.0.0) seems to make use of this PR, see:
https://github.com/google/libultrahdr/pull/413#issuecomment-5225610727

gregbenz:
Thanks @ram-mohan for the original implementation, and @fxthomas for keeping a rebased version available.
This has become an important downstream dependency: libultrahdr v2 uses the gain-map changes, while projects such as libvips appear to be waiting for gain-map support to be available through a normal released libheif rather than a privately patched build.

I’d be happy to help move this forward without duplicating anyone’s work. Which path would the maintainers and original contributors prefer: updating this PR, preparing a new PR against current master with the existing work and attribution preserved, or splitting out a smaller read-side/API change first?

I can help with macOS / Windows builds, tests, and downstream validation with libultrahdr and libvips. If you identify the preferred scope and remaining merge requirements, I’m happy to work within that direction.

Issue #1506, opened 2025-04-20
libheif cannot find libde265 symbols during a static build on Windows


sincos2854:
I performed a static build of libheif together with libde265 on Windows. However, libheif was unable to find the symbols from libde265.

My environment is as follows:
libheif version: v1.19.7
libde265 version: v1.0.15
Directory where libjpeg-turbo and libpng are installed: D:\pkg
Compiler: cl.exe Version 19.43.34810(Visual Studio 2022)
OS: Windows 11 24H2

The CMake execution and errors are as follows:

```cmd
cmake -G "Visual Studio 17 2022" -A x64 ^
-B build ^
-DCMAKE_PREFIX_PATH=D:\pkg ^
-DCMAKE_INSTALL_PREFIX=out ^
-DCMAKE_BUILD_TYPE=Release ^
-DCMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>" ^
-DCMAKE_COMPILE_WARNING_AS_ERROR=OFF ^
-DBUILD_SHARED_LIBS=OFF ^
-DBUILD_TESTING=OFF ^
-DWITH_LIBDE265=ON ^
-DWITH_X265=OFF ^
-DWITH_OpenH264_ENCODER=OFF ^
-DWITH_OpenH264_DECODER=OFF ^
-DWITH_AOM_ENCODER=OFF ^
-DWITH_AOM_DECODER=OFF ^
-DWITH_SvtEnc=OFF ^
-DWITH_RAV1E=OFF ^
-DWITH_OpenJPEG_ENCODER=OFF ^
-DWITH_OpenJPEG_DECODER=OFF ^
-DWITH_OPENJPH_ENCODER=OFF ^
-DWITH_LIBSHARPYUV=OFF ^
-DWITH_GDK_PIXBUF=OFF

cmake --build build --config Release --target install
```

```
MSBuild version 17.13.19+0d9f5a35a for .NET Framework

  1>Checking Build System
  Building Custom Rule D:/a/libheif/libheif/CMakeLists.txt
  bitstream.cc
  box.cc
  error.cc
  context.cc
  file.cc
  file_layout.cc
  pixelimage.cc
  plugin_registry.cc
  nclx.cc
  security_limits.cc
  init.cc
  logging.cc
  compression_brotli.cc
  compression_zlib.cc
  common_utils.cc
  region.cc
  heif.cc
  heif_regions.cc
  heif_plugin.cc
[...]

PR #1507, opened 2025-04-20
Fix static linking with libde265 on Windows

sincos2854:
Please read #1506

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,9 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+if(NOT BUILD_SHARED_LIBS)
+  add_definitions(-DLIBDE265_STATIC_BUILD)
+endif()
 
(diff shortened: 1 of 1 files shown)


farindk:
I don't think this is a good solution. `LIBDE265_STATIC_BUILD` is an implementation detail within libde265 and it should not be exposed or used by any client software (including libheif). Furthermore, this patch assumes that libde265 is built statically exactly when libheif is built statically even though these choices are not related in general.

However, I also do not have a better solution. Support for static linking is not in the main scope of the libraries since it usually indicates a license violation.

sincos2854:
I respect your opinion.

If there are no further discussions, please close the pull request without merging it.

dg0yt:
Transitive usage requirements shouldn't be hardcoded.

sincos2854:
I understand. Thank you very much.

sincos2854:
The issue has been resolved. At least in my environment (Visual Studio 2026), it works properly when passing the following two options to CMake:
```CMake
-DCMAKE_C_FLAGS_INIT="/DLIBDE265_STATIC_BUILD" -DCMAKE_CXX_FLAGS_INIT="/DLIBDE265_STATIC_BUILD"
```
Thank you very much.

PR #1516, opened 2025-05-05
Support versions of libsdl2-dev before 2.0.18

fancycode:
SDL_GetTicks64 was introduced in SDL 2.0.18 but Ubuntu 20.04 still ships 2.0.10

### DIFF
--- a/examples/heif_view.cc
+++ b/examples/heif_view.cc
@@ -56,6 +56,10 @@
 #include "getopt.h"
 #endif
 
+#if !SDL_VERSION_ATLEAST(2, 0, 18)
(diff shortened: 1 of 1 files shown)


farindk:
Isn't Ubuntu 20.04 EOL anyways?

fancycode:
> Isn't Ubuntu 20.04 EOL anyways?

Standard-Support ends this May 31, but there might still be systems that have no upgraded yet or that use extended support until 2030 (and we provide packages starting with 20.04 in the PPA). Also it's a trivial fix ;-)

If you don't want to support SDL2 < 2.0.18, please update the checks in cmake so the incompatibility is already detected earlier and not just through a compile error.

PR #1518, opened 2025-05-06
seq_boxes: saio samples are always stored as uint64

lovell:
The change in this PR also ensures 64-bits of precision are used when calculating the required memory allocation, preventing a possible 32-bit overflow.

This was discovered by the libvips' fuzz tests. If required I can provide an 86 byte (!) AVIF image that results in a memory allocation in excess of the default 4GB limit.


farindk:
Thanks. Good catch. Two bugs fixed in one line :-)

PR #1519, opened 2025-05-09
seq_boxes: check for sbgp read errors within loop

lovell:
Ensures the same fail-fast approach as used with the other boxes parsed by this file.


farindk:
Thank you

PR #1528, opened 2025-06-06
emscripten build: replace basic_string with vector

walkingeyerobot:
expands upon 979f6b5

this fixes emscripten builds that use newer versions of libc++.


farindk:
Thanks

PR #1532, opened 2025-06-17
Fix call to heif_image_handle_is_primary_image

scriby:
The call to heif_image_handle_is_primary_image in HeifImage.prototype.is_primary was missing "Module.", so the function isn't able to be called successfully.


farindk:
Thank you.

PR #1533, opened 2025-06-18
Use nothrow new instead so that we can do a -fno-exception build

msqr1:
We are using try-catch, but this is the one and only time we use it, which prevents building with -fno-exception. Use no-throw new instead.

### DIFF
--- a/libheif/pixelimage.cc
+++ b/libheif/pixelimage.cc
@@ -95,8 +95,6 @@ HeifPixelImage::~HeifPixelImage()
   for (auto& iter : m_planes) {
     delete[] iter.second.allocated_mem;
   }
-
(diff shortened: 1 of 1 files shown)


farindk:
I am fine with using the `nothrow` variant and I made the changes in the above patch. Your PR seems to be to an old libheif version that undoes many of the recent changes. I could not take that one.

msqr1:
Wait, quick question, why didn't you use the `allocation_size` variable when logging the "Allocating 
... bytes failed"? @farindk

farindk:
Right, that probably evolved like this with the duplication. I've simplified it. Thanks for spotting this.

msqr1:
You're welcome

PR #1538, opened 2025-06-27
Fix bad rounding resulting in an out-of-bounds read

hlewin:
We stumbled across an image that produced an out-of-bounds read in `HeifPixelImage::ImagePlane::crop`
See [this valgrind output](https://github.com/user-attachments/files/20941402/valgrind_2.txt)

The problem seems to be caused by using a rounded-up image plane size without checking it against the actual bounds of the source plane.

### DIFF
--- a/libheif/pixelimage.cc
+++ b/libheif/pixelimage.cc
@@ -1087,6 +1087,9 @@ Result<std::shared_ptr<HeifPixelImage>> HeifPixelImage::crop(uint32_t left, uint
     uint32_t plane_top = get_subsampled_size_v(top, channel, m_chroma, scaling_mode::is_divisible); // is always divisible
     uint32_t plane_bottom = get_subsampled_size_v(bottom, channel, m_chroma, scaling_mode::round_up); // keep more chroma
 
+    plane_right = std::min(plane_right, plane.m_width - 1u);
+    plane_bottom = std::min(plane_bottom, plane.m_height - 1u);
+
     auto err = out_img->add_channel(channel,
                                     plane_right - plane_left + 1,
                                     plane_bottom - plane_top + 1,


farindk:
I don't see any issue with the original computation and think that the actual problem might be somewhere else, like an invalid input image. Could you attach the image that is causing this or at least paste the output of `heif-info -d IMG.HEIC` here?

hlewin:
> I don't see any issue with the original computation and think that the actual problem might be somewhere else, like an invalid input image. Could you attach the image that is causing this or at least paste the output of `heif-info -d IMG.HEIC` here?

The state leading to the error:
image

bottom(1663) gets divided by 2, rounding up which gives 832. as plane_bottom is used inclusively this is out-of-bounds.

`➤ ~/.conan/data/libheif/1.19.8/_/_/package/35df3c7e2002d132f45e5417325b19cc71e46fce/bin/heif-info ~/Landschaft027.heic 
MIME type: image/heic
main brand: heic
compatible brands: mif1, heic, miaf

image: 2122x1664 (id=1), primary
  colorspace: YCbCr, 4:2:0
  bit depth: 8
  color profile: no
  alpha channel: no 
  depth channel: no
metadata:
  none
transformations:
  crop: left=0 top=0 right=6 bottom=0
region annotations:
  none
properties:`

I could mail you the image if needed but do not want to post it here as I am unclear about the any copyrights which might apply (it is part of some test-corpus we use internally)

farindk:
Ok, thanks. On second thought, I think there could actually be an error if the cropped distance is even (`6` in your case). If we replace `round_up` with `round_down` in the following lines (just before your patch), it might fix it:
````
    uint32_t plane_left = get_subsampled_size_h(left, channel, m_chroma, scaling_mode::is_divisible);
    uint32_t plane_right = get_subsampled_size_h(right, channel, m_chroma, scaling_mode::round_up); // keep more chroma
    uint32_t plane_top = get_subsampled_size_v(top, channel, m_chroma, scaling_mode::is_divisible); // is always divisible
    uint32_t plane_bottom = get_subsampled_size_v(bottom, channel, m_chroma, scaling_mode::round_up); // keep more chroma
````

If you could send me the image for validation by email, that would be nice. Otherwise, I will generate a similar testcase.

hlewin:
> ./.conan/data/libheif/1.19.8/_/_/package/4ad48c16ce8e9b73864de92028d070f23ae1d3a4/bin/heif-info -d Landschaft027.heic 
MIME type: image/heic
main brand: heic
compatible brands: mif1, heic, miaf
Box: ftyp -----
size: 28   (header size: 8)
major brand: heic
minor version: 0
compatible brands: mif1,heic,miaf

Box: meta -----
size: 483   (header size: 12)
| Box: hdlr -----
| size: 33   (header size: 12)
| pre_defined: 0
| handler_type: pict
| name: 
| 
| Box: pitm -----
| size: 14   (header size: 12)
| item_ID: 1
| 
| Box: iloc -----
| size: 50   (header size: 12)
| item ID: 1
|   construction method: 0
|   data_reference_index: 0
|   base_offset: 519
|   extents: 0,175 175,2206575 2206750,22 
| 
| Box: iinf -----
| size: 35   (header size: 12)
| | Box: infe -----
| | size: 21   (header size: 12)
| | item_ID: 1
| | item_protection_index: 0
| | item_type: hvc1
| | item_name: 
| | hidden item: false
| 
| Box: iprp -----
| size: 339   (header size: 8)
| | Box: ipco -----
| | size: 308   (header size: 8)
| | | index: 1
| | | Box: hvcC -----
| | | size: 224   (header size: 8)
| | | configuration_version: 1
| | | general_profile_space: 0
| | | general_tier_flag: 0
| | | general_profile_idc: 1
| | | general_profile_compatibility_flags: 0110.0000 0000.0000 0000.0000 0000.0000 
| | | general_constraint_indicator_flags: 00000000 00000000 00000000 00000000 00000000 00000000 
| | | general_level_idc: 186
| | | min_spatial_segmentation_idc: 0
| | | parallelism_type: 0
| | | chroma_format: 4:2:0
| | | bit_depth_luma: 8
| | | bit_depth_chroma: 8
| | | avg_frame_rate: 0
| | | constant_frame_rate: 0
| | | num_temporal_layers: 1
| | | temporal_id_nested: 0
| | | length_size: 4
| | | <array>
| | | | array_completeness: 0
| | | | NAL_unit_type: 32
| | | | 40 01 0c 02 ff ff 01 60 00 00 03 00 80 00 00 03 00 00 03 00 ba 00 00 2c 09 00 
| | | <array>
| | | | array_completeness: 0
| | | | NAL_unit_type: 33
| | | | 42 01 02 01 60 00 00 03 00 80 00 00 03 00 00 03 00 ba 00 00 a0 01 0a 20 06 81 72 e4 93 2a 9b 80 86 83 02 00 00 03 00 02 00 00 03 00 32 10 00 
| | | <array>
| | | | array_completeness: 0
| | | | NAL_unit_type: 34
| | | | 44 01 c1 60 ac 60 a6 48 00 
| | | <array>
| | | | array_completeness: 0
| | | | NAL_unit_type: 32
| | | | 40 01 0c 02 ff ff 01 60 00 00 03 00 80 00 00 03 00 00 03 00 ba 00 00 2c 09 00 
| | | <array>
| | | | array_completeness: 0
| | | | NAL_unit_type: 33
| | | | 42 01 02 01 60 00 00 03 00 80 00 00 03 00 00 03 00 ba 00 00 a0 01 0a 20 06 81 72 e4 93 2a 9b 80 86 83 02 00 00 03 00 02 00 00 03 00 32 10 00 
| | | <array>
| | | | array_completeness: 0
| | | | NAL_unit_type: 34
| | | | 44 01 c1 60 ac 60 a6 48 
| | | 
| | | index: 2
| | | Box: ispe -----
| | | size: 20   (header size: 12)
| | | image width: 2128
| | | image height: 1664
| | | 
| | | index: 3
| | | Box: clap -----
| | | size: 40   (header size: 8)
| | | clean_aperture: 2122/1 x 1664/1
| | | offset: -6/2 ; 0/2
| | | 
| | | index: 4
| | | Box: pixi -----
| | | size: 16   (header size: 12)
| | | bits_per_channel: 8,8,8
| | 
| | Box: ipma -----
| | size: 23   (header size: 12)
| | associations for item ID: 1
| | | property index: 1 (essential: true)
| | | property index: 2 (essential: false)
| | | property index: 3 (essential: true)
| | | property index: 4 (essential: false)

hlewin:
As an Info: The bug is present (at least) in versions 1.19.6+ but not in version 1.18.2. I did not check any versions in between

hlewin:
> Ok, thanks. On second thought, I think there could actually be an error if the cropped distance is even (`6` in your case). If we replace `round_up` with `round_down` in the following lines (just before your patch), it might fix it:
> 
> ```
>     uint32_t plane_left = get_subsampled_size_h(left, channel, m_chroma, scaling_mode::is_divisible);
>     uint32_t plane_right = get_subsampled_size_h(right, channel, m_chroma, scaling_mode::round_up); // keep more chroma
>     uint32_t plane_top = get_subsampled_size_v(top, channel, m_chroma, scaling_mode::is_divisible); // is always divisible
>     uint32_t plane_bottom = get_subsampled_size_v(bottom, channel, m_chroma, scaling_mode::round_up); // keep more chroma
> ```
> 
> If you could send me the image for validation by email, that would be nice. Otherwise, I will generate a similar testcase.

Hmm - your rationale for rounding up seemed to good for me to not just use std::min...
I've send you the image per mail

farindk:
Fixed with using `round_down`. The trick is that `right` is a pixel coordinate and not the image width. Thus, the `+1` (or `-1`) for correct rounding is already implicit. If we use a second `+1` for `round_up`, even crop values will result in a too large chroma plane.

PR #1539, opened 2025-06-27
bugfix: separate tai & content id if statements

dukesook:
The closing curly brace for the sample tai timestamps also included the GIMI sample content id if statement.


farindk:
Oops. Thank you for spotting this.

PR #1544, opened 2025-07-02
Update API versions table for v1.20.x

kleisauke:
Compared to v1.19.x, only the `heif_decoding_options` version was incremented; see commit a18cc33fdeefecdcb99a242281911e22d44c794b.


farindk:
Thanks

PR #1545, opened 2025-07-03
Fix clang++ missing field 'request_range' initializer warnings

apocelipes:
Clang++ on macOS will display the following warning message:

image

Clang++ version: Apple clang version 17.0.0 (clang-1700.0.13.5)
libheif version: 1.20.1 (homebrew)

We can eliminate this warning by explicitly specifying the initialization of all struct members.


farindk:
Thank you

PR #1549, opened 2025-07-06
allocate context before getting encoder in tai test

rathann:
This avoids skipping the test when HEVC encoder is built as plugin.


farindk:
Thanks. While this works, I think the clean way would be to add
````
  err = heif_init(nullptr);
  REQUIRE(err.code == heif_error_Ok);
````
at the beginning of the function.

PR #1556, opened 2025-07-15
feat: initial work on adding text items

bradh:
From ISO/IEC 23008-12:2025 Section 6.10

Remaining work:
 - [ ] TextLayoutProperty (`txlo`) descriptive property
 - [ ] Default text layout
 - [ ] Transformative item support 
 - [ ] Extended language tag (`elng`) descriptive property
 - [ ] Font item support


libheif/api/libheif/heif_text.cc
```diff
@@ -0,0 +1,105 @@
...
+  item->text_item = std::move(r);
+  *out = item;
+
+  return heif_error_success;
+}
+
+heif_item_id heif_text_item_get_id(struct heif_text_item* text_item)
+{
+  if (text_item == nullptr) {
+    return -1;
```
farindk, line 93: `heif_item_id` is unsigned. Probably, we can use 0 instead because I think item IDs may not be zero.

libheif/api/libheif/heif_text.cc
```diff
@@ -0,0 +1,105 @@
...
+
+  return text_item->text_item->get_item_id();
+}
+
+const char* heif_text_item_get_content(struct heif_text_item* text_item)
+{
+  if (text_item == nullptr) {
+    return NULL;
+  }
+  return text_item->text_item->get_item_text();
```
farindk, line 104: For consistency, strings should be returned as a copy in newly allocated memory. See for example `heif_track_get_gimi_track_content_id()`. There is already `heif_string_release()` as a catch-all for freeing returned strings.

libheif/text.cc
```diff
@@ -0,0 +1,38 @@
...
+ * along with libheif.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <string>
+#include "text.h"
+#include "error.h"
+#include <iostream>
+
+
+Error TextItem::encode(std::vector<uint8_t>& result) const
```
farindk, line 28: You can pass back the result as a `Result<vector<uint8_t>>`.

farindk:
Thank you. It looks very good. If you want, I can merge this and we do the other parts in separate PRs.

libheif/api/libheif/heif_text.cc
bradh: This is the same practice used in the region items implementation (where I copied it from). See https://github.com/strukturag/libheif/blob/master/libheif/api/libheif/heif_regions.cc#L76-L83

libheif/text.cc
bradh: Again based on the `RegionItem` implementation - https://github.com/strukturag/libheif/blob/c6b98e5d9c2008cde501dfb0184ce89b46d985c2/libheif/region.cc#L118-L172

libheif/api/libheif/heif_text.cc
farindk: Ok, but then I would consider `heif_region_item_get_id()` also buggy. `-1` when represented as `0xFFFFFFFF` is a valid ID. I think we can just fix `heif_region_item_get_id()` to return 0 because it only returns `-1` when the input is invalid. Thus, I would take the risk of this non-backwards-compatible change.

bradh: OK. I'll do the region changes first (as a separate commit / PR).

libheif/api/libheif/heif_text.cc
farindk: Yes, I know. It's not ideal that this is not consistent everywhere, but returning a copy is more future-proof since we can also return strings that are generated or converted on-the-fly. If we just return `.c_str()` we always have the issue who will manage the memory.

libheif/api/libheif/heif_text.cc
bradh: Updated this implementation - the region change is on #1557

libheif/api/libheif/heif_text.h
```diff
@@ -0,0 +1,130 @@
...
+LIBHEIF_API
+heif_error heif_context_get_text_item(const heif_context* context,
+                                      heif_item_id text_item_id,
+                                      heif_text_item** out);
+
+/**
+ * Get the item identifier for a text item.
+ *
+ * @param region_item the text item to query
+ * @return the text item identifier (or -1 if the region_item is null)
```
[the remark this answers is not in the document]
bradh: Updated this implementation to return 0 (also fixed cut-n-paste error).

libheif/api/libheif/heif_text.h
```diff
@@ -0,0 +1,130 @@
...
+LIBHEIF_API
+heif_item_id heif_text_item_get_id(heif_text_item* text_item);
+
+/**
+ * Get the item content for a text item.
+ *
+ * This is the payload text, in the format given by the associated content_type.
+ *
+ * @param region_item the text item to query
+ * @return the text item content (or null if the region_item is null)
```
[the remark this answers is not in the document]
bradh: Fixed.

bradh:
Power cycling to kick CI off again.

libheif/text.cc
bradh: I've switched the implementation to use this.

farindk:
Thank you for the changes. If you remove the "draft" state, I can merge this already.

farindk:
Thanks. I have moved the conversion of the text to a C-string into the C wrapper.

PR #1561, opened 2025-07-24
consolidate property extraction code

bradh:
This consolidates some duplicate code where we extract properties from items.

It prepares for the text item extended language property, which had more of the duplication.

It does require a "embed the implementation in the header" so that the specialisation is visible, which is slightly ugly.

### DIFF
--- a/libheif/api/libheif/heif_experimental.cc
+++ b/libheif/api/libheif/heif_experimental.cc
@@ -38,32 +38,19 @@ struct heif_property_camera_intrinsic_matrix
 
 struct heif_error heif_item_get_property_camera_intrinsic_matrix(const struct heif_context* context,
                                                                  heif_item_id itemId,
-                                                                 heif_property_id propertyId,
                                                                  struct heif_property_camera_intrinsic_matrix** out_matrix)
 {
   if (!out_matrix || !context) {
     return {heif_error_Usage_error, heif_suberror_Invalid_parameter_value, "NULL passed"};
   }
 
-  auto file = context->context->get_heif_file();
-
-  std::vector<std::shared_ptr<Box>> properties;
-  Error err = file->get_properties(itemId, properties);
-  if (err) {
-    return err.error_struct(context->context.get());
-  }
-
-  if (propertyId < 1 || propertyId - 1 >= properties.size()) {
-    return {heif_error_Usage_error, heif_suberror_Invalid_property, "property index out of range"};
-  }
-
-  auto cmin = std::dynamic_pointer_cast<Box_cmin>(properties[propertyId - 1]);
+  auto cmin = context->context->find_property<Box_cmin>(itemId);
   if (!cmin) {
-    return {heif_error_Usage_error, heif_suberror_Invalid_property, "wrong property type"};
+    return cmin.error.error_struct(context->context.get());
   }
 
   *out_matrix = new heif_property_camera_intrinsic_matrix;
-  (*out_matrix)->matrix = cmin->get_intrinsic_matrix();
+  (*out_matrix)->matrix = cmin.value->get_intrinsic_matrix();
 
   return heif_error_success;
 }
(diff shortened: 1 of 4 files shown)


farindk:
There is a function `template<class BoxType> std::shared_ptr<BoxType> HeifFile::get_property_for_item(heif_item_id imageID) const`.
It seems like the content of the new `properties.h` could simply be moved into `HeifFile` as a similar function, or even using that existing function and removing the need for the `propertyId` altogether, as I assume that there can only be on `elng` property for each item.

bradh:
> There is a function `template<class BoxType> std::shared_ptr<BoxType> HeifFile::get_property_for_item(heif_item_id imageID) const`. It seems like the content of the new `properties.h` could simply be moved into `HeifFile` as a similar function, or even using that existing function and removing the need for the `propertyId` altogether, as I assume that there can only be on `elng` property for each item.

I can move it into `HeifFile`. Putting it onto `HeifContext` would be slightly easier, if you're OK with that.

The propertyId can be omitted for `elng`, as well as some of the boxes that are affected by this PR: `clap`, `imir`, `irot` and `cmin`. However `cmex` and `udes` are allowed multiple times. 

Also, skipping the `propertyId` would mean a different API, such that `heif_item_get_property_camera_intrinsic_matrix` would not require the `propertyId` parameter, but `heif_item_get_property_camera_extrinsic_matrix` would require `propertyId`. That seems OK since those are experimental.

However `heif_item_get_property_transform_mirror`, `heif_item_get_property_transform_rotation_ccw`, and `heif_item_get_property_transform_crop_borders` would need to be deprecated and replaced with a simpler version.

farindk:
> However `heif_item_get_property_transform_mirror`, `heif_item_get_property_transform_rotation_ccw`, and `heif_item_get_property_transform_crop_borders` would need to be deprecated and replaced with a simpler version.

We cannot easily remove those functions, but we can change them to simply ignore the `propertyId`. Mark them deprecated and provide a new variant without the parameter.

farindk:
> However `cmex` and `udes` are allowed multiple times.

How do multiple `cmex` make sense? Each using a different representation (quaternion vs. matrix)?

bradh:
> > However `cmex` and `udes` are allowed multiple times.
> 
> How do multiple `cmex` make sense? Each using a different representation (quaternion vs. matrix)?

Its not explicit in the spec, just "Quantity (per item): zero or more". The world-vs-quaternion option is one possibility.

However there are flags that allow specifying partial sets (e.g. just the orientation). So for a case where its multi-cameras on a single mount or a panorama collection,  you could use the same property for the location, and different properties for the orientation, and hence share the extrinsics. Maybe that is what they had in mind?

bradh:
I've done an initial conversion. When looking over the change, possibly it would be better if it used the new `Result<>` style return. I'll look at that as a follow-up.

There is a specific use for where a property identifier is used, which is shown in the `heif-info` example at https://github.com/strukturag/libheif/blob/master/examples/heif_info.cc#L483-L521

I lean to deprecating the `heif_item_get_transformation_properties` function and just getting the mirror, rotation and crop (`imir`, `irot`, `clap`) properties directly. Possibly there is an argument for returning all the transformative properties and making sure that the user deals with them though (e.g. if `iscl` was supported on the library side, we'd hit the default `assert` case in that example switch).

bradh:
I've now switched to the `Result<>` pattern for this.

farindk:
> I lean to deprecating the `heif_item_get_transformation_properties` function and just getting the mirror, rotation and crop (`imir`, `irot`, `clap`) properties directly. Possibly there is an argument for returning all the transformative properties and making sure that the user deals with them though

Getting the list of transformation properties is important because they have to be applied in the correct order. Crop before rotation is not the same as rotation before crop.

farindk:
Proposal: instead of introducing new functions without the `propertyId` parameter, how about defining that if `propertyId==0`, it will return the first property that it finds of this type?
That is easy to use, backwards compatible, there is no confusion whether the `propertyId` parameter has to be given or not, and in cases where there are multiple similar properties, we still can access them.

Note that in an early version of the standard, there was only one `colr` property allowed, but this has changed. Thus, I would not bet on it that there cannot be, e.g., two `imir` in the future.

farindk:
I've rebased your PR to the changes in the `Result<>` class. I've merged it, but with the change that we keep the transform property access function with the `propertyId` parameter. However, it is now allowed to use a zero `propertyId` to just get the first property of that type. I think this is the best overall solution.

bradh:
Thanks for the fixes and merge

PR #1562, opened 2025-07-28
Fix undefined symbol "Error::Ok" in svtenc plugin.

fancycode:
While packaging libheif 1.20.1 for Debian, I noticed this error while running the tests:
```
7: dlopen: /build/libheif-1.20.1/obj-x86_64-linux-gnu/libheif/plugins/libheif-svtenc.so: undefined symbol: _ZN5Error2OkE
```

Similar to #1541, the plugin needs linking to `error.cc`.

Maybe there should be an internal common static library that is linked by all plugins, so you have to specify these dependencies (together with `common_utils.{cc,h}`) only once.

### DIFF
--- a/libheif/plugins/CMakeLists.txt
+++ b/libheif/plugins/CMakeLists.txt
@@ -66,7 +66,7 @@ set(AOM_ENCODER_extra_plugin_sources ../error.cc ../common_utils.cc ../common_ut
 plugin_compilation(aomenc AOM AOM_ENCODER_FOUND AOM_ENCODER AOM_ENCODER)
 
 set(SvtEnc_sources encoder_svt.cc encoder_svt.h)
-set(SvtEnc_extra_plugin_sources ../common_utils.cc ../common_utils.h)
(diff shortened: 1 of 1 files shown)


farindk:
> Maybe there should be an internal common static library that is linked by all plugins, so you have to specify these dependencies (together with `common_utils.{cc,h}`) only once.

Yes, that's on my list already for a long time. The difficult part is `heif_image` which is used in plugins and which would pull in most of libheif. I guess the only way to handle this is to have a "base" image class with only the most basic functionality and an extended image class containing the remaining functionality of the current `HeifPixelImage`.

fancycode:
Would be nice to also have this in the next 1.20.x release so I don't have to keep the patches for Debian.

farindk:
> Would be nice to also have this in the next 1.20.x release so I don't have to keep the patches for Debian.

I've copied them to the 1.20.x branch, but I assume that the next release will be v1.21.0.

PR #1563, opened 2025-07-28
Load plugins in "get_filtered_encoder_descriptors".

fancycode:
The test "tai.cc" checks for HEVC encoding support by calling `heif_context_get_encoder_for_format` (through `get_encoder_or_skip_test`) which internally calls `get_filtered_encoder_descriptors`.

This however doesn't load the plugins, causing the test to be skipped if the HEVC encoder is provided by a plugin.

### DIFF
--- a/libheif/plugin_registry.cc
+++ b/libheif/plugin_registry.cc
@@ -273,8 +273,6 @@ void register_encoder(const heif_encoder_plugin* encoder_plugin)
 
 const struct heif_encoder_plugin* get_encoder(enum heif_compression_format type)
 {
-  load_plugins_if_not_initialized_yet();
-
   auto filtered_encoder_descriptors = get_filtered_encoder_descriptors(type, nullptr);
   if (filtered_encoder_descriptors.size() > 0) {
     return filtered_encoder_descriptors[0]->plugin;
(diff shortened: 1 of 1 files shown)


farindk:
That should actually be resolved already by 3334bbff0b7b1596af5758d26ed2e2d1b785f992.

fancycode:
Thanks, I didn't see this. However other code that just calls `heif_context_get_encoder_for_format` will be affected.

farindk:
Yes, but people (apparently including myself) should be educated to use `heif_init()`. The `load_plugins_if_not_initialized_yet()` is there for the old API functions that existed before `heif_init()` was introduced in order not to break old programs.

fancycode:
I was just wondering why `get_encoder` in the file was implicitly loading the plugins but `get_filtered_encoder_descriptors` was not ;-) For consistency I think it should do so in both.

Feel free to close anyway as your referenced commit also fixes the test, so my Debian patch based on this PR to fix the test will no longer needed with the next release.

farindk:
Ok, I'm going to merge this as the two functions are quite close. Makes sense. Thank you.

fancycode:
Thanks for merging!

fancycode:
Similar to #1562 please backport this to the next 1.20.x release.

PR #1583, opened 2025-08-25
Ignore CMakeUserPresets.json

dukesook:
Users can override the provided `CMakePresets.json` with their own custom `CMakeUserPresets.json`, which should not be tracked by git.


farindk:
Ok. Thanks.

PR #1585, opened 2025-09-01
Defer errors in images until they are decoded.

fancycode:
Potential fix for #1448


fancycode:
The main image from #1448 could be decoded with this change.

farindk:
Alternative fix has been applied in https://github.com/strukturag/libheif/commit/a160484329da186b5c31a86f4cd5b2f6cf7e3ad3.

fancycode:
Thanks!

PR #1586, opened 2025-09-01
Move check for "bit" header to cmake.

fancycode:
This fixes an issue when building on older OSX with clang (reported as `Apple LLVM version 10.0.1 (clang-1001.0.46.1)` but doesn't provide the `bit` header.


farindk:
thanks

PR #1591, opened 2025-09-03
bufix: dereference Result<heif_item_id>

dukesook:
The `heif_context_add_mime_item()` function wasn't returning the correct `heif_item_id`.


farindk:
Thanks for the catch. Funny that this compiled...

PR #1595, opened 2025-09-05
CMake: add option to disable heif-thumbnailer

kmilos:
Modern distros have lately been increasingly switching to sandboxed [glycin](https://gitlab.gnome.org/GNOME/glycin) for thumbnail generation, so this is just an option for streamlined builds...


farindk:
Thank you

PR #1601, opened 2025-09-19
pixelimage: ensure access to ICC profile in colr box

lovell:
See commit 1dee796, which appears to have introduced a guard but using the wrong profile type.


farindk:
Thank you for catching this.

PR #1602, opened 2025-09-26
Add a plugin based on WebCodecs (hardware decoding w/ webassembly)

scriby:
Decodes images using the VideoDecoder API when run from webassembly in a browser.

The plugin calls inlined Javascript code (`decode_with_browser_hevc`) to perform the decoding. The javascript code will either return NV12 (YCbCr) or RGBA formatted data. NV12 is preferred if available.

In order to call the Javascript VideoDecoder API, 3 things are required:

* A ISO/IEC 14496-15 codec string (e.g. "hvc1.3.E.L93.B0")
* A binary serialized HEVCDecoderConfigurationRecord
* The image data (w/ VPS, SPS, PPS NAL unit stripped), prepended with 4 bytes describing the length

The codec string and HEVCDecoderConfigurationRecord are built from information in the NAL units that are pushed into the plugin. In the case of RGBA output, it is used directly (accounting for potentially differing strides), and NV12 input is converted to the YCbCr format used by libheif by de-interlacing the UV planes.

Tested:

* 3 different generations of iPhone photos
* A `heix` image (RGBA output)
* A `mif1` image (examples/example.heic)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,6 +290,7 @@ option(WITH_UNCOMPRESSED_CODEC " Support internal ISO/IEC 23001-17 uncompressed 
 message("\n=== Summary of compiled codecs ===")
 plugin_compilation_info(LIBDE265 LIBDE265 "libde265 HEVC decoder")
 plugin_compilation_info(FFMPEG_DECODER FFMPEG_avcodec "FFMPEG HEVC decoder (HW acc)")
+plugin_compilation_info(WEBCODECS WEBCODECS "WebCodecs HEVC decoder")
 plugin_compilation_info(X265 X265 "x265 HEVC encoder")
 plugin_compilation_info(KVAZAAR KVAZAAR "Kvazaar HEVC encoder")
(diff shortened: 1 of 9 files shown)


scriby:
I checked into the broken build and I don't see any errors. If I had to guess maybe it just took too long?

`All requested installations completed successfully in: 45 min`

scriby:
Hi :)

I'm sure you're busy, but I just wanted to give a friendly ping to see if you have any updates in terms of when you might be able to take a look at this. No pressure at all, but I just wanted to get a better idea of the timeline.

Thanks again,

Chris

farindk:
I am on several business trips these days. Will leave again on Monday for a week.

I think the PR misses at least a modified `plugins/CMakeFiles.txt`. Did you miss to include that?
In case you don't have that and you just hacked in the compilation somewhere, I can add that.
We will then need to add an option to enable the plugin from the `build-emscripten.sh` script.

scriby:
I have been using a different build system, so I hadn't updated this ones in this repo yet.

I worked on some updates to the build files in this repo and was able to produce a libheif.wasm file using `build_emscripten.sh`.

PR #1605, opened 2025-10-09
Webcodecs: Handle limited range NV12 data & transparency fix

scriby:
NV12 data coming from the browser's VideoDecoder API may be using a different range of pixel values (instead of 0-255). The pixel ranges need to be re-mapped to use the full range expected by libheif.

In particular, this caused transparent backgrounds to be light gray due to the bottom end of the range starting at 16 instead of 0.

This PR also adds explicit handling of the mono alpha channel that's used to encode the transparency information in HEIC images when dealing with NV12 data.

P.S.: Not related to the changes in this PR, but I noticed that the alpha channel does not work properly when decode_with_browser_hevc returns RGBA data. I tried a few different ways to get it to work, but wasn't able to (the background is always black). I'm not sure if libheif is setup to handle the alpha channel mask when using RGBA data, or if there was an issue with my approach.

I was able to get it to work by converting RGBA data to NV12 and reusing that pathway. I know you didn't like that approach before, but it has a bit more meaning now to reuse the code b/c there's more bespoke code on each pathway for handling mono images. Let me know if you want a PR for that or if you want to try to work on getting alpha support to work with RGBA formatted data.

Note that from my perspective it is not that big of a problem, because I haven't yet seen any images with transparency that are using the code path that returns RGBA. But it might possibly be more likely on other hardware that I'm not testing on.

For my purposes, the webcodecs plugin is already very useful as it is, b/c HEIC with alpha channels are already somewhat rare to begin with.

### DIFF
--- a/libheif/plugins/decoder_webcodecs.cc
+++ b/libheif/plugins/decoder_webcodecs.cc
@@ -24,11 +24,15 @@
 #include "libheif/bitstream.h"
 #include "libheif/plugins/nalu_utils.h"
 
+#include <algorithm>
 #include <assert.h>
 #include <cstring>
 #include <emscripten/emscripten.h>
 #include <emscripten/bind.h>
 #include <memory>
+#include <queue>
+#include <string>
+#include <vector>
 
 
 struct NALUnit {


scriby:
Going to re-make this for a cleaner diff

PR #1606, opened 2025-10-09
Webcodecs: Handle limited range pixel values in NV12 data.

scriby:
NV12 data coming from the browser's VideoDecoder API may be using a different range of pixel values (instead of 0-255). The pixel ranges need to be re-mapped to use the full range expected by libheif.

In particular, this caused transparent backgrounds to be light gray due to the bottom end of the range starting at 16 instead of 0.

This PR also adds explicit handling of the mono alpha channel that's used to encode the transparency information in HEIC images when dealing with NV12 data.

P.S.: Not related to the changes in this PR, but I noticed that the alpha channel does not work properly when decode_with_browser_hevc returns RGBA data. I tried a few different ways to get it to work, but wasn't able to (the background is always black). I'm not sure if libheif is setup to handle the alpha channel mask when using RGBA data, or if there was an issue with my approach.

I was able to get it to work by converting RGBA data to NV12 and reusing that pathway. I know you didn't like that approach before, but it has a bit more meaning now to reuse the code b/c there's more custom code on each pathway for handling the mono/transparency mask. Let me know if you want a PR for that or if you want to try to work on getting alpha support to work with RGBA formatted data.

Note that from my perspective it is not that big of a problem, because I haven't yet seen any images with transparency that are using the code path that returns RGBA (I have to "force" it by modifying the code). But it might possibly be more likely on other hardware that I'm not testing on.

For my purposes, the webcodecs plugin is already very useful as it is, b/c HEIC with alpha channels are already somewhat rare to begin with.

### DIFF
--- a/libheif/plugin_registry.cc
+++ b/libheif/plugin_registry.cc
@@ -26,7 +26,7 @@
 #include "init.h"
 
 #if HAVE_WEBCODECS
-#include "third_party/libheif/libheif/plugins/decoder_webcodecs.h"
+#include "plugins/decoder_webcodecs.h"
 #endif
 
 #if HAVE_LIBDE265
--- a/libheif/plugins/CMakeLists.txt
+++ b/libheif/plugins/CMakeLists.txt
@@ -135,6 +135,7 @@ if (WITH_WEBCODECS)
    target_sources(heif PRIVATE
    		  decoder_webcodecs.h
 		  decoder_webcodecs.cc)
+   target_compile_definitions(heif PRIVATE HAVE_WEBCODECS=1)
 endif()
 
 


silverbacknet:
There are two different modes in video that rarely existed in images, but now that video and image share formats, they have become relevant. What's often called full-range, PC-mode, or JPEG-mode is from 0-255, while limited-range, TV-mode, or video-mode is from 16-235 (16-240 chroma), with those numbers scaled up for 10+ bit. The latter is by far the most common for any video you'll encounter, the former for images.

Libheif supports both via the full_range_flag in nclx, and will flag an image as being in limited or full range -- the decoder should handle that. (Which decoders have been getting much better about actually doing.) I think you can override it to get the right color conversion even when decoding. That saves the hit of an extra color conversion step plus additional round-off errors from the scaling.

scriby:
I tried to use that information to update the approach, but wasn't able to get something fully working.

Here's what I tried:

Create the nclx color profile:

```
  struct heif_color_profile_nclx* nclx = heif_nclx_color_profile_alloc();
  nclx->color_primaries = get_heif_primaries(primaries);
  nclx->transfer_characteristics = get_heif_transfer(transfer);
  nclx->matrix_coefficients = get_heif_matrix(matrix);
  nclx->full_range_flag = is_full_range ? 1 : 0;
  heif_image_set_nclx_color_profile(*out_img, nclx);
  heif_nclx_color_profile_free(nclx);
```

Implemented conversion functions to convert from the values returned by the browser's decoded VideoFrame:

```
static heif_color_primaries get_heif_primaries(const std::string& p) {
  if (p == "bt709") return heif_color_primaries_ITU_R_BT_709_5;
  if (p == "smpte170m") return heif_color_primaries_ITU_R_BT_601_6;
  // output truncated for simplicity
}

static heif_transfer_characteristics get_heif_transfer(const std::string& t) {
  if (t == "bt709") return heif_transfer_characteristic_ITU_R_BT_709_5;
  if (t == "smpte170m") return heif_transfer_characteristic_ITU_R_BT_601_6;
  // output truncated for simplicity
}

static heif_matrix_coefficients get_heif_matrix(const std::string& m) {
  if (m == "bt709") return heif_matrix_coefficients_ITU_R_BT_709_5;
  if (m == "smpte170m") return heif_matrix_coefficients_ITU_R_BT_601_6;
  // output truncated for simplicity
}
```

It would also be possible to grab this data from the VUI section of the SPS data, but the SPS parsing function we've got in this file doesn't have code to parse the VUI info.

I printed some debug output to make sure this was all getting set properly, and I saw primaries = smpte170m (6), transfer = bt709 (1), matrix = smpte170m (6), and full range = 0.

All these values seem correct to me, so I'm not quite sure why it doesn't work. Is there anything special you have to do when using the image data, such as over in [post.js](https://github.com/strukturag/libheif/blob/0e043e64bf7038e4feefbcd24ca88d46db93d94c/post.js#L74)?

scriby:
I've investigated this further and discovered two things that prevent setting of the nclx profile from working as expected.

* The color conversion functions in yuv2rgb.cc don't re-map the alpha channel from limited to full range. This causes fully transparent alpha channels to be light gray even when a conversion function that handles limited range data is selected.

* I am seeing that even if I return nclx data on the decoded image, libheif may ignore it. I tried to dig into this and I think what's going on is that if a HEIF image contains nclx data in the colr box, it will be used instead of the nclx data from the decoded tiles. This can cause it to incorrectly think that the image is using full range, even if the decoded image data is using limited range. I am testing with an image generated from an iPhone that has an alpha channel and am seeing that `Op_YCbCr420_to_RGB32` is being selected for color conversion even though the decoded tiles are in limited range and `Op_YCbCr420_to_RGB32` only handles full range. In particular I am seeing that decoder_webcodecs.cc is setting primaries = 6, transfer = 1, matrix = 6,  full range = 0  but yuv2rgb.cc is instead seeing primaries = 1, transfer = 13, matrix = 6, full range = 1.

After this investigation I can see that it doesn't make sense to do the limited range color conversion in the decoder plugin. So I will go ahead and update this PR just to properly set the nclx profile info on the decoded image.

However, I still think both things I raised above should be addressed to make color conversion & preserve the alpha channel properly for data returned from the webcodecs plugin. If there's any direction on that and whether you'd like me to take either thing on let me know.

Thanks!

farindk:
@scriby Could you please attach the image with the limited-range alpha channel? Do you know which software created this?

The HEIF standard does not tell anything about the range of the alpha data. I can understand that images originally converted from JPEG use the limited range also in the HEIF file for the color channels, but I assumed that the alpha plane will always be full-range. I also see no way how a decoder could decide which range is correct. The `colr` color profile has the full-range flag, but my interpretation is that this only applies to the color channels. Since there is no alpha channel in legacy JPEG images or analog content, I see no reason why a HEIF file should store alpha with limited range.

scriby:
[2521.heic.zip](https://github.com/user-attachments/files/22889967/2521.heic.zip)

I created this file by taking a picture on an iphone, long pressing on a subject in the photo, copying it, and then pasting into iMessage and sending the message.

This image decodes properly with a transparent background when using libde265. So I'm not sure whether the original image is using limited range, or whether the browser decoding APIs are just producing limited range data anyway.

Update: I tested several HEIC images and they all decode with `colorSpace.fullRange = false`. I am pretty sure that the browser's HEVC decoding API is just always returning limited range colors no matter what the original input is. It also seems possible this is hardware dependent.

farindk:
Thanks for the example image. This image sets `full_range=true` in the `colr` color profile for the color image:
````
| | | Box: colr -----
| | | size: 19   (header size: 8)
| | | colour_type: nclx
| | | colour_primaries: 2
| | | transfer_characteristics: 2
| | | matrix_coefficients: 6
| | | full_range_flag: 1
````

For the alpha plane, there is no `colr` box and hence no `full_range` information, as expected. The encoded data is full-range. Thus, I also think that the browser's decoder just clamps the output (wrongly) to 16-235.

The question is still where the decoder gets the `full_range=false` from.
It might also be that the browser decoder fills the colorspace information from the H.265 bitstream VUI and that this does not match the `colr` box in the HEIF container. The `colr` box takes precedence if both do not agree, but the browser decoder only sees the H.265 bitstream and might thus clamp the values. In that case, the encoder might be the guilty part, as it does not save a correct VUI header into the H.265 bitstream.

The best way to handle this seems to be to work around this browser bug by scaling it back to 0-255 in the decoder plugin.

BTW: I see that there is a alpha-channel metadata box `alpp` under consideration for HEIF 4th edition:
https://www.mpeg.org/wp-content/uploads/mpeg_meetings/151_Daejeon/w25291.zip
https://www.mpeg.org/wp-content/uploads/mpeg_meetings/151_Daejeon/w25442.zip

Instead of a full-range flag, there might be the values `opaque_value` / `transparent_value`, which seem to define the alpha range. However, this might be something for the future, but not for now.

**Comment by farindk** [comment]
PS: I had a look into the VUIs of the H.265 bitstreams of your example image. The VUIs are correct.
For color tiles: `colour_primaries=2, transfer_characteristics=2, matrix_coefficients=6, full_range_flag=1`
For alpha tiles: `colour_primaries=2, transfer_characteristics=2, matrix_coefficients=2, full_range_flag=1`

This matches the `colr` and indicates full-range for alpha. If the browser returns `full_range=0`, it is incorrect.

**Comment by scriby** [comment]
Heads up that this doesn't just affect the alpha channel. The actual color data is also using a limited range. IIRC the color data is using 16 - 240 and the alpha channel uses 16 - 235.

So even the normal colors of browser decoded images are off (at least using my hardware, I haven't been able to test this across different hardware yet).

I think there are 2 main ways to resolve it:

1) Use the nclx color profile information set by the decoder plugin on the returned heif image to map the colors in yuv2rgb.cc.

I've noticed that even though I'm setting nclx color profile info on the decoded image (similar to the decoder_libde265 plugin), it's not actually used by libheif. I think libheif may be defaulting to the info from the HEIF container's metadata.

This would also require an update to the yuv2rgb code to also map the alpha channel, which you might not want in general. It could be added as a setting on the nclx color profile data though (default off).

2) Map colors from limited to full range in the plugin.

As it is now, the plugin doesn't have enough information to know whether it should re-map or not. It would need to check for whether there is a mismatch between the full range flag that was returned by the decoder and what is specified on the colr box, or possibly the VUI info from SPS. The SPS parsing code in the plugin doesn't parse the VUI data, so it would need to be updated to read the full range flag out of VUI.


Investigating into the bug a little more, it looks like the hardware decoder on macOS may always returns limited range colors for HEVC streams: https://github.com/mpv-player/mpv/issues/6546

**Comment by farindk** [comment]
Do you know whether the browser decoder scales the output values or whether it just clips them to the limited range?
I see no reason why it should scale it. I think it is more likely that it just clips the values. In that case, we might make things even worse by scaling the values.

**Comment by scriby** [comment]
Just by looking at it I think it looks scaled as opposed to clamped.

I'll do some comparisons with the raw output, the scaled output, and what libde265 returns to make sure.

**Comment by scriby** [comment]
I've got some interesting results from testing.

| Decoder    | 2521.heic (dog w/ alpha channel) | go board.heic |
| -------- | ------- | ------- |
| Webcodecs w/ set nclx profile (this current PR)  | Gray background, limited range colors, looks bad | same as libde265 |
| Webcodecs w/o set nclx profile (original webcodecs PR) | Same as above | Limited range colors, looks bad |
| [Webcodecs w/ custom scaling](https://raw.githubusercontent.com/strukturag/libheif/5e76a5a0ecd45e9f227c05442a8c4bd57db78b10/libheif/plugins/decoder_webcodecs.cc) | Transparent background, colors look better but not exact match to libde265 | Colors look ok but not exact match to libde265 |
| libde265 | perfect | perfect |

The overall result is that this PR improves colors for some images with the webcodec decoder when the decoder returns limited range colors (such as on my mac and perhaps every mac).

Comparing the colr boxes between the image that works well ([go board.heic.zip](https://github.com/user-attachments/files/22917310/go.board.heic.zip)) and the one that doesn't work well ([2521.heic](https://github.com/user-attachments/files/22889967/2521.heic.zip)), the difference is that 2521.heic has colour_type = nclx and "go board.heic" has
colour_type = prof.

Best guess is that when the overall colour_type is prof, the nclx color profile from the decoder will be used, whereas when there is already a nclx profile in the container, it will be used instead.

As far as solving the issue, my sense is that libheif should use the nclx profile that the decoder sets (if present), and otherwise falls back to the one in the heif metadata.

That would at least allow libheif to re-map the color data in this case. As for fixing up the alpha channel, the plugin itself could do that if libheif doesn't think it's the right call to scale the alpha channel in general

**Comment by scriby** [comment]
https://github.com/strukturag/libheif/blob/0e043e64bf7038e4feefbcd24ca88d46db93d94c/libheif/image-items/image_item.cc#L857-L860

Adding `if (!img->has_nclx_color_profile()) {` around that code resolves the color issues with 2521.heic by using the color profile returned by the decoder. It doesn't resolve the alpha channel issues, but that was still expected.

I can see in the comment that it was an intentional choice to to prefer the nclx profile provided in the HEIF metadata if both are present. I'm not sure if something else would be impacted by making this change.

Given this info do you have any guidance for how you'd like to proceed?

**Comment by bradh** [comment]
> I can see in the comment that it was an intentional choice to to prefer the nclx profile provided in the HEIF metadata if both are present. I'm not sure if something else would be impacted by making this change.

FWIW, this is required by ISO/IEC 14496-12:2022 Section 12.1.5.1:

> If colour information is supplied in both this box, and also in the video bitstream, this box takes precedence, and over-rides the information in the bitstream.

ISO/IEC 23008-12 Section 6.5.5 just points back to 14496-12 for `colr` syntax and semantics, although there are some variations in multiple `colr` boxes are handled.

**Comment by scriby** [comment]
Thanks for that info @bradh.

I'll try to explain why I think it doesn't apply (at least, directly) in this case.

In the 2521.heic file, the colr box and SPS VUI data have matching nclx color profiles. So there's no mismatch in terms of input data and either could be used.

The issue is that the decoder (in this case, device hardware) is for whatever reason changing the color profile when decoding.

So now the "ground truth" about which color profile is being used has changed, and if we use the color profile in the HEIF metadata it will no longer be correct.

Apparently this is the first decoder plugin which exhibits this sort of behavior, and the expectation up until now was that decoders would return color information in the same profile as the input source.

I don't know whether it's a spec violation or whatnot for a decoder to change the color profile when decoding, but there's not much I can do about that and just need to figure out what to do with the output.

**Comment by farindk** [comment]
I think we agree that the browser decoder illegally changes the decoded YCbCr values. It would not matter much if it just returned wrong VUI/nclx parameters, but as you said, it actually changes the decoded pixel values.
Fixing this by introducing a matching error in libheif core by using the decoder's nclx instead of the HEIF `colr` seems like a bad idea. If the decoder is known to modify the output, this has to be compensated by the plugin.

Our real problem is that we do not understand yet what the decoder is actually doing and why it is doing that.
We might have to check the behavior across several browsers and different graphics cards.
Or we could file an issue in the browser repo to clarify what is happening there.

Unfortunately, I didn't find the time yet to run this on my computers. But if you had a version online (e.g. https://strukturag.github.io/libheif/ with the new plugin), it should be easy to collect information on different browser/graphics card behavior.

**Comment by scriby** [comment]
I've pushed an update to the pull request to convert limited range to full range in the plugin if the result from the webcodecs API indicates it's using limited range colors.

It works well in my local testing but I'll see if I can get something hosted so you can see it as well.

**Comment by scriby** [comment]
Alright, I've got a hosted version of this PR up at https://scriby.github.io/libheif/.

I just made one modification to print out a debugging line to make it easier to see what the browser API is returning. Check for something like this below the image:

`libheif webcodecs decoded. format: NV12, full range: false, primaries: bt709, matrix: bt709, transfer: bt709`

**Comment by scriby** [comment]
I tried it out on a few different devices.

Chrome on Mac, Android, and Windows all returned NV12 data with `full range: false`.

The only place I've seen `full range: true` so far is in Safari on both Mac and iOS.

**Update**: I did some more testing on Windows. I found a [HEIC image online](https://heic.digital/download-sample/sewing-threads.heic) for which Chrome will decode with `full range: true`, but my iPhone photos decode as `full range: false`.

In Edge, both the image I found online and my iPhone photos decode as `full range: true`. However, the sample images decode as `full range: false` and also have some glitches / corruption.

Neither Chrome nor Edge on my Windows machine will decode 2521.heic at all (image with alpha channel).

**Comment by farindk** [comment]
I tried your hosted decoder, but only get some errors.

Linux (Ubuntu 24.04.2), Chrome 138.0.7204.183:
image

Linux (Ubuntu 24.04.2), Firefox 144.0:
image

Is there some browser feature that has to be enabled for this?
Maybe the plugin should check whether video decoding support is available and return false from `webcodecs_does_support_format()` if it is not present so that libheif can fallback to libde265.

**Comment by scriby** [comment]
Unfortunately Linux support isn't very good for hardware HEVC decoding through browsers out of the box, but it may be possible with some tweaks.

Here's some references:

https://github.com/brave/brave-browser/issues/37963#issuecomment-2263608335
https://askubuntu.com/questions/1429206/proper-way-to-watch-hevc-on-supported-browser-like-chromium

Thanks for the tip on `webcodecs_does_support_format`, I will work on adding that.

**Comment by scriby** [comment]
Oh, by the way, I set the priority of the webcodecs decoder to be less than libde265.

At least for now, if you have the option to use libde265 I think it's the better choice due to supporting the widest range of possible HEIC files. For instance, HEICs with alpha channels don't decode on Windows devices I tested on.

At this point webcodecs decoder isn't really faster than libde265 (especially w/ SIMD optimizations), although it could be if tile decoding was parallelized. There is a benefit in that webcodecs decoder won't block the Javascript main thread as long, but this could also be avoided by using a web worker.

I see the main motivation of using webcodecs to avoid any potential patent infringement issues with shipping a software HEVC decoder.

My expectation and recommendation is that most people would choose either webcodecs or libde265 for webassembly. The main reason being that if you can use libde265, then there's not currently a strong reason to increase the binary size by also including webcodecs (although maybe there would be if we improved the performance of webcodecs).

**Comment by farindk** [comment]
> For instance, HEICs with alpha channels don't decode on Windows devices I tested on.

Alpha channel images are stored as two independent HEVC images. One for the color channels and one usually monochrome image for the alpha channel. Thus, I don't see a reason they should not decode with the webcodec.
Maybe the decoder does not decode monochrome images?

Could also be that libheif does not expect to get the alpha channel as an RGB image.

**Comment by scriby** [comment]
The error is thrown by the VideoDecoder API in the browser, so I think monochrome support is the issue. I've also seen similar things decoding HEICs on Android devices.

**Comment by scriby** [comment]
I took a look at implementing support for `webcodecs_does_support_format`, and there's a potential issue.

In order to determine whether the image can be decoded by the browser API, we need to know which profile is in use (e.g. hvc1.3.E.L93.B0) and the size of the image. However, the only argument to `webcodecs_does_support_format` doesn't have this information.

The best I can do without any additional information is to use some hardcoded parameters, which is probably better than nothing, but will also have some false positives.

The hardcoded parameters I would suggest are codec of hvc1.3.E.L93.B0 (main still image) and tile size of 512 x 512, which is commonly what I see on modern iPhone photos.

Would you like me to proceed with that solution?

**Comment by scriby** [comment]
Friendly ping on this if you're able to take a look :)

I'm curious about your thoughts in terms of what's needed to get the webcodecs plugin finished up & available in the master branch.

There are a few teams within Google that are waiting on this work to add HEIC support to their products, so I just wanted to check on what we can expect in terms of remaining work and timeline to see if we need to be adjusting any plans internally.

Many thanks again and I appreciate all the time you've devoted to this so far :)

**Comment by farindk** [comment]
> I took a look at implementing support for webcodecs_does_support_format, and there's a potential issue.
> ...
> The hardcoded parameters I would suggest are codec of hvc1.3.E.L93.B0 (main still image) and tile size of 512 x 512

Yes, I also came across this problem. I think, we need a new function in the plugin API for that where we can pass more details on the bitstream to be decoded. It is not easy to make a list of parameters since these will be different for each codec. My current suggestion would be to have a list of key,value pairs that define whatever parameters we know. Then the plugin can match this against its capabilities.

Yes, your proposed hardcoded values are good for a fallback until we have such a function.

I did some more testing of your online page and got these results:

- Windows 11, GeForce RTX 3060
    - Firefox: does not work
    - Chrome: works fine
    - Edge: decodes an image, but output image is corrupted:
Screenshot 2025-10-31 135516

- Windows 11, Surface Pro 9 tablet
    - Edge:  decodes an image, but output image is corrupted (same as above)

- Windows 10, Intel HD Graphics 4600, Haswell CPU
    - Chrome: does not work
    - Edge: does not work, error:
image

- macOS 15.5, Mac mini M2
    - Safari 18.5: decodes, but image is corrupted
Bildschirmfoto 2025-10-31 um 16 32 20

**Comment by farindk** [comment]
I have merged this. There are still many issues which are out of our control and which probably cannot be worked around with the current state of the web-decoders implementation. It seems that the only system where it runs reliably is Chrome on Windows 11. Thus, we should mark this as experimental.

Unfortunately, I could not really help much here because I still could not make it run on my Linux development system.

**Comment by scriby** [comment]
```
const result = await VideoDecoder.isConfigSupported({
    codec: 'hvc1.3.E.L93.B0',
    codedWidth: 512,
    codedHeight: 512,
  });

return !!result.supported;
```

We are using some code like that to check whether the browser likely supports HEVC decoding before attempting to load the libheif webassembly binary. We're seeing that over 80% of organic traffic indicates that it has support.

The most consistent support is actually Chrome on Mac (over 99%). Safari on Mac surprisingly doesn't work very well, but it's ok b/c we can use native HEIC rendering in Safari (so we don't need to use libheif).

I only saw the image corruption in Edge on the sample images provided in the library - I never had any issues with pictures taken from an iPhone.

I agree with the "experimental" designation. It's definitely a "best effort" solution, but also quite a bit better than nothing.

farindk:
> I only saw the image corruption in Edge on the sample images provided in the library - I never had any issues with pictures taken from an iPhone.

The example images were early images encoded with libheif. They are not using tiles, but encode the image in one piece. Still strange to get that extra corrupted data at the bottom. I would assume that the browser API is mainly for video decoding and in that case it will always be in one piece.

scriby:
Given that this is considered experimental, do you plan to leave it in the webcodecs branch? Or are you planning to merge it into the master branch at some point?

farindk:
I will merge it into master asap.

scriby:
Great! Thanks!

farindk:
Merged into the `master` branch.

PR #1608, opened 2025-10-20
Improve documentation on how to use SVT as a builtin encoder

Dinika:
Simple PR to improve documentation on how to use SVT as a builtin encoder (ie not a plugin) with libheif.

Tbh, perhaps the only absolutely necessary change might be [the one in third-party/svt.cmd](https://github.com/strukturag/libheif/pull/1608/files#diff-f0becc742116ba8c1db2d3f2c07c57b65199f1995655ea976bed5bd801205e68R5).
Others are improvements that I think would've been helpful to me (image encoding noob) when I started configuring libheif with SVT and facing issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,16 +256,16 @@ If you want to compile SVT-AV1 yourself,
 * Run the `svt.cmd` script in the `third-party` directory to download SVT-AV1
   and compile it.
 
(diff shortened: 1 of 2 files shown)


farindk:
Thank you.

PR #1613, opened 2025-10-26
Track::alloc_track(): Fix bad null check

hpjansson:
This was causing crashes. Looks like a copy/paste oversight.


farindk:
Thanks.

PR #1616, opened 2025-10-30
Ignore missing / incomplete tiles if not strict decoding.

fancycode:
I though about adding a dedicated flag but probably `strict_decoding` is fine to use for these cases.

Feel free to ping me for example images.

### DIFF
--- a/libheif/image-items/grid.cc
+++ b/libheif/image-items/grid.cc
@@ -290,6 +290,7 @@ Result<std::shared_ptr<HeifPixelImage>> ImageItem_Grid::decode_full_grid_image(c
 
   int progress_counter = 0;
   bool cancelled = false;
+  std::vector<Error> warnings;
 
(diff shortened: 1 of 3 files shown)


farindk:
Looks like this is related to https://github.com/strukturag/libheif/issues/1448
Can you send me some example images? I guess that depending on which tiles are missing, there might be more edge cases not handled yet.

fancycode:
In some cases the image doesn't exist yet, so I'm not sure how/where to store the warnings then.

farindk:
Is this still a draft?
I am considering rewriting the whole grid decoding since the original function was already messy and handling all the new cases makes it even worse.

fancycode:
This is ready for review now. I didn't get any errors in my tests and fuzzing seems to be happy, too.

farindk:
Thanks.

PR #1617, opened 2025-11-01
Sanitize auxiliary type when used as filename suffix

mgenware:
Sanitize the auxiliary type string, especially the colon (:), which is commonly found in HEIC auxiliary types. On macOS, `fopen` can create files containing colons, but they appear as / in Finder. On Windows, `fopen` fails outright when the filename contains :. This change ensures filename compatibility across platforms.


farindk:
Thank you.

PR #1618, opened 2025-11-03
Check "max_output_buffer_entries" when returning warnings.

fancycode:
Otherwise reading only a subset of the warnings is not possible.


farindk:
Thank you

PR #1620, opened 2025-11-04
Forward warnings to output image.

fancycode:
Noticed this while working on #1616. Warnings are no longer present in the output image if this is processed (e.g. rotated) before being returned.


farindk:
thanks

PR #1622, opened 2025-11-04
Wait for decoding jobs to finish before returning.

fancycode:
If an early job returns an error and the method returns, the remaining jobs continue to run and might try to access objects that no longer exist.

https://github.com/strukturag/libheif/blob/f0e803edb05251ae6ec9a1981c938a32df9f0c04/libheif/image-items/grid.cc#L376

Reason for crashes like this:
https://github.com/strukturag/libheif/actions/runs/19065177216/job/54453948767?pr=1616

As it's a race condition even without the changes from #1616 I split it up into a separate PR for review.


farindk:
Thank you.

PR #1628, opened 2025-11-19
Add num_threads to heif_decoding_options

y-guyon:
Add new_decoder_with_options() to heif_decoder_plugin. Implement it for dav1d and aom decoders.

### DIFF
--- a/libheif/api/libheif/heif_decoding.cc
+++ b/libheif/api/libheif/heif_decoding.cc
@@ -43,7 +43,7 @@ int heif_have_decoder_for_format(heif_compression_format format)
 
 static void fill_default_decoding_options(heif_decoding_options& options)
 {
-  options.version = 9;
+  options.version = 10;
 
   options.ignore_transformations = false;
 
@@ -86,6 +86,10 @@ static void fill_default_decoding_options(heif_decoding_options& options)
   // version 9
 
   options.output_image_nclx_profile = nullptr;
+
+  // version 10
+
+  options.num_threads = 0;
 }
 
 
@@ -110,6 +114,9 @@ void heif_decoding_options_copy(heif_decoding_options* dst,
   int min_version = std::min(dst->version, src->version);
 
   switch (min_version) {
+    case 10:
+      dst->num_threads = src->num_threads;
+      [[fallthrough]];
     case 9:
       dst->output_image_nclx_profile = src->output_image_nclx_profile;
       [[fallthrough]];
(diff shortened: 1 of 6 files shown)


farindk:
Thanks, Yannis.

I'm wondering what advantage it would have to slow down the decoding.
Do you see any use case for such an option for a typical end user except your benchmarking?

I'll have to do some changes to the PR as I'd like to avoid passing `heif_decoding_options` through to the plugin.
Most of these options are irrelevant to the codec while there will be more options that cannot be provided by the user (bitstream subtypes). Thus, I would like to separate those. But you can leave that to me as I am going to do some work on the decoder plugin interface anyway.

y-guyon:
> I'm wondering what advantage it would have to slow down the decoding. Do you see any use case for such an option for a typical end user except your benchmarking?

Being able to set the number of threads is a common feature. Sometimes you may want to tune that to use more or fewer threads than the default option (which is currently on the codec side, so it may not even be consistent across all libheif codecs).

Use cases I can think of:
- backend decoding where there are more assets than cores, and setting more than one thread per decoding is at best useless and at worst harmful to performances,
- reserving some cores for some jobs on the machine, ending in a different count of hardware cores and actual useful threads running at the same time,
- leveraging more threads than cores because of slow I/O operations throttling some threads.

> I'll have to do some changes to the PR as I'd like to avoid passing `heif_decoding_options` through to the plugin. Most of these options are irrelevant to the codec while there will be more options that cannot be provided by the user (bitstream subtypes). Thus, I would like to separate those. But you can leave that to me as I am going to do some work on the decoder plugin interface anyway.

Sure, thanks.

farindk:
> Use cases I can think of:

Ok, that makes sense. Probably even within libheif we could switch off codec multithreading when tiles decoding is parallelized already at the HEIF level.

farindk:
Thanks again for your PR. While I was working on a new plugin interface for supporting image sequences, I also added an option for the number of decoding threads (separate setting for codec threads and libheif threads). I've implemented this in 3273db413d326ec6e7fc58830276becb15b62ffa and 5319b363ca5722cce43c9d6da1f071696216a80d for all plugins that support it. (Note: it is still in the `inter-sequences-encoding` development branch.)

y-guyon:
> Thanks again for your PR. While I was working on a new plugin interface for supporting image sequences, I also added an option for the number of decoding threads (separate setting for codec threads and libheif threads). I've implemented this in [3273db4](https://github.com/strukturag/libheif/commit/3273db413d326ec6e7fc58830276becb15b62ffa) and [5319b36](https://github.com/strukturag/libheif/commit/5319b363ca5722cce43c9d6da1f071696216a80d) for all plugins that support it. (Note: it is still in the `inter-sequences-encoding` development branch.)

Thank you for the notice. Please let me know when codec and libheif thread settings are available in a tagged libheif release.

PR #1650, opened 2025-12-29
Support Other Uncompressed Component Formats

dukesook:
Using the experimental `heif_image_add_channel()`, the user can specify the `heif_channel_datatype`. This PR writes that value into the uncC boxes `component_format` field for monochrome images.

TODO: Repeat for RGB & YUV.


farindk:
Thanks. I'll merge this later for the v1.22.0 release which focuses on the uncompressed codec.

PR #1652, opened 2025-12-30
Add climits includes

heitbaum:
Fixes:
../libheif/api/libheif/heif_image_handle.cc: In function 'int heif_image_handle_get_width(const heif_image_handle*)': ../libheif/api/libheif/heif_image_handle.cc:48:13: error: 'INT_MAX' was not declared in this scope
   48 |     if (w > INT_MAX) {
      |             ^~~~~~~
../libheif/api/libheif/heif_image_handle.cc:23:1: note: 'INT_MAX' is defined in header '<climits>'; this is probably fixable by adding '#include <climits>'
   22 | #include "api_structs.h"
  +++ |+#include <climits>
   23 | #include <string>

PR #1655, opened 2025-12-30
Fix: added <cstdio>

Jamaika1:
```
sdl.cc: In member function 'bool SDL_YUV_Display::init(int, int, SDL_Chroma, const char*)':
sdl.cc:43:5: error: 'printf' was not declared in this scope
   43 |     printf("SDL_Init() failed: %s\n", SDL_GetError( ) );
      |     ^~~~~~
sdl.cc:29:1: note: 'printf' is defined in header '<cstdio>'; this is probably fixable by adding '#include <cstdio>'
   28 | #include <assert.h>
  +++ |+#include <cstdio>
```

### DIFF
--- a/examples/sdl.cc
+++ b/examples/sdl.cc
@@ -25,7 +25,7 @@
 */
 
 #include "sdl.hh"
-#include <assert.h>
+#include <cassert>
 #include <cstdio>
 
 


kmilos:
Is this also w/ upcoming GCC 16 like https://github.com/strukturag/libheif/pull/1652?

farindk:
@Jamaika1 @kmilos Thank you. I've fixed both issues by removing `printf()` altogether and properly returning the error message. The `assert.h` was not needed and has been removed.

PR #1673, opened 2026-01-16
Ignore alpha in yuv2rgb and rgb2rgb if aplpha has different size or BPP

DarthSim:
Hey there 👋

Our users sent us a few images where the alpha channel's BPP doesn't match the other channels' BPP. libheif throws an error in such cases, yet Apple Preview and Photoshop seem to just ignore mismatched alpha channels. This PR makes libheif ignore the alpha channel instead of throwing an error.

### DIFF
--- a/libheif/color-conversion/rgb2rgb.cc
+++ b/libheif/color-conversion/rgb2rgb.cc
@@ -81,9 +81,8 @@ Op_RGB_to_RGB24_32::convert_colorspace(const std::shared_ptr<const HeifPixelImag
     return Error::InternalError;
   }
 
-  if (has_alpha && input->get_bits_per_pixel(heif_channel_Alpha) != 8) {
-    return Error::InternalError;
-  }
+  // Check if alpha channel is valid. If not, just ignore it.
+  has_alpha = has_alpha && input->get_bits_per_pixel(heif_channel_Alpha) == 8;
 
   auto outimg = std::make_shared<HeifPixelImage>();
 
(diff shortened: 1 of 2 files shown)


farindk:
Thanks. Do you have an example image that you could share (privately by email if necessary)?

Silently dropping alpha may be very confusing to users and may lead to hard to debug problems. The better approach might be to have a color conversion step that adjusts the bit-depths of the alpha channel to match the output format.

DarthSim:
Unfortunately, I don't think I can share them 😞 The images were shared privately with us by our users and are owned by our users' users. I couldn't find any of such samples in the web either. I'll try to craft something.

> Silently dropping alpha may be very confusing to users and may lead to hard to debug problems

Absolutely valid point. However, we were patching libheif like this for two years with no claims from the users 🙂

> The better approach might be to have a color conversion step that adjusts the bit-depths of the alpha channel to match the output format.

Agree. If you could point out the place where it's better to put this conversion, I could try to implement it.

farindk:
The alpha channel bpp is now tracked through the color-conversion pipeline. I also have implemented a color-conversion function that converts the alpha channel bpp if required. This should now process the images without dropping the alpha channel.

Can you please confirm that this works with your image?

DarthSim:
Hey @farindk!

That works perfectly, thanks! 👍

I started working on this, but was distracted by other stuff 😅 And your solution is definitely more elegant than mine.

PR #1677, opened 2026-01-21
Print kvazaar version if such function is available

agruzdev:
The latest kvazaar version now supports getting version string in runtime. 
So I added version information to the kvazaar plugin, when libheif is compiled with the latest kvazaar version. Support of the version function is cheched in cmake.

image


farindk:
Thank you, this is good to have.

PR #1678, opened 2026-01-23
Update encoder_kvazaar.cc

kmilos:
@agruzdev @farindk  A follow-up to #1677 

Has to be ifdef since `-DHAVE_KVAZAAR_VERSION_STRING=0` is not defined.

### DIFF
--- a/libheif/plugins/encoder_kvazaar.cc
+++ b/libheif/plugins/encoder_kvazaar.cc
@@ -102,7 +102,7 @@ static void kvazaar_set_default_parameters(void* encoder);
 
 static const char* kvazaar_plugin_name()
 {
-#if HAVE_KVAZAAR_VERSION_STRING
+#ifdef HAVE_KVAZAAR_VERSION_STRING
   snprintf(plugin_name, MAX_PLUGIN_NAME_LENGTH, "kvazaar HEVC encoder %s", kvz_get_version_string());
 #else
   strcpy(plugin_name, "kvazaar HEVC encoder");


agruzdev:
@kmilos Hello,
It is not needed. Any undefined name in #if is treated as 0

https://en.cppreference.com/w/cpp/preprocessor/conditional.html 

> After all macro expansion and evaluation defined and the expressions described above, any identifier which is not a [boolean literal](https://en.cppreference.com/w/cpp/language/bool_literal.html) is replaced with the number ​0​ (this includes identifiers that are lexically keywords, but not alternative tokens like and).

agruzdev:
Also, #ifdef doesn't allow explicitly disable condition, for example by passing -DHAVE_KVAZAAR_VERSION_STRING=0 from cmake. Even in this case #ifdef will be true, while #if will be false

kmilos:
Nonetheless (I can't bet 100% of preprocessors follow this), semantically I prefer the "write what you mean" coding style... I'll let Dirk close this as he prefers.

kmilos:
> Even in this case #ifdef will be true, while #if will be false

You'll find this paradigm is typically coded explicitly as `#if defined(FOO) && (FOO==1)` for clarity.

Anyhow, we're into coding style preferences now, which is pointless to discuss.

agruzdev:
Then use expression `#if defined(HAVE_KVAZAAR_VERSION_STRING) && (DHAVE_KVAZAAR_VERSION_STRING != 0)`
But if you dont belive standard 

But this is standard behaviour
image


https://eel.is/c++draft/cpp.cond

So.. if you dont believe standard, it's hard to write code :)

farindk:
Let's explicitly set this to '0' ([3d0b151](https://github.com/strukturag/libheif/commit/3d0b151ef832a0cd0dc3dc8491c4eebec5a5c68f)). Then we are consistent with how we did with `HAVE_KVAZAAR_ENABLE_LOGGING`.

https://github.com/strukturag/libheif/blob/3d0b151ef832a0cd0dc3dc8491c4eebec5a5c68f/CMakeLists.txt#L143-L152

PR #1679, opened 2026-01-23
Support unicode paths on windows (#1358)

mgenware:
Fixes https://github.com/strukturag/libheif/issues/1358

PR #1683, opened 2026-01-26
Fix passing compiler options to preprocessor

kmilos:
x-ref https://github.com/strukturag/libheif/pull/1679#issuecomment-3799791952


farindk:
I just tried to solve the same with this approach
````cmake
if(NOT MSVC)
  set(LIBHEIF_WARNING_FLAGS
    -Wall
    -Wsign-compare
    -Wconversion
    -Wno-sign-conversion
    -Wno-error=conversion
    -Wno-error=unused-parameter
    -Wno-error=deprecated-declarations
  )

  # cmake 3.24 introduces this variable, but for backward compatibility, we set it explicitly
  if (CMAKE_COMPILE_WARNING_AS_ERROR)
    list(APPEND LIBHEIF_WARNING_FLAGS -Werror)
  endif()

  if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
      list(APPEND LIBHEIF_WARNING_FLAGS
          -Wno-error=tautological-compare
          -Wno-error=tautological-constant-out-of-range-compare
      )
  endif()

  # Apply warning flags only to C and C++ (ignoring RC/windres)
  add_compile_options("$<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:${LIBHEIF_WARNING_FLAGS}>")
endif()
````

But yours seems to work too and is simpler.

PR #1687, opened 2026-01-31
Update heif_enc.cc

rjksmith:
Updated VMT timestamp parser


farindk:
Thanks, I have merged it.

rjksmith:
These changes add the following features:
 * optional hours in VMT timestamps
 * optional milliseconds in VMT timestamps
 * optional end times in VMT cues
 * negative cue times are set to zero
 * VMT cues with identical start times are appended to the previous metadata

PR #1690, opened 2026-02-02
Fix: Set dav1d thread count before opening decoder

Dinika:
The `num_threads` setting was being applied after dav1d_open(), which meant the thread configuration was ignored. Moving the thread count assignment to occur **before** opening the decoder (just like it's done in other decoders like [decoder_aom](https://github.com/strukturag/libheif/blob/master/libheif/plugins/decoder_aom.cc#L119-L126)) so that the setting takes effect.


farindk:
Thank you

PR #1692, opened 2026-02-04
Bugfix: Set Primary Item When Using unci API

dukesook:
Previously, when calling `heif_context_add_empty_unci_image()` the primary item primary item would be set to 0.


farindk:
Thanks

PR #1693, opened 2026-02-05
Input Validation: assert tile coordinates must be valid

dukesook:
Throws an error if the caller tries to add a tile outside the range of the image.


farindk:
Thanks.
Note: `heif_error_Invalid_input` is for reading input files, `heif_error_Usage_error` is for wrong API parameters.

Issue #1241, opened 2024-07-11
HEIF (10 bit) to AVIF (10 bit) conversion, e.g., for Android compatibility


paswai:
Hi, I am a Canon user and my camera produces 10 bit HIF files. The use of these files is (currently) heavily limited on Android, e.g., Google Photos, WhatsApp,... don't support it. I checked and tested that AVIF (10 bit) is however supportet by these apps. Might a conversion feature from HEIF to AVIF be of interest to other users?

Here a HIF file example: [example](https://github.com/user-attachments/files/15759323/20240525_162308_IMG_0009.zip)

farindk:
In v1.22.0 (just released), you can directly convert between HEIF and AVIF.

PR #1697, opened 2026-02-10
Add support for decoding PNG cICP, cLLI and mDCV chunks

torusrxxx:
It adds decoding support for PNG `cICP`, `cLLI` and `mDCV` chunks. It works with the `--color-profile auto` parameter of `heif-enc` tool to pass HDR metadata to HEIC encoder. Together with #1695, it enables simple round-trip conversion of HDR AVIF-PNG-HEIC files using reference tools. Tested, it produces HDR AVIF, PNG and HEIC files that look the same.

### DIFF
--- a/heifio/decoder_png.cc
+++ b/heifio/decoder_png.cc
@@ -30,6 +30,7 @@
 #include <iostream>
 #include <memory>
 #include "decoder_png.h"
+#include "common_utils.h"
 #include "exif.h"
 
 extern "C" {
@@ -75,6 +76,19 @@ heif_error loadPNG(const char* filename, int output_bit_depth, InputImage *input
 #endif
   uint8_t* profile_data = nullptr;
   png_uint_32 profile_length = 5;
+#ifdef PNG_cICP_SUPPORTED
+  uint8_t color_primaries, transfer_characteristics, matrix_coefficients, video_full_range;
+  bool has_cICP = false;
+#endif
+#ifdef PNG_cLLI_SUPPORTED
+  uint32_t maximum_content_light_level_scaled_by_10000, maximum_frame_average_light_level_scaled_by_10000;
+  bool has_cLL = false;
+#endif
+#ifdef PNG_mDCV_SUPPORTED
+  png_fixed_point white_point_x, white_point_y, display_primaries_x[3], display_primaries_y[3];
+  png_uint_32 max_display_mastering_luminance, min_display_mastering_luminance;
+  bool has_mDCV = false;
+#endif
 
   /* Create and initialize the png_struct with the desired error handler
    * functions.  If you want to use the default stderr and longjump method,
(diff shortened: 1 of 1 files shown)


torusrxxx:
Partially fixes #1241
```
heif-dec test.heic test.png
heif-enc -b 10 --color-profile auto test.png test2.avif
```
Preserves HDR.

heifio/decoder_png.cc
```diff
@@ -419,6 +465,65 @@ heif_error loadPNG(const char* filename, int output_bit_depth, InputImage *input
...
+          case heif_color_primaries_ITU_R_BT_601_6:
+            matrix_coefficients = heif_matrix_coefficients_ITU_R_BT_601_6;
+            break;
+          case heif_color_primaries_SMPTE_240M:
+            matrix_coefficients = heif_matrix_coefficients_SMPTE_240M;
+            break;
+          case heif_color_primaries_ITU_R_BT_2020_2_and_2100_0:
+            matrix_coefficients = heif_matrix_coefficients_ITU_R_BT_2020_2_constant_luminance;
+            break;
+          }
```
farindk, line 492: I don't understand this. Why don't we simply set `matrix_coefficients` to the value we've read from the PNG?

torusrxxx:
PNG always use matrix coefficients 0 (RGB). HEIC files don't use the RGB
matrix coefficients. Specifically, there would be an error coming from aom
when trying to encode AVIF with RGB matrix coefficients. Here the suitable
matrix coefficients are chosen to match color primaries, and there will be
a color space conversion.

Dirk Farin ***@***.***> 於 2026年4月13日週一 上午6:01寫道：

farindk:
Thank you for the explanation. That makes sense.
Maybe it would be even better if we moved setting the matrix coefficients into libheif itself as defaults when matrix_coefficients are "undefined" but color_primaries are set, but I'll leave that for the future since it will also be a breaking change.

kmilos:
Is the RGB->YCbCr conversion happening _always_, even for "lossless" modes? (libaom issue/limitation aside...)

farindk:
> Is the RGB->YCbCr conversion happening _always_, even for "lossless" modes? (libaom issue/limitation aside...)

If you set matrix_coefficients=0, there will be no RGB->YCbCr conversion. The values that are set here in the PNG reader are just the defaults, but it can be overridden by `heif_encoding_options::output_nclx_options`.

torusrxxx:
Thank you. I looked into the matrix coefficients choice here. It seems
nobody is using the ITU_R_BT_2020_2_constant_luminance as the default,
maybe it should be changed into more commonly used
 heif_matrix_coefficients_ITU_R_BT_2020_2_non_constant_luminance?

farindk:
> maybe it should be changed into more commonly used heif_matrix_coefficients_ITU_R_BT_2020_2_non_constant_luminance?

Yes, I have changed that and I tried to extend the table a bit.

matemaciek:
Can't wait for all the necessary release propagations to see whether it helps for #1241 (Canon HEIFs in Android et al). 🤞

paswai:
Hi folks, many thanks for the feature update! The conversion of HIF to avif produces files viewable on Android :-). It seems that the conversion does not preserve the colors (it looks desaturated). I could not find parameters to correct this behaviour. Below is some color information of the input and output file :

> exiftool test.hif | grep -iE "primaries|transfer|matrix"
> Color Primaries                 : BT.2020, BT.2100
> Transfer Characteristics        : SMPTE ST 2084, ITU BT.2100 PQ
> Matrix Coefficients             : BT.2020 non-constant luminance, BT.2100 YCbCr

> exiftool test2.avif | grep -iE "primaries|transfer|matrix"
> Color Primaries                 : BT.709
> Transfer Characteristics        : sRGB or sYCC
> Matrix Coefficients             : BT.601

Maybe this was meant in the discussion between @torusrxxx and @farindk above which I could not follow. 

I attached a [HIF example file](https://github.com/user-attachments/files/28198199/test.zip).

PR #1698, opened 2026-02-13
Optional Primary Item

dukesook:
Sometimes, we want a meta box with no primary item.

For example, a GIMI sequence may have a track and a file-level metabox with a single item (the security markings as a mime item). I suppose you could mark it as the primary item, but it's not intended for display.

The current behavior of libheif writes a 'pitm' to item id 0, which is always invalid and leads to parsing errors.

ISO/IEC 23001-17 8.11.4 indicates that the 'pitm' box is NOT mandatory, but I'm not sure if any particular brand requires it.

### DIFF
--- a/libheif/file.cc
+++ b/libheif/file.cc
@@ -176,10 +176,8 @@ void HeifFile::init_for_image()
   m_iloc_box = std::make_shared<Box_iloc>();
   m_iinf_box = std::make_shared<Box_iinf>();
   m_iprp_box = std::make_shared<Box_iprp>();
-  m_pitm_box = std::make_shared<Box_pitm>();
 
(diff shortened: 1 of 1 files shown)


farindk:
Yes, I understand the issue. I'll have to look closer into it because these are all things that were never considered in the initial design. I am currently just wondering why `init_for_images()` is called when only a sequence is added.

Do you have a code snippet where you add the security-marking item into the `heif_context` ?

dukesook:
I wrote some custom software that simply calls `heif_context_add_mime_item()` to add the security markings. It's essentially just that one line of code.

farindk:
I think this is now already handled by 248c1de0.
Can we close this?

dukesook:
This looks great, thank you.

Issue #1686, opened 2026-01-27
Sticker created by iPhone is rendering as four repeated sections


charleswcho:
This image is a screenshot of a heic image that an iPhone created
<img width="354" height="337" alt="Image" src="https://github.com/user-attachments/assets/56b0496c-b8a3-4258-b410-58227f951f92" />

This is how it is rendered after libheif processing using the webcodecs decoder
<img width="1065" height="447" alt="Image" src="https://github.com/user-attachments/assets/faabb376-0995-49fc-9181-3d01d6fea5c2" />

@scriby

farindk:
Can you also attach the HEIC file?

charleswcho:
Ironically Github doesn't support attaching HEIC images.

<img width="548" height="39" alt="Image" src="https://github.com/user-attachments/assets/87adc661-016c-4c6f-9fed-9a585b609369" />

https://drive.google.com/file/d/1bhu2DroyOhautSN_Jm7_KW8T-YTKNemW/view

bradh:
> Ironically Github doesn't support attaching HEIC images.

It does support attaching zip files - that is the usual workaround.

farindk:
I tested it with v1.21.2 and v1.19.8 and the image decoded fine on the command line (no webcodec decoder).
Thus, it is a webcodecs problem.
And the output looks consistent to what I have got with Safari:
https://github.com/strukturag/libheif/pull/1606#issuecomment-3473576338

charleswcho:
Hi, can someone take a look at the PR @scriby created, thanks!

farindk:
Please close if it works for you. I cannot test this.

PR #1703, opened 2026-02-19
Add support for I420, I422, and I444 pixel formats in the webcodecs decoder.

scriby:
Fixes issue #1686 related to decoding images produced by copying & pasting subjects out of images on iOS devices, which are created with 4:4:4 chroma.

I was able to successfully decode the attached deer image, as well as some similar images I created on my iPhone. Also, tested normal iPhone photos to ensure they still can be decoded.

This CL also refactors code a bit to share more code between NV12 and new I420, I422, and I444 code paths.


farindk:
Thanks. I cannot test this myself, but it looks fine.

PR #1707, opened 2026-02-22
feat: add support for omnidirectional (OMAF) image projection

bradh:
This is defined in ISO/IEC 23090-2.

There is a test sample at https://github.com/nokiatech/omaf/raw/refs/heads/master/Player/VideoPlayback/Android/Handheld/Assets/omaf_360.heic (29.9MB)

As far as I can tell, the only thing that is required is `prfr` which is the projection format (equirectangular or cube-map) - its an essential property, so we need to handle it. Very basic structure though. There are a few other properties in OMAF that we could add as a follow-up.

This PR includes basic API to expose the projection format on both reading and writing. That is worth more detailed consideration - I'm not convinced that its sufficiently consistent with other libheif API, but its usable.

I implemented this with an option flag, defaulting to ON. Since OMAF isn't very common, some users may not want to include support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,8 @@ include (TestBigEndian)
 TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
 add_compile_definitions(IS_BIG_ENDIAN=${IS_BIG_ENDIAN})
 
+option(HEIF_WITH_OMAF "Enable omnidirectional media format (OMAF) support." ON)
+
 # --- codec plugins
 
 option(ENABLE_PLUGIN_LOADING "Support loading of plugins" ON)
@@ -528,6 +530,12 @@ if(HAVE_LIBCPP)
   set(LIBS_PRIVATE "-lc++")
 else()
   set(LIBS_PRIVATE "-lstdc++")
+endif()
+
+if(HEIF_WITH_OMAF)
+  set(WITH_OMAF "1")
+else ()
+  set(WITH_OMAF "0")
 endif()
 
 configure_file(libheif.pc.in ${CMAKE_CURRENT_BINARY_DIR}/libheif.pc @ONLY)
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -25,6 +25,7 @@
         "WITH_DAV1D" : "ON",
         "WITH_DAV1D_PLUGIN" : "OFF",
         "ENABLE_EXPERIMENTAL_MINI_FORMAT" : "ON",
+        "HEIF_WITH_OMAF" : "ON",
         "WITH_LIBDE265" : "ON",
         "WITH_LIBDE265_PLUGIN" : "OFF",
         "WITH_RAV1E" : "ON",
@@ -75,6 +76,7 @@
         "BUILD_SHARED_LIBS": "ON",
         "BUILD_TESTING" : "OFF",
         "ENABLE_EXPERIMENTAL_FEATURES" : "OFF",
+        "HEIF_WITH_OMAF" : "ON",
         "CMAKE_COMPILE_WARNING_AS_ERROR" : "OFF",
 
         "ENABLE_PLUGIN_LOADING" : "ON",
@@ -131,6 +133,7 @@
         "BUILD_SHARED_LIBS": "ON",
         "BUILD_TESTING" : "OFF",
         "ENABLE_EXPERIMENTAL_FEATURES" : "OFF",
+        "HEIF_WITH_OMAF" : "ON",
         "CMAKE_COMPILE_WARNING_AS_ERROR" : "OFF",
 
         "ENABLE_PLUGIN_LOADING" : "OFF",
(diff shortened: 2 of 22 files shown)


farindk:
Thanks, Brad. Two minor things that I consider changing:
- adding the `omaf` prefix to make it clearer what this refers to. For example: `heif_image_projection` -> `heif_omaf_image_projection`
- in `heif-enc` use the textual names (e.g. `cubemap`) instead of just 0, 1.

If you think this is ok, I can do the changes. Otherwise everything is fine.

bradh:
>     * adding the `omaf` prefix to make it clearer what this refers to. For example: `heif_image_projection` -> `heif_omaf_image_projection`

Maybe for this one.  OMAF doesn't really have a value corresponding to flat, so it is more than just "omaf projection". 

image

So I'm extending what OMAF offers to make sure the property returns something sensible. That said, the general concept does come from OMAF. 

Whatever makes more sense to you is OK for me.

>     * in `heif-enc` use the textual names (e.g. `cubemap`) instead of just 0, 1.

Yes. This would definitely be an improvement.

> If you think this is ok, I can do the changes. Otherwise everything is fine.

I can do them next weekend if you have other tasks. If you'd like to do it before that, I'm definitely good with that too.

farindk:
I think the extra `flat` enum constant is no problem. It is clearly described and outside the reserved range.

Did you find any other OMAF images? I saw that another box `rwpk` is needed for cube-map images, but I did not find any samples. I found this [software for encoding](https://github.com/fraunhoferhhi/omaf.js/blob/master/omaf-file-creation/README.md), but it seems to generate video only.

bradh:
> Did you find any other OMAF images? I saw that another box `rwpk` is needed for cube-map images, but I did not find any samples. I found this [software for encoding](https://github.com/fraunhoferhhi/omaf.js/blob/master/omaf-file-creation/README.md), but it seems to generate video only.

I don't know of any other public images.

ISO/IEC 23090-17:2021 (which is the reference software and conformance tests for OMAF Ed 1), says:

> A set of test vectors for conformance are available at https://standards.iso.org/iso-iec/23090/-17/ed-1/en/

It then goes on to describe what each of the 16 test vectors contain. Three of those test vectors (four files) are image profile. 

However when I downloaded the zip package, I couldn't find any of the referenced files. Possibly they are generated (somehow) by the software package, but it doesn't compile for me, and a quick grep of the code didn't suggest any way it could generate the files.

farindk:
Thanks for the link. I also cannot see any support for images there. In `include/omaf.h`, there is a `OMAF_HEVC_Image_Brand`, but it is not used anywhere in the code.

bradh:
I subsequently got the omafimage code working (see https://github.com/nokiatech/omaf/wiki/Usage-instructions-for-OMAF-Creator-Still-Image-file-creation).

Some basic examples are attached.
[equirectangular_omaf_samples.zip](https://github.com/user-attachments/files/25620224/equirectangular_omaf_samples.zip)

I'm not sure it can produce cubemap images, but will keep digging.

bradh:
I have now managed to produce a top-bottom stereo output, but I'm confident that the `stvi` box generation is wrong. So that will take a bit more work.

farindk:
The Nokia software also seems to support equirectangular images only. That code is also unmaintained for years now. I am not sure if that OMAF standard is going anywhere. I am not aware of anything but a few (research) demos. There seem to be some ways to encode stereo images, but there is also an (alternative) way in HEIF itself.

Do you have any particular application in mind or is this just out of interest?
I have a Ricoh Theta 360° camera here which captures only JPEGs, but I think there is some kind of custom markers in there to indicate equirectangular. So in theory, we could read that out in the JPEG reader and generate the `prfr` box accordingly.

bradh:
> The Nokia software also seems to support equirectangular images only. That code is also unmaintained for years now. I am not sure if that OMAF standard is going anywhere. I am not aware of anything but a few (research) demos. There seem to be some ways to encode stereo images, but there is also an (alternative) way in HEIF itself.

I also don't see a lot of use at this stage. Stereo OMAF is looking particularly niche.

> Do you have any particular application in mind or is this just out of interest?

The application is sequences (but probably not image sequences in the HEIF sense) of images, of roads and walking tracks. Conceptually like the google streetview application. 

So I'm not very interested in stereo for this application.

The other HEIF capabilities (and possibly some of the GIMI capabilities) are more useful.

> I have a Ricoh Theta 360° camera here which captures only JPEGs, but I think there is some kind of custom markers in there to indicate equirectangular. So in theory, we could read that out in the JPEG reader and generate the `prfr` box accordingly.

I think there are two ways to guess 360 equirectangular - the EXIF box to get the camera make and model or sensor width / sensor height , and image width is twice the height.
Example of that from Panoramax: https://gitlab.com/panoramax/server/geo-picture-tag-reader/-/blob/develop/geopic_tag_reader/camera.py?ref_type=heads#L124-156

The projection format property looks a lot cleaner than guesswork...

farindk:
Right, the info is also in the JPEG's EXIF:
````
XMP Toolkit                     : RICOH THETA SC Ver1.00
Projection Type                 : equirectangular
Image Size                      : 5376x2688
````

Issue #1350, opened 2024-10-28
OSS-Fuzz build does not find codec dependencies


farindk:
OSS-Fuzz builds libaom, x265, and libde265, but does not find them when building libheif, resulting in only enabling the 'uncompressed' codec (mentioned in https://github.com/strukturag/libheif/issues/1345#issuecomment-2441552780). [Link to Build Logs.](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libheif)

Additionally to fixing this, we should also enable the other codecs in the build.

bobfriesenhahn:
The GraphicsMagick oss-fuzz build is doing much better regarding libheif.  I think that currently only one useful dependency is excluded (because a pkg-config file is now missing).

Feel free to borrow ideas from GraphicsMagick's oss-fuzz support at https://foss.heptapod.net/graphicsmagick/graphicsmagick/-/tree/branch/default/fuzzing?ref_type=heads, and particularly the "oss-fuzz-build.sh" script at https://foss.heptapod.net/graphicsmagick/graphicsmagick/-/blob/branch/default/fuzzing/oss-fuzz-build.sh?ref_type=heads

If you see something that I should be doing better, please make me aware of it.

bobfriesenhahn:
The GraphicsMagick oss-fuzz build does not actually need x265 yet, but I tried to re-add it to the oss-fuzz build while using https://bitbucket.org/multicoreware/x265_git/src/stable/ as the source repository path (as libheif's oss-fuzz build does).

The build failed while building libheif with this error:

```
[ 81%] Building CXX object libheif/CMakeFiles/heif.dir/plugins/encoder_mask.cc.o
[...]

PR #1714, opened 2026-02-24
Add script to build oss-fuzz fuzzers.

fancycode:
Moved here from the build script in oss-fuzz so it can be easily updated from libheif in the future without having to create lots of PRs inside the oss-fuzz repository.

@farindk this is what we discussed yesterday in prepration of #1350. I'll create the PR in oss-fuzz once this has landed.


fancycode:
No changes to the actual building yet, the idea was to get the current state of building for oss-fuzz first and then iterate from there.

fancycode:
oss-fuzz PR is here: https://github.com/google/oss-fuzz/pull/15015

PR #1732, opened 2026-03-16
AOM/Kvazaar/Rav1e/SVT/uvg266/vvenc: fix memory leak when encoder is reused for alpha encoding

DarthSim:
Hey there 👋

Most encoders initialize internal encoders/codecs in `*_start_sequence_encoding`, but destroy them only when they are destroyed. When such an encoder is reused for alpha encoding, a new internal encoder/codec is initialized without destroying the old one, leading to a memory leak and sometimes a thread leak.

This PR adds internal encoder/codec destruction at the start of `*_start_sequence_encoding`.

I checked this fix only for AOM and Kvazaar, since we link libheif against them only.

### DIFF
--- a/libheif/plugins/encoder_aom.cc
+++ b/libheif/plugins/encoder_aom.cc
@@ -865,6 +865,10 @@ static heif_error aom_start_sequence_encoding_intern(void* encoder_raw, const he
 {
   encoder_struct_aom* encoder = (encoder_struct_aom*) encoder_raw;
 
+  // destroy the codec in case it was already initialized
+  // (e.g. when the encoder is reused for alpha encoding after being used for YUV encoding)
+  aom_codec_destroy(&encoder->codec);
+
   heif_error err;
 
   const int source_width = heif_image_get_width(image, heif_channel_Y);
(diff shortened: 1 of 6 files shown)


DarthSim:
@farindk Could you take a look? The leak is quite nasty

dloebl:
We are also affected by this and can confirm that the fix proposed by @DarthSim works well. We'd love to get this in an official upstream libheif release.

farindk:
Sorry for the delay. I have fixed this a bit differently in the patch above.
The patch is creating a new encoder instance for the alpha channel. This has the advantage that we can use a different codec or different parameters for the alpha channel in the future.

Please let me know if you still have any issues with this change.

DarthSim:
Hey @farindk!

Thanks for the fix! I can confirm that the issue is resolved 👍

However, I'd like my fix merged as well, if you don't mind. While your fix eliminates encoder reuse, my fix prevents the leak even when an encoder is reused, eliminating the possibility of making the same mistake again.

farindk:
Hi @DarthSim, I'd prefer to leave the cleanup out because:

- it hides misuse. The intent is now "one encoder, one use." I'd rather have a stray reuse fail loudly (via an assert) than be silently papered over.
- it locks in the wrong choice for future reuse. If we ever support reuse, we'd want to reset/reconfigure rather than destroy+reinit. I imagine that reinitializing thread pools or even a hardware encoder is costly. We should then reuse as much as possible.

DarthSim:
Fair enough. Thanks for looking into this!

farindk:
You fix was better. Mine does not work with grid images, which also use the same encoder for several images.
In the future, we might introduce an explicit "reset" function in the plugins.

PR #1745, opened 2026-03-26
Update decoder_webcodecs.cc to plugin API v5

scriby:
Also, fix references to NAL unit constants (renamed HEVC_...).

Update code to fix use-after-free errors that came up after moving to the v5 API.

### DIFF
--- a/libheif/plugins/decoder_webcodecs.cc
+++ b/libheif/plugins/decoder_webcodecs.cc
@@ -37,8 +37,7 @@
 
 
 struct NALUnit {
-  void* data;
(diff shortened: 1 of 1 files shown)


scriby:
I would appreciate if you could take this update before cutting the next release, so that our code won't break internally when we update.

Thanks!

farindk:
Thanks.

The new API is mainly used for decoding image sequences. Have you tried this in your application? I assume it won't work because it's not passing through the frame numbers (the user data pointer), for example.
But if it is used only for images, it should still work.

PR #1748, opened 2026-04-01
Support long paths on Windows

kmilos:
See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#application-manifest-updates-to-declare-long-path-capability

Note that it might require a registry change to actually enable long paths on the system.


farindk:
Thanks

PR #1763, opened 2026-04-12
box: add bounds check when sorting ipma properties

lovell:
This is currently only used by the encoder but good to protect against a self-inflicted denial-of-service should the number of `associations` be zero.

### DIFF
--- a/libheif/box.cc
+++ b/libheif/box.cc
@@ -3366,14 +3366,23 @@ void Box_ipma::sort_properties(const std::shared_ptr<Box_ipco>& ipco)
 
   for (auto& item : m_entries) {
     size_t nAssoc = item.associations.size();
+    if (nAssoc < 2) {
+      continue;
+    }
 
     // simple Bubble sort as a stable sorting algorithm
 
     for (size_t n = 0; n < nAssoc - 1; n++)
       for (size_t i = 0; i < nAssoc - 1 - n; i++) {
+        uint16_t idx_a = item.associations[i].property_index;
+        uint16_t idx_b = item.associations[i + 1].property_index;
+        if (idx_a == 0 || idx_a > properties.size() ||
+            idx_b == 0 || idx_b > properties.size()) {
+          continue;
+        }
         // If transformative property precedes descriptive property, swap them.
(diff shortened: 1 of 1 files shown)


lovell:
The fuzz test failure appears to be genuine but unrelated to this change.
```
SUMMARY: MemorySanitizer: use-of-uninitialized-value /src/libheif/libheif/plugins/encoder_jpeg.cc:408:15 in jpeg_encode_image(void*, heif_image const*, heif_image_input_class)
```

libheif/box.cc
```diff
@@ -3366,14 +3366,23 @@ void Box_ipma::sort_properties(const std::shared_ptr<Box_ipco>& ipco)
...
 
     // simple Bubble sort as a stable sorting algorithm
 
     for (size_t n = 0; n < nAssoc - 1; n++)
       for (size_t i = 0; i < nAssoc - 1 - n; i++) {
+        uint16_t idx_a = item.associations[i].property_index;
+        uint16_t idx_b = item.associations[i + 1].property_index;
+        if (idx_a == 0 || idx_a > properties.size() ||
+            idx_b == 0 || idx_b > properties.size()) {
+          continue;
```
bradh, line 3381: Does it really make sense to continue here? This looks like something is badly wrong and the resulting file would be invalid.

farindk: First, we should never generate any `property_index==0`. A zero value is allowed according to the HEIF standard, but it means "no property associated", which is pretty useless.

If we would decide that there might be any zero property_index in the list, the implementation above does not work because the zero-index will act like a barrier in the bubble sort that no other entry can cross.

I would propose that we either add an `assert` that all indices are >0 or we sort the 0-indices properly like a non-transformative property.

farindk:
I realized that the whole bubble sort is overkill. We only need to separate descriptive from transformative properties while preserving their relative order. This can be done faster and simpler with `std::stable_partition`. I have replaced the whole sort with that. Zero and invalid indices will be treated like a descriptive property.

Thanks @lovell for spotting the `nAssoc == 0` integer overflow edge case.

lovell:
Thanks both, :+1: for the use of `stable_partition`.

PR #1766, opened 2026-04-14
Optimize buffer allocation in PNG decoder

torusrxxx:
It’s strange that each row needs a separate malloc. Replaced it with just one malloc for better cache locality and performance.

I turned on auto-vectorization reports, nothing was auto-vectorized. I may spend some time doing performance optimizations by enabling auto-vectorization in future.


bradh:
Interesting. Can you say how much performance difference this made?

farindk:
Thanks, I have added some checks for integer overflows.

PR #1771, opened 2026-04-24
Fix SDL YV12 buffer overflow when converting 444 into 420

torusrxxx:
This function is pointing base of U plane to the end of U plane, causing a buffer overflow.

### DIFF
--- a/examples/sdl.cc
+++ b/examples/sdl.cc
@@ -251,7 +251,7 @@ void SDL_YUV_Display::display444as420(const unsigned char *Y,
     }
 
   uint8_t *startV = mPixels + (rect.h*mStride);
-  uint8_t *startU = startV + (rect.h*mStride/2);
+  uint8_t *startU = startV + (rect.h*mStride/4);
   for (int y=0;y<rect.h;y+=2)
     {
       uint8_t* u = startU + y/2*mStride/2;


torusrxxx:
I have a few questions here. How far do you want to push libheif examples forward? Currently, `heif-view` can't even view photos of my smartphone, and I want `heif-view` to support all HEIF features including HDR (bumping SDL 2 to SDL 3). Also, `libheif` is already compiling `libwebp` dependency, why not add WebP support to the examples? I already have added that in my builds, I miss that WebP lossless mode. Just to make sure `libheif` wants it as well. And what about SIMD? Currently, there is no SIMD. Is it acceptable to import some generic portable SIMD wrappers here? Is it acceptable to add AVX code here? Does `libheif` want color space conversions between all different HDR color spaces in `libheif` code with possible SIMD accelerations?

farindk:
Thanks.

The main reason for me to add `heif-view` was that I needed this for testing during the image sequences implementation. I'm open to having also an image viewer application. Note that there is a geo-spatial imaging community that is in particular interested in large images, "strange" content like multi-spectral images. For this reason, I started (also mainly for my own testing) this viewer: https://github.com/farindk/tiled-image-viewer
Maybe that would be a better starting point than `heif-view`.

Concerning SIMD: sure, that would be great. Maybe integrating libyuv would safe us a lot of implementation time for the color conversions.

Issue #215, opened 2020-04-19
Converting to Go image.Image produces flat colors


kyroy:
I would like to use this library to convert my heic images from my iPhone to jpeg. Unfortunately, the colors of the resulting images differ from the originals.

Example code:
```go
// error handling omitted
file := "example.heic"
f, _ := os.Open(file)
defer f.Close()

goImg, _, _ := image.Decode(f)
var writer bytes.Buffer
jpeg.Encode(&writer, goImg, &jpeg.Options{
	Quality: 100,
})
ioutil.WriteFile("example.jpeg", writer.Bytes(), 0644)
```

This is also true when using the [example.heic](https://github.com/strukturag/libheif/blob/master/examples/example.heic) from this repo.

Result:
![res](https://user-images.githubusercontent.com/4368180/79685020-e714d080-8235-11ea-92d4-2e09f07cd762.jpeg)

countingpine:
Hi kyroy, I hope you don't mind my input on this issue.
If I interpret it correctly: you're experiencing some loss of quality when converting the image to JPEG?
If that's right, then I would say the issue is that JPEG is a "lossy" format, even at quality 100, and that if you try outputting to PNG, which is a lossless format, it should perfectly preserve the colour information.

silverbacknet:
@countingpine It seems more like a ICC or colorspace that isn't being carried over, though I didn't see any on a cursory glance, since everything else lines up. It could be just a special bit Apple throws in because Apple. It's definitely not JPEG; if you compare the two you'll see that there's no perceptual difference at all unless you're on iPhone.

Issue #1728, opened 2026-03-15
gdk-pixbuf: Ignores HLG and gives incorrect, dull output


kennylevinsen:
libheif's GdkPixbuf loader currently ignores `transfer_characteristics` entirely, yielding images that have extremely dull colors when the image is HLG encoded. This ultimately affect things like GNOME's image previewer and thumbnail generator.

[Thes test files](https://github.com/kennylevinsen/hlg-test-images) can be used to validate the presentation, and [tev](https://github.com/Tom94/tev) renders these images as expected as a reference.

For a fix, one could either extend libheif color conversion to take care of transfer_characteristics when requested, or maintain that it is the callers responsibility and add local conversion to the GdkPixbuf loader.

kennylevinsen:
#1729 takes the approach of converting inside the pixbuf loader. It probably isn't many formats that we would ever have to deal with - other popular HDR image formats use gain maps and can just be treated as boring standard srgb content for example. I doubt anything emits PQ-encoded images right now (famous last words).

kmilos:
> I doubt anything emits PQ-encoded images right now (famous last words).

I believe a few SLR cameras started offering this recently instead of JPEGs, for example Canons?

> For a fix, one could either extend libheif color conversion to take care of transfer_characteristics

As mentioned in the PR comment, one can't just neglect color_primaries.

kmilos:
FWIW, a lot of the distros (e.g. Fedora and Ubuntu) are not shipping the gdk-pixbuf plugin any longer as glycin seems to be taking over as image loader, so not sure a huge investment here is worth it in the long run?

kennylevinsen:
> As mentioned in the PR comment, one can't just neglect color_primaries.

Of course, but ITU-R BT.2100 specifies BT.2020 primaries specifically, hence the hardcoded assumption. If there are systems outside BT.2100 that combine HLG with other primaries, then we'd have to handle that separately - I suspect the constants from the BT.2100 OOTF would require adjustment to fit other primaries?

EDIT: Yes, the the formula embeds the BT.2020 luminance matrix - those coefficients would need to be replaced for other color spaces.

kmilos:
> Of course, but ITU-R BT.2100 specifies BT.2020 primaries specifically, hence the hardcoded assumption.

[...]

Issue #1572, opened 2025-08-17
Implement RGB to YCgCo-R


farindk:
We miss the RGB to YCbCr conversion for YCgCo(-R). This is `matrix_coefficients=8`.

While we are at it, we could also add support for `matrix_coefficients=16` and `17` (YCgCo-Re / YCgCo-Ro) which were recently added to H.273 (https://www.itu.int/ITU-T/workprog/wp_item.aspx?isn=18689).

novomesk:
I have AVIF with matrix_coefficients=16

http://188.121.162.14/zoo.avif

heif-dec gives following output:

![Image](https://github.com/user-attachments/assets/349a2037-5719-422a-9075-8a756cfeed48)

output of avifdec:

![Image](https://github.com/user-attachments/assets/a1e1b07a-24ee-4b8b-813a-d08bfd87c2f5)

farindk:
The above is a first, quick implementation of YCgCo-Re (`matrix_coefficients=16`). It is probably still incomplete, but works for the example image. A better implementation is still future work.

farindk:
Note: the example image's `pixi` box says that the image has 10 bits per pixel:
````
| | | Box: pixi ----- (Pixel Information)
| | | size: 16   (header size: 12)
| | | bits_per_channel: 10,10,10
````
This is the internal YCgCo-Re space. The desired output RGB depth is 8 bits per pixel. How can we communicate this cleanly through the API?

novomesk:
I understand it this way:

The -RE indicate that "even" (2) bits were added to intended bit depth,
so 10-2=8 is desired RGB depth for the output.

farindk:
> I understand it this way:
> 
> The -RE indicate that "even" (2) bits were added to intended bit depth, so 10-2=8 is desired RGB depth for the output.

Yes, correct. My problem is that the color-conversion function where I added it, must have the same output bpp as the input. And the whole conversion pipeline does not expect that it is desired to have less bits in the output than in the input. This needs more work. But at least, we get a usable picture now.

Issue #1183, opened 2024-06-09
Canon HIF (10 bit) Conversion to JPG : Tone Mapping & EXIF


paswai:
Hello all, I tried the [demo ](https://strukturag.github.io/libheif/) (version 1.16.2) to convert a [HIF file (10 bit) ](https://we.tl/t-sgYLO9lAY7) to JPG. The conversion went smoothly, however, I noticed two things : 1) No tone mapping has been applied (JPG image looks flat) and 2) All EXIF have been lost.
Is this an expected behaviour and is it still present in the newest release?

bradh:
Can you attach the file instead? Zip it if needed.

farindk:
I have downloaded the image: [issue-1183-data.zip](https://github.com/user-attachments/files/15759152/issue-1183-data.zip)

The image contains no auxiliary tone-mapping image, just standard EXIF and XMP metadata.

The NCLX color profile says that it uses a PQ transfer characteristic, which is probably the reason that the image looks dull when it is not taken care of. During the conversion to JPEG, libheif does not modify the image data itself. The EXIF is preserved, but the NCLX data is lost.

The question is: should libheif handle the color profile conversion? This may still be doable for simple things like here, but when we want to also convert between attached ICC profiles, we would need to use `lcms2` as an additional dependency.

paswai:
I was curiously following the discussion - Many thanks for the insights! @farindk If the tone mapping is out-of-scope for this project, please close this issue.

farindk:
For the library itself, this should not include any color management other than basic nclx handling. However, I would consider doing some (optional) color profile conversion in `heif-convert` if we would otherwise lose that information.

I'll leave this issue open.

PR #1778, opened 2026-05-01
HDR tone mapping

torusrxxx:
This PR enables color space conversions between different NCLX profiles and tone mapping from HDR to SDR.
The color conversion pipeline no longer ignores NCLX differences. Different transfer functions are now converted properly. And in the case of converting from HDR to SDR, either the Reinhard tone mapping (HDR to sRGB) or clamping (HDR to linear SDR) is performed. Conversion from Rec. 2020 and P3 color spaces to sRGB is also supported.
HDR to HDR color conversions are also supported, including conversion between PQ, HLG and sRGB, between Rec. 2020, P3 and BT. 709.
Conversion from and to the constant luminance ICtCp matrix coefficient defined in Rec. 2100 is also supported.
This is going to close many existing issues. Existing software that wants sRGB will now receive tone-mapped sRGB version of HDR image. It is expected that some images with invalid NCLX might fail to load because the NCLX-aware color conversion pipeline. When a conversion from HDR to SDR is performed, the ICC profile and HDR metadata are also removed.
There is no new dependency. ICC profile is untouched, other than its removal when HDR is tone-mapped to SDR.

----

Edit: removed Reinhard tone mapping (HDR to sRGB) or clamping (HDR to linear SDR) for now. The sole color conversion operator added in this PR is based on zimg. zimg has a poor HDR to SDR quality, the removed Reinhard tone mapping might be added back in a future PR if necessary.


torusrxxx:
I'll evaluate https://github.com/sekrit-twc/zimg/ for a complete color conversion pipeline.

torusrxxx:
Added zimg color conversion library.

torusrxxx:
Probably fix #215, #580, #1163, #1183, #1572, #1728 and make #1729 obsolete as zimg has comprehensive support for all kinds of CICP conersion. My previous code for HDR and SDR conversion might also be removed in favor of zimg.

torusrxxx:
Because most downstream applications still don't have support for all kinds of CICP combinations even when HDR is supported, I'll restrict libheif output color space to BT.2020 nonconstant luminance and PQ/HLG transfer characteristics for HDR. The application should provide a callback to let libheif query whether a CICP combination (like YCgCo) is supported or not before passing through YCgCo.

farindk:
Thanks, this is great. I just need time to review this. It is a large PR. And I also think I need to learn more about HDR first, since my knowledge about it is still a bit limited.

CMakeLists.txt
```diff
@@ -642,6 +642,8 @@ if (CMAKE_CXX_COMPILER MATCHES "clang\\+\\+$")
     add_compile_options(-Wno-tautological-constant-out-of-range-compare)
 endif()
 
+add_subdirectory (third-party/zimg)
```
kmilos: Once the feature stabilizes, please also provide a way to unbundle and use the system zimg.

P.S. Also can't say I'm a great fan of having this as an ever present submodule rather than being fetched on demand if actually doing the bundling (like present deps)...

farindk: Yes, it should be optional and (if not using the system zimg) use the same idiom as the other locally compiled dependencies in the `third-party` folder.

torusrxxx: Hey, I’m using Microsoft Windows, there is no system zimg. This feature is currently mostly stable. It will need your help to convert this into system zimg.

kmilos: > I’m using Microsoft Windows

platform != toolchain

MSVC? MinGW?

> there is no system zimg

At least for MinGW (MSYS2) there is: https://packages.msys2.org/base/mingw-w64-zimg

Unfortunately not available in vcpkg for MSVC builds indeed...

However, it seems to be widely available for all others: https://repology.org/project/zimg-image-processing/versions

> It will need your help to convert this into system zimg

As hinted, just follow the idiom for all the other deps under `third-party`, and at least populate and make sure the "local" path works for you.

P.S. Any particular reason for your fork of zimg rather than upstream?

torusrxxx: I'm using Microsoft Visual Studio 2022 and vcpkg. Since there is no zimg in vcpkg, I created a fork of zimg merely to add cmake build files to build it. There is no other changes to zimg. I am not able to test how a system zimg works, nor can I simply build official zimg because MSVC doesn't have GNU autotools.

kmilos: Hm, it seems there is a MSVC sln ready in the upstream `_msvc` folder? This is [how e.g. conda-forge builds it OOTB](https://github.com/conda-forge/zimg-feedstock/blob/0d198bce9f1442920f659f3bf30bda199792a95e/recipe/bld.bat#L1). (Note also that they [patch the .def file](https://github.com/conda-forge/zimg-feedstock/blob/main/recipe/libfile-link-to-zimg-dll.patch) to match the renamed artifacts.)

So you'd have to also recreate these steps in the local build command (and also add the automake build ones for all others.)

Although it would be of course nicer if there was a cross-platform build system for zimg, but I see the maintainer didn't respond to an upstream feature request for e.g. meson...

P.S. Even checking in just your CmakeLists.txt and clone+build script in `third-party` is a "lesser evil" than having an actual submodule pointing to a fork IMHO...

torusrxxx: Building a MSVC project dependency with cmake is horrible. Your suggestion to copy CMakeLists.txt into `third-party/zimg` by a set up script seems more reasonable.

torusrxxx: Removed zimg submodule. Now, it needs to clone https://github.com/sekrit-twc/zimg/ into `third-party/zimg`, then compile the MSVC project in `_msvc` folder with Visual Studio. The MSVC project needs to be modified if the MSVC version is different.

torusrxxx: Now added zimg setup script for MSVC 2022. zimg is kind of mandatory, because libheif color conversion pipeline is (properly) tightened up. If zimg is unavailable, libheif will refuse to output a washed out image. It breaks applications that are currently showing a washed out image, unless my small color conversion routines are added back directly in libheif to remove zimg dependence and instead convert directly in libheif.

PR #1779, opened 2026-05-01
decoders: guard against possible integer overflow

lovell:
This keeps things consistent with other parts of the codebase where stride is typically represented as `size_t`.

I don't think this is exploitable, plus existing security limits guard should protect against it being reached, but this change at least silences static analysis tooling.


farindk:
Thanks. I've also added a similar integer promotion to the memcpy source argument.

PR #1780, opened 2026-05-01
decoder_libde265: guard against version length in plugin name

lovell:
Another small fix to keep static analysis happy and to protect against the unlikely event that someone builds a version of libde265 with a pathologically long version string.


farindk:
Thanks. I slightly improved it. Bounds check was off by one.

PR #1782, opened 2026-05-01
decoder_aom: ensure security limits apply to all frames

lovell:
It looks like these were being ignored for animated AVIF sequences as `nullptr` skips the check here:

https://github.com/strukturag/libheif/blob/7283380ec21003a6812d8a1e7e4328bdda8b2026/libheif/pixelimage.cc#L654

I'm unsure if applying the "global" limits is the correct way to implement this but that's probably better than not having any limits. An alternative approach might be to modify the `heif_decoder_plugin_options` struct to add custom `limits` and pass it along that way.


farindk:
If we use the global security limits, it may reject large images and users have no way to override this. I've implemented it instead by passing the limits through the decoder options as you proposed.

lovell:
Danke Dirk!

PR #1783, opened 2026-05-01
region: ensure offset is incremented after parsing inline mask

lovell:
Most region parsers rely on `parse_unsigned` to increment `dataOffset` but inline masks are slightly different in that they involve copying data into a buffer. I think this means we need to manually increment the offset before further parsing.

Disclosure: I used LLM-based tooling to help find this bug but the (attempt at a) fix is mine.


farindk:
Thank you. I've added a regression test for this.

PR #1789, opened 2026-05-07
Print NCLX color profile and CLL

torusrxxx:
Added output of NCLX color profile and content light level metadata like ffprobe


farindk:
Thank you.

farindk:
I've extended it to human-readable output.

PR #1790, opened 2026-05-07
Remove Rec 2020 CL matrix coefficients from preset

torusrxxx:
Rec 2020 CL is not properly supported by `libheif`. The image is actually Rec 2020 NCL.


farindk:
Thank you

PR #1805, opened 2026-05-18
Fix webp not announced in usage

torusrxxx:


farindk:
I changed it to build the list dynamically, listing only the compiled codecs.

torusrxxx:
Yes, thank you.
https://github.com/strukturag/libheif/blob/86beddfcd1db072b8c60d49f672287abda1dee50/examples/heif_dec.cc#L104
Double jpeg should be jpg, jpeg.

farindk:
> Double jpeg should be jpg, jpeg.

Right, thank you. Fixed.

PR #1812, opened 2026-05-19
heif_properties.h: fix "bad_pixels" type

stefantalpalaru:
Problem found when trying to build the latest ImageMagick with the latest libheif.


farindk:
Thank you.

PR #1815, opened 2026-05-21
svt: don't allow getting qp unless it's set

flaviozavan:
Prevents getting the qp parameter unless `qp_set` is `true`. Without this check, no distinction is made between real values and uninitialized ones, which can lead to false assumptions.

----

After upgrading to 1.22.0, I noticed some of my svt encoding calls started failing with `Svt[error]: CRF must be [0 - 63]`. Upon further investigation, it seems this started happening after c992822.

Looking at the changes introduced in that commit, when copying parameters for the alpha encode, the uninitialized `qp` is happily returned, and subsequently set, flipping `qp_set` to true in the copy. A TODO annotation even points out that this could be problematic.

To aid in understanding this issue, I had AI generate sample code for reproducing it, which turned out a bit verbose, but useful nonetheless. Without the proposed changes, we get the failure. With the changes applied, there are no failures.
[reproduce.c](https://github.com/user-attachments/files/28111505/a.c)

I can also confirm this fixes the failures I saw in my real use-case of libheif.

### DIFF
--- a/libheif/plugins/encoder_svt.cc
+++ b/libheif/plugins/encoder_svt.cc
@@ -475,10 +475,16 @@ heif_error svt_get_parameter_integer(void* encoder_raw, const char* name, int* v
   else if (strcmp(name, heif_encoder_parameter_name_lossless) == 0) {
     return svt_get_parameter_lossless(encoder, value);
   }
+  else if (strcmp(name, kParam_qp) == 0) {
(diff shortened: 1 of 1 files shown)


farindk:
Thank you. This makes sense. In the future, we should switch to a dedicated `heif_error_parameter_not_set`, but I cannot add that in a patch release.

PR #1828, opened 2026-05-28
Pass through original HEIF NCLX in HEIF transcoding

torusrxxx:
It turned out the heif transcoding feature didn't support HDR. Pass through original HEIF NCLX in HEIF transcoding.

If this were not for heif transcoding, but for displaying, the client application also needs to query colr box to determine whether that is supported, then set output color space to one compatible with the client application. Currently most applications don't do that, and either receive sRGB only, or anything including YCgCo. I'm not sure whether to force it via breaking changes (https://github.com/strukturag/libheif/issues/1778#issuecomment-4527524999).

### DIFF
--- a/heifio/decoder_heif.cc
+++ b/heifio/decoder_heif.cc
@@ -81,7 +81,10 @@ heif_error loadHEIF(const char* filename, InputImage* input_image)
   // during decode (default ignore_transformations=0), so the resulting pixels are
   // already in display orientation and input_image->orientation stays 'normal'.
 
+  heif_color_profile_nclx *nclx = nullptr;
+  heif_image_handle_get_nclx_color_profile(handle, &nclx);
   heif_decoding_options* opts = heif_decoding_options_alloc();
+  opts->output_image_nclx_profile = nclx;
 
   heif_image* image = nullptr;
   err = heif_decode_image(handle, &image,
@@ -89,6 +92,8 @@ heif_error loadHEIF(const char* filename, InputImage* input_image)
                           opts);
 
   heif_decoding_options_free(opts);
+  if(nclx)
+    heif_nclx_color_profile_free(nclx);
 
   if (err.code != heif_error_Ok) {
     heif_error stable = stable_error(err);


farindk:
Thanks. I'm thinking whether it would be easier to have an option "enable NCLX passthrough" in the `heif_decoding_options` so that simply passing `nullptr` for the output nclx means passthrough instead of sRGB.

farindk:
The patch above adds the passthrough option and might be the easier solution.

torusrxxx:
The problem is that the application will then also receive ICtCp/YCgCo with this option, which the application doesn't support. The application really needs to check whether the image is using an unsupported color space, and only pass in a NCLX compatible with the application (it is almost always Rec.2020 non-constant luminance PQ/HLG). Otherwise, the application will not support HDR correctly.

Also, currently libheif still passes through CLL even when the output is converted to SDR, which is incorrect. CLL has to be dropped for any output transfer characteristics except PQ.

farindk:
> The application will then also receive ICtCp/YCgCo with this option, which the application doesn't support.

I think this is exactly why making it an opt-in flag is the safe design: the default behavior (nullptr = sRGB) is unchanged, so an application only ever receives the original color space if it explicitly asks for passthrough. An app that can't handle arbitrary NCLX simply never enables the flag and is unaffected. It can still request a NCLX that it supports.

The flag is really meant for two cases:
1. Transcoding. Here you want the original NCLX passed through verbatim (ICtCp, YCgCo, PQ, …) because the data is going straight back into a HEIF file and must be preserved losslessly. There is no "supported color space" question in this path.
2. Display apps that genuinely handle arbitrary NCLX, or that inspect the profile first.

For the display case, the application has to query the profile (heif_image_handle_get_nclx_color_profile) and request a specific output NCLX it can render. The flag doesn't prevent that. libheif fundamentally can't know which color spaces a given application supports, so that decision has to stay with the caller.

> currently libheif still passes through CLL even when the output is converted to SDR, which is incorrect. CLL has to be dropped for any output transfer characteristics except PQ.

Right, that's a real bug, but it's independent of the passthrough flag. Let's track it separately. There may be further HDR processing issues in the conversion pipeline. This is probably a bigger task.

torusrxxx:
In #1778, which I request a review, I made a change that it will only set sRGB defaults if the application sets `convert_hdr_to_8bit` to `true`. If `convert_hdr_to_8bit` is true, then the output is converted into sRGB and ICC, CLL, MDCV are stripped. If `convert_hdr_to_8bit` is false, then the application requests HDR, original transfer characteristics is preserved. There's no need to add a separate setting just to handle the rare case in which the application does support 10-bit color without supporting anything but sRGB. On the other hand, it is likely that the application which supports 10-bit color still don't support YCgCo, and it is still desirable to convert YCgCo into RGB even when the application sets `convert_hdr_to_8bit` to false.

Regarding transcoding/display use cases, real world applications overwhelmingly need display capability. Transcoding applications are highly centralized and supported (ffmpeg, heif-enc itself). The correctness requirement in transcoding applications usually needs somewhat verbose code anyway. We just need display applications to quickly load image in sRGB and toggling `convert_hdr_to_8bit` to false gives HDR, that's it, without even learning about what is YCgCo.

The changes made here is independent to #1778. With #1778 in place, the changes made here is no longer HDR/SDR difference (`convert_hdr_to_8bit` is `false` so it will support HDR), but to ensure even YCgCo gets passed through, if we decide to not pass through YCgCo by default.

farindk:
We have to be careful not to change the meaning of `convert_hdr_to_8bit`. You never know who relies on the current behavior. These are all old parameters preceding the time anybody worried about HDR. Now, I would do that differently.
Maybe we can clean that up in a future libheif v2. For the moment, we have to live with them and stay compatible. I also do not want to make the logic too complicated (if `convert_hdr_to_8bit` is off, default is sRGB, otherwise it is passthrough, except for YCgCo, because that is usually not supported, but maybe it will change in the future when it becomes common enough, ...).
Simple, predictable, and the rest is left to the application.

torusrxxx:
So `convert_hdr_to_8bit` is more about making sure the output is 8-bit than about color space, right? If an application sets `convert_hdr_to_8bit` to true, then the application only supports 8-bit color. If the `output_image_nclx_profile_passthrough` is set to false, then sRGB. If `output_image_nclx_profile_passthrough` is set to true, then the application is receiving 8-bit PQ, which is probably wrong. If the application sets `convert_hdr_to_8bit` to false, then the application supports 8-bit color and 10-bit to 16-bit color. If the `output_image_nclx_profile_passthrough` is set to false, then 10-bit sRGB, kind of weird. If the `output_image_nclx_profile_passthrough` is set to true, then any HDR format including YCgCo is allowed.

In the end, 2 out of 4 combinations are kind of weird. The only way to use them correctly would be overriding the NCLX by the application manually to restore expected color space. If `libheif` starts to correctly transform HDR to SDR per #1778, these applications would produce explosive images by trying to render images in correct sRGB as HDR. These broken use cases are arguably not worth preserving. That's before talking about YCgCo. If we restrict HDR to mean only Rec. 2020 nonconstant luminance with PQ/HLG full range, we have generally good compatibility for all practical HDR use cases.

farindk:
Yes, `convert_hdr_to_8bit` was historically plain bit-depth without any color-space considerations.

I understand your view and I agree with its principles. That's how it should be after a cleanup.
At some point, we should deprecate these old options, or at least document a preferred way how to set the parameters to work correctly. We could also have a new `heif_decoding_options_alloc2()` that uses new defaults. Or we do a new `heif_decoding_options2` and keep the old one as deprecated API. We just have to make this carefully step by step.

torusrxxx:
Yeah, keep `convert_hdr_to_8bit` unchanged (or correctly implement HDR to SDR for it, instead of reinterpreting HDR as dull SDR). Recommending `output_image_nclx_profile_passthrough` for all HDR applications will however create new broken applications again, just like how YCgCo images are displayed by unsupported applications currently. Let's not do it and make applications opt into transcoding use case like this (#1828) explicitly. The new option for HDR applications should not pass through YCgCo without application opt-in.

torusrxxx:
Please close this and change this into an issue to continue discussion.

PR #1832, opened 2026-05-31
fix broken PNG HDR support

torusrxxx:
* If PNG_cICP_SUPPORTED, then let HEIF decoder pass through NCLX
* If color conversion is performed, a discrepancy exists between colr box NCLX and actual image NCLX. The actual image NCLX should be encoded.


farindk:
Right. Looks good to me. Thank you.

PR #1833, opened 2026-05-31
Move the TAI clock_type into the 2 highests bits

ocrete:
Reading the MP4 spec, I think it should be in the 2 highest bits, not the whole 8-bit


farindk:
Yes, you are correct. Thank you.

PR #1836, opened 2026-06-05
SVT-AV1: add tune=iq encoder parameter

### LINKED ISSUE #1835: Missing tune=iq SVT-AV1 encoder parameter
There's a missing tune=iq SVT-AV1 encoder parameter: https://github.com/strukturag/libheif/blob/master/libheif/plugins/encoder_svt.cc#L67
introduced by https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/2489 in https://gitlab.com/AOMediaCodec/SVT-AV1/-/releases/v4.0.0

pszemus:
Fixes #1835

### DIFF
--- a/libheif/plugins/encoder_svt.cc
+++ b/libheif/plugins/encoder_svt.cc
@@ -68,7 +68,10 @@ struct encoder_struct_svt
   {
     Tune_VQ = 0,
     Tune_PSNR = 1,
-    Tune_SSIM = 2
+    Tune_SSIM = 2,
+#if SVT_AV1_CHECK_VERSION(4, 0, 0)
+    Tune_IQ = 3
+#endif
   };
 
   uint8_t tune = Tune_PSNR;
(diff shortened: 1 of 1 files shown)


pszemus:
@farindk ping

farindk:
Thank you. I've also added "ms-ssim", which is another tune parameter introduced in SVT 4.0.0.

pszemus:
Thanks @farindk ! What do you think about setting "iq" as the default value for the "tune" parameter?
Apparently, this provides the best quality-to-size ratio for AVIF images.

farindk:
> What do you think about setting "iq" as the default value for the "tune" parameter?

Yes, but it is a bit more complicated (see `encoder_aom.cc`). There, "iq" is not used for sequences (uses SSIM) or alpha channels (uses PSNR). I would expect that SVT behaves similarly, but this should be tested first.

I'll reopen this issue (#1835) to indicate that there is still work needed here.

PR #1837, opened 2026-06-05
file_layout: add box size integer overflow checks

lovell:
Implements a couple of "TODO" comments, plus similar missing checks in neighbouring code. I suspect the existing security limits would prevent this also, so unlikely to be exploitable when using defaults.


farindk:
Thank you. I have additionally changed `BitReader` to use `size_t` instead of `int`, which makes one of the checks (and many similar casts at other places) unnecessary.

PR #1839, opened 2026-06-11
ci: add permissions and pin action SHAs (13 workflows)

Alb3e3:
## Summary

The CI workflows are missing top-level \`permissions\` declarations and reference actions by mutable version tags, which are supply-chain risk vectors.

**Changes:**
- Add \`permissions: read-all\` at the workflow level for all workflows (least-privilege default)
- Pin all third-party action references to full commit SHAs

Mutable action references (\`@v4\`, \`@main\`, etc.) allow a compromised upstream action to run arbitrary code in your CI. Pinning to a commit SHA ensures only the audited version runs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     strategy:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -8,6 +8,9 @@ on:
   schedule:
     - cron: "0 2 * * SUN"
 
+permissions:
+  contents: read
+
 jobs:
   Fuzzing:
     runs-on: ubuntu-22.04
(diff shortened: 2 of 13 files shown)


farindk:
Thank you.

permission blocks: yes, happy to take these. Maybe you could even tighten permissions: `read-all` to `permissions: { contents: read }`? That's the actual least-privilege scope these build/test/lint jobs need.

SHA pinning: I'd prefer to skip that since my main concern is maintenance. We currently rely on Dependabot for action updates. The SHA pins will rot and every bump becomes an unreviewable 40-character hash diff.

Alb3e3:
Thanks for the review! I've updated the branch (`415abcc`) to match both points:

- **Permissions:** switched every workflow from `permissions: read-all` to `permissions:\n    contents: read` — the actual least-privilege scope these build/test/lint jobs need.
- **SHA pinning:** dropped entirely. The action refs are back to their version tags (`@v6`, `@v5`, etc.) so Dependabot can keep updating them with reviewable diffs.

The PR is now permissions-only: 13 workflow files, each gaining a single `contents: read` block and nothing else.

farindk:
Thank you for the changes.

PR #1840, opened 2026-06-18
uncompressed: enforce security limits on icef unit allocation

SaudSatopay:
Box_icef::parse() resized m_unit_infos to the file-controlled num_compressed_units without routing the allocation through the security-limits memory accounting (the existing "// TODO: should we impose some security limit?"). A crafted icef box could therefore drive a very large allocation before the per-unit reads run.

Allocate via m_memory_handle.alloc(num_compressed_units, sizeof(CompressedUnitInfo), limits, ...) before the resize, mirroring Box_snuc and the sequence boxes, so the configured heif_security_limits apply.

### DIFF
--- a/libheif/codecs/uncompressed/unc_boxes.cc
+++ b/libheif/codecs/uncompressed/unc_boxes.cc
@@ -833,9 +833,12 @@ Error Box_icef::parse(BitstreamRange& range, const heif_security_limits* limits)
             sstr.str()};
   }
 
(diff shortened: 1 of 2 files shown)


farindk:
Thank you. The amount of memory allocated is actually already limited by the box size and hence the file size. But since there can be an amplification of factor 16, I think the additional check is still warranted.

PR #1846, opened 2026-06-22
Enable FFmpeg AV1/VVC/JPEG/JPEG2000 decode

torusrxxx:
Fixed FFmpeg decoder AV1/VVC/JPEG/JPEG 2000 decode support. This is because only ITU codecs use ITU style NALs.

### DIFF
--- a/libheif/plugins/decoder_ffmpeg.cc
+++ b/libheif/plugins/decoder_ffmpeg.cc
@@ -69,6 +69,11 @@ struct ffmpeg_decoder
   }
 };
 
+// H. 264, H. 265, H. 266 require NAL. Others have no NAL.
+static bool supportsNal(AVCodecID id) {
+  return id == AV_CODEC_ID_H264 || id == AV_CODEC_ID_H265 || id == AV_CODEC_ID_H266;
+}
+
 static const int FFMPEG_DECODER_PLUGIN_PRIORITY = 90;
 
 #define MAX_PLUGIN_NAME_LENGTH 80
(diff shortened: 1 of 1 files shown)


farindk:
Thank you.

Could you test VVC? My ffmpeg 6.1 has no VVC support.
For JPEG2000, my ffmpeg outputs rgb24, which the plugin cannot process.
HTJ2K is not supported by my ffmpeg.
AV1 and JPEG works for me.

farindk:
Merged. Note:
- VVC requires FFMPEG 7.0/7.1 (but is cleanly disabled by 6.1)
- HTJ2K decodes incorrectly with FFMPEG 6.1. I've added detection of ffmpeg 7.x when it queries for decoder support. It is also required to check whether the bitstream it actually HTJ2K because it is not indicated in the file whether it is plain JPEG2000 or HTJ2K.
- For JPEG2000 and HTJ2K, ffmpeg output rgb24 even though the image data is YCbCr 4:4:4 since there is no way it can differentiate the two. The decoder plugin still treats this as YCbCr. This may need further work.

torusrxxx:
Yes, I confirm VVC works. FFmpeg version 8.1.1 MSVC (https://www.gyan.dev/ffmpeg/builds/).
I encoded a VVC HEIF file with vvenc/heif-enc, in case other people want to test VVC.
[1.zip](https://github.com/user-attachments/files/29341450/1.zip)
decodes in to

torusrxxx:
Can you update CMakeLists.txt to show additional decoding support if FFmpeg is found?

farindk:
Added in above commit.

torusrxxx:
Please add these FFmpeg decode capabilities also in README.md, thanks.

PR #1848, opened 2026-06-22
Fix pixi writing for multi-channel cases

### LINKED ISSUE #1847: pixi channel count regression: 3-channel AVIF written with one pixi channel
heif-enc from libheif 1.23.0 can write a 3-channel AVIF whose pixi item property only lists one channel. This is not correct, but is tolerated by some decoders (Chrome), but will fail in others (including Adobe software).

The encoded AV1 payload is 10-bit YUV444 and decoders report three channels, but the container pixi box says only:

    bits_per_channel: 10

Expected:

    bits_per_channel: 10,10,10

This appears to be caused by short-circuiting in ImageItem::encode_to_bitstream_and_boxes():

https://github.com/strukturag/libheif/blob/v1.23.0/libheif/image-items/image_item.cc#L335-L355

For example:

    valid_pixi = (pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Y)) ||
                  pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Cb)) ||
                  pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Cr)));

Box_pixi::add_channel_bits() returns true when it successfully appends a channel depth. Because the calls are joined with ||, the expression short-circuits after the first successful channel, so channels 2 and 3 are never appended.

Box_pixi::write() then serializes m_bits_per_channel.size() as the channel count:

https://github.com/strukturag/libheif/blob/v1.23.0/libheif/box.cc#L2692-L2705

So the resulting file gets num_channels = 1 even though the encoded image is 3-channel.

Reproduction

Input: a valid 10-bit YUV444 AVIF whose source pixi is correct:

    heif-info -d goodSource.AVIF | sed -n '/Box: pixi/,/Box: ipma/p'

Source shows:

    Box: pixi ----- (Pixel Information)
    bits_per_channel: 10,10,10

Transcode with libheif 1.23.0:

    heif-enc -A -q 90 --bit-depth 10 -p chroma=444 \
      --colour_primaries 12 \
      --transfer_characteristic 16 \
      --matrix_coefficients 6 \
      --full_range_flag 1 \
      -o out.avif \
      goodSource.AVIF

Example source and transcoded result are attached for convenience.

Inspect the output:

    heif-info -d pixi-repro.avif | sed -n '/Box: pixi/,/Box: ipma/p'

Actual output:

    Box: pixi ----- (Pixel Information)
    bits_per_channel: 10

But the AV1 payload is still 3-channel YUV444:

    avifdec --info pixi-repro.avif

Reports:

    Bit Depth      : 10
    Format         : YUV444
    Alpha          : Absent
    Range          : Full

ExifTool also reports the mismatch as:

    ChromaFormat    : YUV 4:4:4
    ImagePixelDepth : 10

Regression range

This appears to have regressed in:

https://github.com/strukturag/libheif/commit/d0f4669c287b87b0d2352db30f6f58b7e77995b0

Commit message:

    fix type mismatches because of new uint16_t bpp

That commit changed the previous separate add_channel_bits() calls into short-circuiting || expressions.

Observed locally:

- libheif 1.21.2: writes pixi as 10,10,10
- libheif 1.23.0: writes pixi as 10

The commit is contained in tags v1.22.0, v1.22.1, v1.22.2, and v1.23.0, but not v1.21.2.

[pixi source and out.zip](https://github.com/user-attachments/files/29215287/pixi.source.and.out.zip)

bradh:
Resolves #1847

### DIFF
--- a/libheif/image-items/image_item.cc
+++ b/libheif/image-items/image_item.cc
@@ -333,14 +333,14 @@ Result<Encoder::CodedImageData> ImageItem::encode_to_bitstream_and_boxes(const s
     valid_pixi = pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Y));
   }
   else if (colorspace == heif_colorspace_YCbCr) {
-    valid_pixi = (pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Y)) ||
-                  pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Cb)) ||
+    valid_pixi = (pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Y)) &&
+                  pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Cb)) &&
                   pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_Cr)));
   }
   else if (colorspace == heif_colorspace_RGB) {
     if (chroma == heif_chroma_444) {
-      valid_pixi = (pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_R)) ||
-                    pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_G)) ||
+      valid_pixi = (pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_R)) &&
+                    pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_G)) &&
                     pixi->add_channel_bits(image->get_bits_per_pixel(heif_channel_B)));
     }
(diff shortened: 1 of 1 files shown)


bradh:
The SVT fuzzing result looks unrelated to the change: https://github.com/strukturag/libheif/actions/runs/27991514292/job/82844759436?pr=1848#step:5:3266

Have filed a work item upstream at: https://gitlab.com/AOMediaCodec/SVT-AV1/-/work_items/2373

farindk:
Thank you.

PR #1851, opened 2026-06-24
OSS-Fuzz: Add new fuzzer targets sequence processing

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets sequence processing logic, together with initial corpus for the fuzzer


fuzzing/sequence_fuzzer.cc
```diff
@@ -0,0 +1,72 @@
+/*
+ * HEIF codec.
+ * Copyright (c) 2026 struktur AG, Joachim Bauch <bauch@struktur.de>
```
farindk: You can put your email here.

arthurscchan: Thanks. Have changed now.

PR #1852, opened 2026-06-26
Add 10-bit vvenc encoding

torusrxxx:
vvenc plugin now supports 10-bit color depth. The encoded 10-bit VVC image can be decoded by FFmpeg decoder plugin.


farindk:
Thanks. Merged into v1.24.x. Also note my fix-up patch 53bf258b4f9b62468023c182ab81f5619a398c19.

PR #1857, opened 2026-07-01
Fix: parse 'meta' box with size 0 (extends to end of file)

d-polikhranidi:
## Problem

`FileLayout::read()` rejected any file whose top‑level `meta` box uses a box **size of 0** — which per ISO/IEC 14496‑12 clause 4.2 means "the box extends to the end of the file" (legal for the last box):

```cpp
if (box_header.get_box_size() == 0) {
    // TODO: get file-size from stream and compute box size
    return { ... "Cannot read meta box with unspecified size" };
}
```

Such files are read and rendered by libavif (the AVIF decoder used by Chromium), so libheif‑based pipelines (ImageMagick, libvips, sharp, …) fail on input that browsers display. Reproduces on `master` (v1.23.1).

## Fix

Implements the existing `// TODO`: when the `meta` box size is `0`, resolve it to the file size (via `StreamReader::request_range`, exactly as the surrounding code already does for known‑size boxes) and parse normally. This is the same "extends to end of file" handling libheif already applies generically (`Box::size_until_end_of_file` / `range.skip_to_end_of_file()`), and it is bounded by the file size, so there is no unbounded‑read concern.

`libheif/file_layout.cc` — the `meta` branch now:
- `size == 0` → `end_of_meta_box = request_range(meta_box_start, UINT64_MAX)` (file size);
- otherwise → the previous `size + start` path (unchanged), keeping the overflow / "invalid size" / "full box" checks.

(The identical `mini` / `moov` TODOs are left untouched to keep this change focused.)

## Test

Adds `tests/file_layout.cc → "meta box with size 0 (extends to end of file)"` and the 498‑byte fixture `tests/data/meta_size_zero.avif` (a file from libavif's own test corpus in which `meta` has size 0 and appears after the media data). This also covers the pre‑existing `// TODO: read file where 'meta' box is not the first one after 'ftyp'`.

## Verification

- Before: `heif-info meta_size_zero.avif` → *"Cannot read meta box with unspecified size"*.
- After: `heif-info` reads it as `33x11 (id=1), primary`; `avif_diff` / libavif render it identically.
- `tests/file_layout` → **All tests passed (3 assertions in 2 test cases)**.
- No regression on normal AVIF/HEIF files.


bradh:
I think we've seen this on `mdat`, but not on `meta`. Can you say what creates such a file?

farindk:
Please attach a test file, for example your `meta_size_zero.avif`.
I never came across such a file.

bradh:
That file is in the PR, and it does render in Chrome. Does not work in MP4Box.js though - that zero size `meta` isn't supported there.

Its a weird case though, with no `mdat`, and therefore you have to know the `idat` size, and since the rest of the header is predictable, there is no reason why it couldn't write the `meta` box size.

farindk:
Ah, sorry, I didn't see the file. Yes, it is quite strange. Still, it renders in Chrome, but not in Firefox.
Looks like an AI generated PoC, while scanning the code for TODOs to find security issues :-)

Since it's valid and not complex, we should support it.

d-polikhranidi:
> Ah, sorry, I didn't see the file. Yes, it is quite strange. Still, it renders in Chrome, but not in Firefox. Looks like an AI generated PoC, while scanning the code for TODOs to find security issues :-)
> 
> Since it's valid and not complex, we should support it.

You're right, but not quite :-)

I encountered this error when using `imgproxy`, which utilizes `libvips` with `libheif` under the hood.

I started looking through the error log and found this.
AI helped to arrange all this beautifully in PR in the style of the existing code.

If you approve, I can go through the rest of the not complex TODOs and implement them.

bradh:
> I encountered this error when using `imgproxy`, which utilizes `libvips` with `libheif` under the hood.

What sensor or software created the (original) file that error'd?

farindk:
That you. I have fixed an integer overflow resulting from passing UINT64_MAX and I applied the same fix to the `mini` and `moov` top-level boxes.

PR #1859, opened 2026-07-08
Validate bit depth in uvg266 encoder

torusrxxx:
uvg266 seems to only support 8-bit, and the existing bit depth detection is flawed because the upstream ignores the bit depth passed in (https://github.com/ultravideo/uvg266/blob/fb17bbc1ca64ca059293f45627886b1faa2faff7/src/uvg266.c#L423). Adds direct validation to only accept 8-bit images.


farindk:
Thanks, merged. Also added some changes to more accurately check bpp of the input images (51360e193a1288f54592aa6fc0b7037160636514, b3940a4a5539e7c9e15f34de3c51e75d7fabd88f).

PR #1860, opened 2026-07-08
Set premultiplied alpha and pixel aspect ratio in ffmpeg decoder

torusrxxx:


farindk:
As far as I know, the alpha mode field is only set by ffmpeg for the JPEG-XL codec, not for anything we use. If we set it like proposed in the PR, there could even be a mismatch between the value coming from ffmpeg and the property in the HEIF layer.

For aspect ratio, this may be set for H.265, but not for AV1. And we will also have potential conflicts with the HEIF layer.

I would prefer not to read these fields from ffmpeg.

farindk:
`alpha_mode` also needs FFMPEG 8.0, which would break some builds.

farindk:
For aspect ratio, the code would have to check at least for the undefined case (num > 0 && den > 0).

farindk:
BTW: If we read the SAR from H.265, we should also write it in encoders.

See: a26561d9

torusrxxx:
I think it's always necessary to encode and decode metadata in the bitstream by codec plugins whenever it is possible, let the HEIF core decide what to do with it. I'm closing this one, it seems this isn't necessary anymore.

PR #1867, opened 2026-07-16
svt encoder: use fractional QP via "cqp" option with SVT-AV1 >= 4.2.0

pszemus:
When libheif is built against SVT-AV1 >= 4.2.0, the encoder QP is passed through `svt_av1_enc_parse_parameter(&config, "cqp", ...)` instead of assigning `svt_config.qp` directly. The `"cqp"` option (introduced in SVT-AV1 4.2.0):

- accepts fractional QP values with quarter-QP precision
- sets `rate_control_mode = 0`

The QP derived from the `quality` parameter is no longer rounded to an integer:
e.g. `quality=50` now maps to CQP 31.50 instead of 32, so the 0–100 quality scale gets finer granularity. For SVT-AV1 < 4.2.0 the behavior is unchanged — the fractional value is rounded to the same integer the previous formula produced.

### Testing

Built against SVT-AV1 v4.2.0:

- `heif-enc -A -q 50` → SVT log shows `BRC mode / CQP Assignment: CQP / 31.50`
- `heif-enc -A -p qp=40` → `CQP / 40.00`,
- out-of-range QP values are rejected by SVT-AV1 at encoder initialization.

### Possible follow-ups (separate PRs)

Things worth considering next, intentionally left out of this PR:

- Use `svt_av1_enc_parse_parameter()` for setting the other encoder parameters as well — libheif would gain SVT-AV1's own validation of the values it sets.
- Allow setting a fractional CQP directly from the CLI / encoder parameter API. This PR only enables the fractional `quality` → QP mapping; the `qp` parameter still accepts integers only.

### DIFF
--- a/libheif/plugins/encoder_svt.cc
+++ b/libheif/plugins/encoder_svt.cc
@@ -28,6 +28,8 @@
 #include <deque>
 #include <memory>
 #include <limits>
+#include <cmath>
+#include <cstdio>
 
 #include "svt-av1/EbSvtAv1.h"
 #include "svt-av1/EbSvtAv1Enc.h"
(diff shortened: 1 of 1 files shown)


pszemus:
Hello @farindk
What do you think about this change?

farindk:
AV1 uses an 8-bit quantizer index. Moving from a 0-63 scale to fractional gives 4x the resolution. Not sure whether that is noticeable, but it doesn't hurt either.

Setting this value through a parameter string is unnecessary (and dangerous because of potentially different locales). You can also set it through `extended_crf_qindex_offset` (0-3, fractional part). This should even work with SVT-AV1 4.0.0.

Also make sure that you do not switch the rate-control mode from CRF to CQP as your original PR does.

farindk:
I've merged it and implemented the proposed changes on top of it.
4d638f15c64af222a6ade1c2521f13cc8c40d96f + 32e3afcdffe446234d4fa759d172812dda89727c

Thanks for the PR.

PR #1874, opened 2026-07-20
Update svt.cmd for 4.2.0

bradh:


bradh:
@farindk It looks like this PR would fix the current fuzzer problem with SVT and is now flagging on a different problem - where `refill` is now spinning. Do you read it the same way? Any idea of why it would timeout?

farindk:
>  where `refill` is now spinning. Do you read it the same way? Any idea of why it would timeout?

I think I fixed that in 49f78a57ab2b55c8c2f8b58613c4092766afd380. The problem was that for uncompressed codec input with hilariously large numbers for the row alignment, there was a loop in the reader that skipped over the alignment byte by byte, taking forever. The new skip implementation is constant time.

farindk:
Thanks, merged. I've also added the new `vmaf` tune option of v4.2.0. But since that cannot be applied to still images, libheif errors unless it is used for video sequences.

PR #1875, opened 2026-07-21
gdk-pixbuf: set GError on stop_load() failure

### LINKED ISSUE #1861: gdk-pixbuf loader: stop_load() fails without setting GError, crashing callers (GNOME Shell segfault at login)
### Summary

The gdk-pixbuf loader (`gdk-pixbuf/pixbufloader-heif.c`) never sets the `GError**` out-parameter when `stop_load()` fails. Every failure path only calls `g_warning()` and returns `FALSE`:

https://github.com/strukturag/libheif/blob/d035da94ba7835b597550337f4aedb51f3b0ce82/gdk-pixbuf/pixbufloader-heif.c#L66-L115

This violates the gdk-pixbuf module contract ("If an error occurs, error should be set and FALSE returned"). gdk-pixbuf detects it and prints:

```
Bug! gdk-pixbuf loader 'heif/avif' didn't set an error on failure
```

but callers that propagate the load through `GTask` end up with a NULL result *and* a NULL error, which can crash them.

### Real-world impact: GNOME Shell segfault loop, login impossible

On Ubuntu 24.04 (libheif 1.17.6, GNOME 46), a user had a HEIF (HEVC) photo as desktop wallpaper. After `libheif-plugin-libde265` was removed by `apt autoremove`, decoding started failing with "Unsupported feature: Unsupported codec". Because the loader sets no `GError`, mutter's background loading code (`meta_background_image` → `gdk_pixbuf_new_from_stream_at_scale_async` → `GTask`) hit:

```
Unsupported feature: Unsupported codec
Bug! gdk-pixbuf loader 'heif/avif' didn't set an error on failure
g_task_return_error: assertion 'error != NULL' failed
g_task_propagate_pointer: assertion 'task->result_set' failed
GNOME Shell crashed with signal 11
```

i.e. a NULL-pointer dereference in mutter's `file_loaded()` (`meta-background-image.c:176`, backtrace obtained with debug symbols). The shell crashed on every session start, so **the user could not log in at all** until the wallpaper setting was reset from a console. mutter should arguably be more defensive too, but the root cause is the missing `GError` here.

### Steps to reproduce

1. Have a `.heif` file whose codec cannot be decoded (e.g. HEVC content with the libde265 plugin not installed).
2. `GDK_PIXBUF_MODULE_FILE=... gdk-pixbuf-thumbnailer file.heif out.png` or any `gdk_pixbuf_new_from_*` call on it.
3. Observe `FALSE`/NULL returned with the `GError` left unset, and gdk-pixbuf printing the "didn't set an error on failure" warning.

### Suggested fix

In `stop_load()` (and any other failure path with access to the `GError**`), replace the bare `g_warning(...); goto cleanup;` with something like:

```c
g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_FAILED,
            "HEIF decoding failed: %s", err.message);
```

(`GDK_PIXBUF_ERROR_UNSUPPORTED_OPERATION` may fit the missing-codec case better.)

### Environment

- libheif 1.17.6 (Ubuntu 24.04 package `heif-gdk-pixbuf` 1.17.6-1ubuntu4.4); failure paths unchanged on current `master` (d035da9)
- gdk-pixbuf 2.42.10
- Observed via mutter 46.2 / GNOME Shell 46

Socialpranker:
*Made with AI assistance (Claude). I built this branch and verified the fix myself before opening this -- see Verification below.*

Fixes #1861.

## Problem

`stop_load()` in the gdk-pixbuf loader only calls `g_warning()` on every failure path and returns `FALSE` without setting the `GError**` out-parameter, which violates the gdk-pixbuf module contract ("If an error occurs, error should be set and FALSE returned"). Callers that drive the load through `GTask` (eg. mutter's `meta_background_image` -> `gdk_pixbuf_new_from_stream_at_scale_async`) hit `g_task_return_error`/`g_task_propagate_pointer` assertion failures on a NULL error, which crashes the caller.

The issue includes a real-world report: a HEIF wallpaper with an unsupported codec crashed GNOME Shell on every session start after `libheif-plugin-libde265` was removed, since the missing `GError` led to a NULL-pointer dereference in mutter's `file_loaded()` -- the user could not log in until the wallpaper was reset from a console.

## Fix

Replaced the `g_warning(...); goto cleanup;` on each of the 5 failure paths in `stop_load()` with `g_set_error()`, picking an error code that fits each case (`GDK_PIXBUF_ERROR_INSUFFICIENT_MEMORY` for the context allocation failure, `GDK_PIXBUF_ERROR_UNSUPPORTED_OPERATION` for `heif_error_Unsupported_feature` -- the exact case from the crash report -- `GDK_PIXBUF_ERROR_CORRUPT_IMAGE`/`GDK_PIXBUF_ERROR_FAILED` for the rest).

## Verification

Built just the `pixbufloader-heif` CMake target (gdk-pixbuf loader only, no codec backends needed since this doesn't touch decode logic) and wrote a small C test that `dlopen()`s the built `.so`, calls `fill_vtable()` to get the module's function pointers, and drives `begin_load`/`load_increment`/`stop_load` directly with garbage bytes (not a valid HEIF/HEIC container, so `heif_context_read_from_memory_without_copy` fails).

- **Before the fix** (reverted locally to confirm): `stop_load` returns `FALSE`, `g_warning` prints to stderr, and `GError` stays `NULL` -- exactly the bug described in the issue.
- **After the fix**: `stop_load` still returns `FALSE`, but `GError` is correctly set (`domain=gdk-pixbuf-error-quark`, message `"HEIF decoding failed: Invalid input: No 'ftyp' box: ..."`).

I did not have a way to reproduce the full GNOME Shell crash (no GNOME Shell/mutter environment available), so I verified at the level the fix actually operates: that the module contract is honored (`GError` set whenever `FALSE` is returned).

### DIFF
--- a/gdk-pixbuf/pixbufloader-heif.c
+++ b/gdk-pixbuf/pixbufloader-heif.c
@@ -81,25 +81,29 @@ static gboolean stop_load(gpointer context, GError** error)
 
   err = heif_init(NULL);
   if (err.code != heif_error_Ok) {
-    g_warning("%s", err.message);
+    g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_FAILED,
+                "HEIF decoding failed: %s", err.message);
     goto cleanup;
   }
 
   hc = heif_context_alloc();
   if (!hc) {
-    g_warning("cannot allocate heif_context");
+    g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_INSUFFICIENT_MEMORY,
+                "cannot allocate heif_context");
     goto cleanup;
   }
 
   err = heif_context_read_from_memory_without_copy(hc, hpc->data->data, hpc->data->len, NULL);
   if (err.code != heif_error_Ok) {
-    g_warning("%s", err.message);
+    g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_CORRUPT_IMAGE,
+                "HEIF decoding failed: %s", err.message);
     goto cleanup;
   }
 
   err = heif_context_get_primary_image_handle(hc, &hdl);
   if (err.code != heif_error_Ok) {
-    g_warning("%s", err.message);
+    g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_CORRUPT_IMAGE,
+                "HEIF decoding failed: %s", err.message);
     goto cleanup;
   }
 
@@ -109,7 +113,10 @@ static gboolean stop_load(gpointer context, GError** error)
                           has_alpha ? heif_chroma_interleaved_RGBA : heif_chroma_interleaved_RGB,
                           NULL);
   if (err.code != heif_error_Ok) {
-    g_warning("%s", err.message);
+    g_set_error(error, GDK_PIXBUF_ERROR,
+                err.code == heif_error_Unsupported_feature ? GDK_PIXBUF_ERROR_UNSUPPORTED_OPERATION
+                                                            : GDK_PIXBUF_ERROR_CORRUPT_IMAGE,
+                "HEIF decoding failed: %s", err.message);
     goto cleanup;
   }
 


Socialpranker:
Friendly ping — open since 21 July, CI green and no conflicts.

It makes `stop_load()` in the gdk-pixbuf loader set the `GError**` out-parameter on its failure paths instead of only calling `g_warning()`, which is what the gdk-pixbuf module contract requires. Callers driving the load through `GTask` (mutter's `meta_background_image`, in the reproducer from #1861) currently hit a `g_task_return_error` assertion on the NULL error and crash.

One file, and it's confined to the loader's error paths, so it should be a quick read whenever someone has a moment.

Socialpranker:
Bumping this one too — coming up on two weeks, still green on my end.

farindk:
Thank you. Looks good, merged.

PR #1882, opened 2026-08-10
Do not propagate compile macros in public interface

kmilos:
IMHO these should not appear in the `heif` target's public interface, just like `HAVE_ZLIB` or `HAVE_WEBCODECS` don't:
```
set_target_properties(heif PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "LIBHEIF_EXPORTS;HAVE_VISIBILITY;HAVE_LIBSHARPYUV=1;HAVE_BROTLI=1"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
```
If it's a must for some reason, I suppose they shouldn't have generic names like `HAVE_FOO` with a likelihood they might be recycled elsewhere...

P.S. `HAVE_VISIBILITY` naming might need to be revisited as well.


examples/CMakeLists.txt
```diff
@@ -70,6 +70,10 @@ target_include_directories(heif-enc PRIVATE ${libheif_SOURCE_DIR})
 install(TARGETS heif-enc RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 install(FILES heif-enc.1 DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
 
+if (LIBSHARPYUV_FOUND)
+    target_compile_definitions(heif-enc PRIVATE HAVE_LIBSHARPYUV=1)
+endif()
+
```
kmilos, line 76: There's also some doubt whether `heif-enc` code should [actually use this a prori](https://github.com/strukturag/libheif/blob/1a3583bcce77de6d3f8701c0758e3954863681ba/examples/heif_enc.cc#L1667-L1672), or just receive an error from libheif and handle/pass it instead ...

kmilos:
The alternative is, of course, to actually make these a part of the interface in e.g. generated `heif_version.h` or some other similar mechanism, with more appropriate names...

examples/CMakeLists.txt
farindk: Right, that would be cleaner. Probably, we should have a `heif_have_chroma_downsampling_algorithm()` so that we can show the error before starting the expensive decoding.

farindk: But we have to do that in a `v1.24.x` branch

farindk:
Thanks. There will be a follow-up for the `heif_have_chroma_downsampling_algorithm()` in a `v1.24.x` branch.

farindk:
`heif_have_chroma_downsampling_algorithm()` added in e4f4d67f.

PR #1883, opened 2026-08-11
Re-enable Coverity scan job

moshekaplan:
Re-enable Coverity job since Coverity Scan is now available again.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -9,10 +9,6 @@ permissions:
 
 jobs:
   scan:
-    # TEMPORARILY DISABLED: scan.coverity.com is down and the download step
-    # fails before the build can run. Re-enable by removing this `if` line
-    # once the Coverity service is back.
-    if: false
     runs-on: ubuntu-22.04
     env:
       TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}


.github/workflows/coverity.yml
```diff
@@ -9,10 +9,6 @@ permissions:
 
 jobs:
   scan:
-    # TEMPORARILY DISABLED: scan.coverity.com is down and the download step
-    # fails before the build can run. Re-enable by removing this `if` line
-    # once the Coverity service is back.
-    if: false
     runs-on: ubuntu-22.04
```
kmilos, line 12: Might as well bump the runner at the same time, 22.04 will be deprecated very soon...

moshekaplan: @kmilos : It seems like that's something that needs to occur other CI jobs too (e.g., `build.yml`, `clang.yml`), and so might be worth doing across the board in a separate PR?

**Update**: I tested upgrading the runner OS and it failed due to the PPA's do not include APT repositories for Ubuntu 26.04 (failed jobs visible here: https://github.com/moshekaplan/libheif/pull/1 )

kmilos: True, can do it that way of course.

> it failed due to the PPA's do not include APT repositories for Ubuntu 26.04

I meant the next supported version, 24.04, not skipping two 😉; [26.04 is still in "preview"](https://github.com/actions/runner-images#available-images)...

moshekaplan: _whoops_. But the same issue manifests for building with 24.04 - some of the CI jobs are still failing with the same error (also visible at https://github.com/moshekaplan/libheif/pull/1).

kmilos: So either remove that PPA from CI as libde265 is available in universe, or, if newer versions are really preferred, raise a new issue [here](https://github.com/strukturag/libde265/issues) to provide the PPA for 24.04 (and 26.04?).

moshekaplan: Honestly, I'm not familiar enough with the ecosystem to determine what the ideal solution is for upgrading the runners. I was trying to solve a simpler problem - getting the Coverity CI job running again.

farindk:
Thanks. Coverity is running again and all workflows are now running on Ubuntu 24.04.

PR #1885, opened 2026-08-17
Fix item IDs returned by generic item writers

gregbenz:
## Problem

`heif_context_add_item()`, `heif_context_add_precompressed_mime_item()`, and
`heif_context_add_uri_item()` assign a successful `Result<heif_item_id>` directly to
`out_item_id`. This uses `Result`'s boolean conversion, so successful calls report item ID `1`
instead of the ID that was allocated. References built from that output can therefore target the
wrong item.

## Fix

Dereference each successful result before assigning it to `out_item_id`, matching the
already-correct `heif_context_add_mime_item()` implementation.

## Test

Add a regression that first allocates item ID 1, then creates items through all three affected
APIs and verifies that each returns a distinct ID with the expected item type.

## Verification

- Before: the new test fails because `heif_context_add_item()` returns the existing ID 1.
- After: `item_writing` passes.
- All 15 tests in the configured suite pass (10 existing fixture/configuration skips).
- `cpplint` reports no errors for the changed C++ files.


farindk:
Thanks.
There is another location with the same issue:
https://github.com/strukturag/libheif/blob/c5e94d628de785fe4489d407b5395346ee3fc090/libheif/api/libheif/heif_experimental.cc#L289

We should be able to detect this class of bug by making the `operator bool` explicit.

gregbenz:
Thanks, @farindk — good catch, and thanks for merging this. I’ll open a small follow-up PR for that improvement.

farindk:
No need for a follow-up PR. I've included it in 748656fb already.

PR #1890, opened 2026-08-25
Update to libaom v3.15.0

wantehchang:
Set the new cmake option ENABLE_APPS to 0.


wantehchang:
@farindk Dirk: Please review. Thanks!

farindk:
Thank you, I have merged it into master.

BTW: I have disabled tune=IQ for lossless mode since they do not seem to work together (unrelated to v3.15.0).

wantehchang:
Dirk: Please follow [libaom issue 383595066](https://aomedia.g-issues.chromium.org/issues/383595066) for the inability to enable lossless mode with tune=iq. You are right. That problem was not introduced in libaom v3.15.0.

PR #1893, opened 2026-08-27
Add item visibility and specialized entity group writers

gregbenz:
## Problem

libheif can read item visibility and entity groups, but its public API cannot write them. This prevents callers from constructing valid derived-image layouts that need hidden items or ordered groups without private patches.

## Change

This adds:

- an API to set an existing item's hidden flag
- a specialized writer for ordered alternative (`altr`) groups
- a specialized writer for left/right stereo (`ster`) groups
- a specialized writer for stereo pairs with a monoscopic fallback (`stem`)

The specialized group APIs enforce each group type's requirements, making them easier to use without creating non-compliant files.

The `stem` writer requires distinct left and right images followed by a monoscopic fallback. The fallback may be a separate image or may reuse either stereo view. `stem` parsing is also registered so its ordered entities remain available through the public reader after serialization.

The visibility setter rejects hiding the primary image and keeps top-level image queries synchronized. Selecting a hidden image as primary makes it visible first.

Track entities are intentionally out of scope.

## Tests

Round-trip tests write and reopen files to verify:

- hidden flags and top-level image synchronization
- ordered alternative membership
- stereo-pair ordering
- all supported monoscopic fallback arrangements

Additional coverage checks invalid and non-image references, duplicate and overlapping alternatives, malformed `ster` and `stem` groups, security limits, region-mask handling, and hidden-primary promotion.

### DIFF
--- a/libheif/api/libheif/heif_encoding.cc
+++ b/libheif/api/libheif/heif_encoding.cc
@@ -790,6 +790,11 @@ heif_error heif_context_add_overlay_image(heif_context* ctx,
 heif_error heif_context_set_primary_image(heif_context* ctx,
                                           heif_image_handle* image_handle)
 {
+  Error visibility_error = ctx->context->set_item_hidden(image_handle->image->get_id(), false);
+  if (visibility_error) {
+    return visibility_error.error_struct(ctx->context.get());
+  }
+
   ctx->context->set_primary_image(image_handle->image);
 
   return heif_error_success;
--- a/libheif/api/libheif/heif_encoding.h
+++ b/libheif/api/libheif/heif_encoding.h
@@ -365,6 +365,7 @@ heif_error heif_context_add_overlay_image(heif_context* ctx,
                                           const uint16_t background_rgba[4],
                                           heif_image_handle** out_iovl_image_handle);
 
+// Select the primary image. If the image is hidden, it is made visible first.
 LIBHEIF_API
 heif_error heif_context_set_primary_image(heif_context*,
                                           heif_image_handle* image_handle);
--- a/libheif/api/libheif/heif_entity_groups.cc
+++ b/libheif/api/libheif/heif_entity_groups.cc
@@ -27,6 +27,78 @@
 #include <vector>
 
 
+namespace {
+
+heif_error add_entity_group(heif_context* ctx,
+                            const std::shared_ptr<Box_EntityToGroup>& group,
+                            const std::vector<heif_item_id>& item_ids,
+                            bool require_image_items,
+                            bool enforce_unique_alternative_membership,
+                            heif_entity_group_id* out_group_id)
+{
+  const auto* limits = ctx->context->get_security_limits();
+  if (limits->max_size_entity_group && item_ids.size() > limits->max_size_entity_group) {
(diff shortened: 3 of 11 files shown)


torusrxxx:
I prefer specialized entity group writers for each entity group type, the specialized writer makes it easier to use and to prevent non-compliant files.

gregbenz:
@torusrxxx Thanks, that makes sense. I’ve updated the PR to replace the generic group writer with specialized writers for alternative (`altr`) and stereo (`ster`) groups. Each writer now enforces the requirements for its group type.

torusrxxx:
I think this need to target the branch https://github.com/strukturag/libheif/tree/v1.24.x

Also, did you consider to also support stereoscopic pair with monoscopic fallback (`stem`) defined in ISO/IEC 23008-12:2025/Amd. 1:2025 Section 6.8.11?

gregbenz:
@torusrxxx Thanks, targeting v1.24.x makes sense for these new public APIs. Since that branch has diverged from master, it may be best to wait for @farindk’s confirmation and then rebase onto it rather than simply changing the PR base.

I had not included stem, since the immediate use case requires altr, but it does fit naturally alongside the specialized ster writer. I’d be happy to add a specialized stem writer that enforces the three-image ordering while allowing the monoscopic fallback to repeat the left or right image. Would you prefer that in this PR, or is a follow-up PR best?

farindk:
Yes, for new public APIs, please target branch `v1.24.x`. But I can also do that simply during merge.

gregbenz:
Thanks @farindk, that would be great. Please feel free to target v1.24.x during merge.

Do you have a preference on including the specialized stem writer in this PR, or handling that separately? I'm happy to update this PR if that's your preference.

farindk:
> Do you have a preference on including the specialized stem writer in this PR, or handling that separately? I'm happy to update this PR if that's your preference.

I don't care so much. Same PR, separate commits sounds logical to me.

gregbenz:
Thanks @farindk and @torusrxxx. I’ve added `stem` support to this PR as a separate commit, including the specialized writer, parser registration, validation, and round-trip tests. I’ve left the base branch unchanged so it can be retargeted to `v1.24.x` during merge as discussed.

torusrxxx:
Reading stereoscopic pairs is currently too cumbersome. It needs to enumerate all `ster` and `stem` entity groups, find which entity group contains the current image id, then read the paired image id. Leaving the application to do it requires too much redundant work for each application, and will likely not work for some configurations like stereoscopic pairs in an image burst. I think adding a helper function to read the stereoscopic pair that contains the given image id, to abstract away the underlying entity group, would simplify things a lot.

torusrxxx:
Perhaps this also needs to coordinate with #1121 as both uses `altr`. I would like to see both merged into v1.24.x then we can polish them.

gregbenz:
Thanks, sounds good. A helper that takes an image ID and finds the relevant stereo images without making each caller walk the `ster` and `stem` groups would be easier to use, especially for image bursts.

The sequencing you suggest sounds good: merge this PR and #1121 into v1.24.x, then coordinate the `altr` handling and add the stereo read helper as part of the follow-up.

Issue #1856, opened 2026-06-29
Cropped HEIC images fail with “Security limit exceeded” after decoder security limit changes


anatoliiKumunzhiev:
Hi,

I am upgrading libheif from 1.19.7 to 1.23.0 and libde from 1.0.15 to 1.1.1 and faced with a decoding regression for some HEIC files with transformation properties.

This looks related to this issue:

https://github.com/strukturag/libheif/issues/1798
https://github.com/strukturag/libheif/commit/69a0ca1bed228e5766c298db8ee86b5569bba71a

As I understand, for cropped HEIC images, it looks like the new limits may be calculated from the transformed size after clap, while the HEVC decoder still has to decode the full coded frame.

**1) C013.heic**

heif-dec with libheif 1.23.0:

_File contains 2 images
Could not decode image: Memory allocation error: Security limit exceeded:
Image size 1280x720 exceeds the maximum image size 132496_

heif-info:

_image: 300x300 (id=1002), primary
transformations:
  crop: left=490 top=210 right=490 bottom=210_

_image: 300x300 (id=1005)
transformations:
  crop: left=490 top=210 right=490 bottom=210_

From heif-info -d, both image items have:

_ispe = 1280x720
clap = 300x300_

The reported limit matches the crop size plus coding_unit_size:

_(300 + 64) * (300 + 64) = 132496_

But the decoder still needs to decode the full coded HEVC frame:

_1280 * 720 = 921600_

**2) MIAF007.heic**

heif-dec with libheif 1.23.0:

_File contains 1 image
Could not decode image: Memory allocation error: Security limit exceeded:
Image size 1280x720 exceeds the maximum image size 298496_

heif-info:

_image: 360x640 (id=1002), primary
transformations:
  crop: left=320 top=180 right=320 bottom=180
  angle (ccw): 90
  mirror: vertical_

From heif-info -d:

_ispe = 1280x720
clap = 640x360
irot = 90 degrees
imir = vertical_

The reported limit again matches the clap size plus coding_unit_size:

_(640 + 64) * (360 + 64) = 298496_

But the actual frame is still:

_1280 * 720 = 921600_

**So, the questions are:**
Do I understand correctly that the files look valid and this behavior is a regression?
Should the decoder limit be based on the uncropped ispe size instead of the transformed?

Samples:
[sl_heic_samples.zip](https://github.com/user-attachments/files/29470462/sl_heic_samples.zip)

riyasy:
[...]

PR #1895, opened 2026-08-28
fix the size passed to tighten_image_size_limit_for_ispe()

programingKing:
While upgrading libheif I hit "Security limit exceeded" on cropped HEIC files. I think this is the cause of #1856. 

`tighten_image_size_limit_for_ispe()` receives the post-transform size (`get_width()`) instead of the ispe size, so images with a large clap crop can no longer be decoded. Passing `get_ispe_width()/get_ispe_height()` fixes it for me.

### DIFF
--- a/libheif/image-items/image_item.cc
+++ b/libheif/image-items/image_item.cc
@@ -1291,7 +1291,7 @@ Result<std::shared_ptr<HeifPixelImage>> ImageItem::decode_compressed_image(const
   // when the codec bitstream lies about its dimensions.
   heif_security_limits tightened = tighten_image_size_limit_for_ispe(
       get_context()->get_security_limits(),
-      get_width(), get_height(),
+      get_ispe_width(), get_ispe_height(),
       max_coding_unit_size_for_codec(get_compression_format()));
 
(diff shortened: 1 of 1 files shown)


torusrxxx:
The real problem with #1856 is that the security limit defines the maximum size of any allocated framebuffer, not just the image size. If the security limits were applied to image size only, you can crop a gigapixel image down to 1x1 and bypass the security limits, but that is not desirable if the system is unable to allocate a gigapixel framebuffer just to extract a 1x1 region from it. You need to physically crop the image to fit in security limits, or raise the security limits.

farindk:
> The real problem with https://github.com/strukturag/libheif/issues/1856 is that the security limit defines the maximum size of any allocated framebuffer, not just the image size.

~~Right. Even worse, `ispe` could lie in a malicious file, bypassing the security limit and forcing an OOM crash.~~

programingKing:
Fair enough, makes sense as a policy. The naming (..._for_ispe, ispe_width) receiving the crop size is what made me read it as a bug. I'll close this PR. thanks for the context.

farindk:
Sorry, my above comment was nonsense. Your fix is exactly right. Merging it.

PR #1896, opened 2026-08-28
Add support for chroma=400 when encoding with aom

### LINKED ISSUE #1894: `heif-enc -e aom -p chroma=400` gives "Error: Invalid parameter value"
Hello, title is pretty explicit, since I want to make sure that my grayscale PNGs are converted into grayscale AVIF, I tried that (similar to `avifenc -l -y 400`, basically) but it gave me that error.

Other values like 420, 422 and 444 were recognized normally. Any suggestion?

This is using libheif 1.23.1 on Gentoo (so possibly some compile options missing).

q3cpma:
Fixes https://github.com/strukturag/libheif/issues/1894

Tested locally and it works fine.

### DIFF
--- a/libheif/plugins/encoder_aom.cc
+++ b/libheif/plugins/encoder_aom.cc
@@ -189,7 +189,7 @@ static const char* kParam_speed = "speed";
 
 static const char* kParam_chroma = "chroma";
 static const char* const kParam_chroma_valid_values[] = {
-    "420", "422", "444", nullptr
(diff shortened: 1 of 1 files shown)


farindk:
Thanks. Unfortunately, it does not work like this. For a color input, `aom_query_input_colorspace2()` will then return a `heif_colorspace_YCbCr` / ` heif_chroma_monochrome` combination, which is invalid.

PR #1902, opened 2026-09-08
POC: recover non-strict decoding when `ispe` and decoded frame size differ

### LINKED ISSUE #1820: decoded image size check in v1.22.0: was the opt-out removal intentional?
Hi,

Trying to figure out whether the new "Decoded image does not have the size signaled in the file." check in v1.22.0 (from PR #1709) is supposed to be strict like this, or whether the opt-out got lost in a later refactor.

Timeline as I read it:

- `c5ae55a6` (PR #1709, "check that decoded image size matches the claimed size") added the check in `context.cc`, gated on `!options.ignore_transformations`.
- `b1943651` (same PR, follow-up for non-visual unci) kept the gate.
- In issue #1709 you wrote: *"However, when the option `ignore_transformations=true` in the decoding options, this check is disabled, because it is expected that the image size will not match the reported metadata size anymore (because of rotations, crop)."* So the original design clearly intended an opt-out.
- `d13e8518` ("improved check of decoded image size", May 15) moved the check out of `context.cc` into per-item `ImageItem::check_decoded_image_size()`, called from `decode_image()` before transformations. The `ignore_transformations` gate was dropped in the move and not replaced.

Removing the gate makes sense for the new architecture: the check now compares decoded planes against pre-transform `ispe`, so `ignore_transformations` doesn't really apply anymore. But after the refactor there's no opt-out at all, which is what I'm trying to confirm.

Why I'm asking: pillow_heif has an `ALLOW_INCORRECT_HEADERS` option for users who want to decode files with a mismatched `ispe`. Historically libheif returned the actual decoded size and we carried on with that. Against v1.22.0 these decodes now fail at the new check, surfaced downstream as pillow_heif#424.

Was the loss of the opt-out intentional or accidental? If intentional we'll drop our `ALLOW_INCORRECT_HEADERS` option, otherwise we'll wait for the next libheif release.

homm:
Fixes [libheif #1820](https://github.com/strukturag/libheif/issues/1820).

This is a proof of concept for a HEVC file where the dimensions signaled by `ispe` do not match the frame returned by the decoder:

- signaled `ispe`: 464×464
- decoded frame: 462×462
- clean aperture output: 461×462

The file appears to contain both padding in the container-level dimensions and an HEVC conformance window that already removes part of that padding. The existing exact-size validation rejects it before transformations are applied.

Other decoders, including macOS ImageIO, decode this file successfully. The output produced by this change was compared with the macOS-decoded result and matches.

### What this POC does

The normal strict path is unchanged:

- `strict_decoding = true` still rejects the image with `heif_suberror_Invalid_image_size`.

For non-strict decoding, a size mismatch is accepted only when:

- the decoded frame does not exceed the `ispe` dimensions;
- all stored planes are internally consistent;
- the normal decoder security limit derived from `ispe` has already bounded allocations.

The original size error is retained as a decoding warning.

When transformations are enabled, the recovery requires a `clap` before any rotation or mirroring that would make the coordinate mapping ambiguous. The clean aperture is evaluated against the actual decoded frame. If its signaled position extends slightly outside that frame, the aperture is shifted into the available frame while preserving its size. If the aperture cannot fit, decoding still fails with the original size error.

For the sample file, this changes the effective crop on the 462×462 decoded frame to `left=0`, `top=0`, `right=1`, `bottom=0`. This produces the expected 461×462 output without creating synthetic padding.

With `ignore_transformations = true`, libheif now returns the real decoded 462×462 frame and attaches the same size warning. This is useful for callers that explicitly request the decoder output rather than the transformed display image.

### Transformation API consistency

`heif_item_get_property_transform_crop_borders()` now uses the same clean-aperture adjustment as the decoder. A caller that supplies the actual decoded frame size therefore receives borders that describe the crop applied by the decoder.

The public function signature and ABI are unchanged, but this is a behavioral change: an aperture that would otherwise lie partly outside the supplied image is shifted into that image when it fits.

### Tests

A new HEVC regression fixture and test cover:

- strict decoding still failing;
- non-strict decoding with transformations producing 461×462;
- non-strict `ignore_transformations` producing 462×462;
- the warning in both non-strict cases;
- adjusted crop borders returned through the public transformation API.

The full configured test suite passes: 68/68 tests.

### Open questions / why this is a POC

This patch is intentionally offered as a starting point, not necessarily as the final design.

In particular, maintainers may prefer to:

- limit aperture adjustment to the non-strict recovery path instead of changing the general crop-border API behavior;
- define a broader policy for mismatches combined with `irot`, `imir`, grids, tiles, alpha images, or other transformative properties;
- keep raw file properties separate from a normalized transformation view.

The important property this POC tries to preserve is that non-strict recovery never invents visible pixels: it only accepts the file when the requested clean-aperture output is fully contained in the decoded frame.


homm:
After tracing the public API lifecycle more carefully, I realized that my current POC fixes the size mismatch too late: the recovery happens in `ImageItem::decode_image()`, after the public `heif_image_handle` has already been created and clients may already have read its properties.

For this file, the inconsistency is masked when transformations are enabled because the signalled `clap` happens to produce the same final 461×462 display size. However, this is not a coherent API result: `ignore_transformations=true` decodes to 462×462 while the handle metadata was created from the signalled dimensions, and callers deriving crop borders from `get_ispe_*()` still use the uncorrected 464×464 coordinate space. Component descriptions have the same issue.

Therefore I do not think the POC should be merged in its current form. A proper solution needs to determine and store the effective pre-transform coded size before exposing an image handle, then derive the public dimensions and transformations from that value.

homm:
I think this file has a more specific problem than a generic “crop was applied in the wrong coordinate system” case.

The HEVC SPS declares a coded canvas of 464×464 (`pic_width_in_luma_samples` / `pic_height_in_luma_samples`). Its conformance window removes 0 pixels at the left/top and 2 pixels at the right/bottom (the SPS offsets are one chroma unit each, i.e. two luma pixels for 4:2:0), so the decoder-visible frame is 462×462. The container `ispe`, however, is also 464×464: it describes the coded HEVC canvas rather than the image produced by the codec after its mandatory conformance-window crop.

The `clap` is internally consistent with that coded 464×464 coordinate space. It specifies an aperture of 461×462 at `(0, 0)`: `x = (464 - 461) / 2 - 3 / 2 = 0`, `y = (464 - 462) / 2 - 1 = 0`. Mapping that aperture through the SPS conformance window gives the same rectangle `[0, 461) × [0, 462)` in the actual 462×462 decoder frame. Thus the intended results are unambiguous: 462×462 with transformations ignored, and 461×462 with `clap` applied.

So this does not need a heuristic such as shifting a crop until it fits. A narrow non-strict recovery path could recognize the following verifiable condition: `ispe` exactly equals the SPS coded size, the SPS conformance window exactly explains the decoder output size, and every mapped HEIF transformation is valid in that output frame. It could then use the SPS visible frame as the pre-transform coordinate space. Otherwise the existing validation failure should remain.

The file is still malformed: the container signals 464×464 where it should signal the HEVC visible size 462×462. But it is a mathematically consistent and recoverable form of malformed input, not an ambiguous alternate crop interpretation.

One important implementation consequence is that my POC should not be merged as-is. It applies the correction in `ImageItem::decode_image()`, after `heif_image_handle` and its public properties have already been created. A proper implementation needs to establish the effective pre-transform size and mapped transformations before those public properties are exposed; `ispe` itself should remain the literal container property.

PR #1903, opened 2026-09-09
Add heifio just for examples and testing

torusrxxx:
Don't add `heifio` in `CMakeLists.txt` for `libheif` consumers.

----

Original review https://github.com/microsoft/vcpkg/pull/53645#discussion_r3966981286


farindk:
Thanks, but `BUILD_TESTING` is not defined yet where the new check runs.
Move the `BUILD_TESTING` option up, for example where the `WITH_EXAMPLES` are.

PR #1907, opened 2026-09-15
Fix HEVC test

scaronni:
The x265 plugin calls `api->encoder_headers()` as soon as a sequence encoder is opened, and `encode_sequence_frame()` calls `get_data()` immediately afterwards; before any frame is encoded.

That call sees three parameter-set NALs and nothing else, so the optional is still empty when line 260 dereferences it. That is turned into the `_M_is_engaged()` abort instead of silent UB because of Fedora's default GCC option `-Wp,-D_GLIBCXX_ASSERTIONS`.

Sample error without this patch:

```
42/48 Test #20: tai .........................................   Passed    0.19 sec
43/48 Test  #2: encode ......................................   Passed    0.23 sec
44/48 Test #15: sequence_mixed_bit_depth ....................Subprocess aborted**Exception:   0.23 sec
/usr/include/c++/16/optional:1237: constexpr _Tp std::optional<_Tp>::operator->() [with _Tp = Encoder::CodedImageData]: Assertion 'this->_M_is_engaged()' failed.
Randomness seeded to: 312269191

sequence_mixed_bit_depth is a Catch2 v3.7.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
a sequence frame must carry the bit depth of the first frame
  HEVC
-------------------------------------------------------------------------------
/builddir/build/BUILD/libheif-1.23.4-build/libheif-1.23.4/tests/sequence_mixed_bit_depth.cc:206
...............................................................................

/builddir/build/BUILD/libheif-1.23.4-build/libheif-1.23.4/tests/sequence_mixed_bit_depth.cc:206: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 60 | 59 passed | 1 failed


45/48 Test  #8: item_properties .............................   Passed    0.32 sec
46/48 Test #45: uncompressed_encode .........................   Passed    0.40 sec
47/48 Test #28: omaf ........................................   Passed    0.45 sec
48/48 Test  #3: encode_grid .................................   Passed    0.65 sec

98% tests passed, 1 tests failed out of 48

Total Test time (real) =   0.65 sec

The following tests FAILED:
     15 - sequence_mixed_bit_depth (Subprocess aborted)
Errors while running CTest
error: Bad exit status from /var/tmp/rpm-tmp.lqvMvD (%check)
```


rathann:
I can confirm this fixes the test in 1.23.4 for me (Fedora rawhide + RPM Fusion).

farindk:
Thanks, confirmed. I reproduced the abort on master with -D_GLIBCXX_ASSERTIONS and the patch fixes it; a two-frame heif-enc -S round trip with the patched build also looks right (hvcC in the sample entry, both frames decode).
VVC doesn't actually hit this path (vvenc doesn't pre-emit headers), but the guard is harmless there. Merging.

PR #1908, opened 2026-09-15
split-out the AVIF part of the "ispe odd size" test

rathann:
Skip respective part if AV1 or HEVC codec is unavailable.

This helps tests pass in Fedora builds, where we have to disable HEVC.


farindk:
Merged, thanks. I renamed the first case to "HEVC ispe odd size" and fixed the SKIP() indentation in a follow-up commit. 24df7143d42651bac838fdd58039e5dc897bd6b4

rathann:
> Merged, thanks. I renamed the first case to "HEVC ispe odd size" and fixed the SKIP() indentation in a follow-up commit.

You're welcome and thanks for the fix-up.

PR #1913, opened 2026-09-20
Jasper JPEG 2000 decoder

torusrxxx:
Adds support for Jasper JPEG 2000 and JPEG decoder.


farindk:
Thanks. Any advantages compared to OpenJPEG ?

As far as I see, JasPer is a more limited implementation, slower, and not as widely adopted.

torusrxxx:
The problem with OpenJPEG is it is [unmaintained](https://github.com/uclouvain/openjpeg#unmaintained-repository). It's better to support a maintained alternative.

bradh:
Jasper isn't really getting any more attention than OpenJPEG...

farindk:
I see that OpenJPEG says that it is "unmaintained", but still it seems to get more attention than Jasper.

I'm currently flooded with AI detected security reports which costs me more time than I actually have available. Thus, I'd like to keep the attack surface as small as possible. In theory, a plugin is easy to switch off if there is a problem, but as long as it does not give *any* advantage, it just increases the maintenance burden.

We can still keep this open for the future.

torusrxxx:
[JasPer](https://github.com/jasper-software/jasper) supports setting security limits explicitly and has fewer issues, while [OpenJPEG](https://github.com/uclouvain/openjpeg) seems to be overwhelmed by AI security reports. [I think having an alternative is beneficial just to avoid known unpatched security issues with OpenJPEG](https://github.com/jasper-software/jasper/issues/357#issuecomment-4529287474).

I know there's a lot of AI security reports. When you finished patching all of it, could you update `v1.24.x` so that I can apply patches to every one of my `v1.24.x` branches as well? Thanks.

PR #1914, opened 2026-09-21
Fix sample duration preservation and mapping in visual sequence tracks

twn39:
### Summary of Changes

This PR resolves three interrelated issues where frame durations (sample durations) in visual sequence tracks (such as Animated AVIF `avis` and Animated HEIC `msf1`) are lost or incorrectly mapped during encoding and decoding:

1. **Duration lost during sequence encoding (`Track_Visual::encode_image`)**:
   When `convert_colorspace_for_encoding` converts the input image, the resulting `colorConvertedImage` is a newly constructed `HeifPixelImage` whose `m_sample_duration` defaults to 0. The duration previously assigned via `heif_image_set_duration()` was lost, resulting in 0 duration being forwarded to `encoder->encode_sequence_frame()` and recorded in `m_frame_user_data`.
   - **Fix**: Explicitly forward `colorConvertedImage->set_sample_duration(image->get_sample_duration())`.

2. **Sample duration index misaligned during sequence decoding (`Track_Visual::decode_next_image_sample`)**:
   `Track_Visual::decode_next_image_sample` currently assigns duration using `m_stts->get_sample_duration(sample_idx_in_chunk)`. However, `sample_idx_in_chunk` is an inner loop variable tracking data pushed into the decoder. Once the decoder flushes or finishes pushing samples, this cursor wraps back to 0. Consequently, all subsequent decoded frames in variable frame-rate animations erroneously receive the duration of frame 0.
   In `track.cc:1219` (`Track::get_next_sample`), the codebase already maps the emitted sample duration via the presentation timeline:
   ```cpp
   const auto& sampleTiming = m_presentation_timeline[m_next_sample_to_be_output % m_presentation_timeline.size()];
   sample->duration = m_stts->get_sample_duration(sampleTiming.sampleIdx);
   ```
   - **Fix**: Align `Track_Visual::decode_next_image_sample` with `track.cc` by looking up duration via `m_presentation_timeline[m_next_sample_to_be_output % m_presentation_timeline.size()].sampleIdx`.

3. **Duration lost during decoded output colorspace conversion (`heif_track_decode_next_image`)**:
   When `heif_track_decode_next_image` converts decoded output to caller-requested RGB/RGBA, `convert_to_output_colorspace` produces a new `HeifPixelImage` without preserving `m_sample_duration`.
   - **Fix**: Preserve and reapply `sample_duration` to the converted image.

### Test Coverage

Added Catch2 regression test `tests/sequence_vfr_durations.cc` validating that:
- A sequence with variable frame rates ([100ms, 250ms, 400ms]) encodes properly into `stts`.
- Decoding in native colorspace returns `[100, 250, 400]`.
- Decoding with RGB colorspace conversion returns `[100, 250, 400]`.


farindk:
Thanks for the report and the test. I landed this as f17635b1 with your regression
test. I dropped the two hunks that re-apply the duration after colour conversion,
since the conversion pipeline already copies it in copy_metadata_from(), and
changed the description: the index runs ahead of the output by the decoder
latency, so every frame got the next sample's duration, not only the ones after
the flush.

PR #1915, opened 2026-09-21
kvazaar: fix NULL chroma pointer dereference when encoding monochrome/alpha images

twn39:
### Summary
Fixes a crash (NULL pointer dereference / SIGSEGV / Windows Access Violation `0xC0000005`) when encoding monochrome images (`heif_colorspace_monochrome`) or images with an Alpha channel using the Kvazaar HEVC encoder.

### Root Cause
In `kvazaar_start_sequence_encoding_intern()` (`libheif/plugins/encoder_kvazaar.cc`):
- When encoding a monochrome image (such as an auxiliary alpha image), the code set:
  ```cpp
  config->input_format = KVZ_FORMAT_P400;
  ```
  However, `kvz_config_init()` initializes `config->chroma_format` to `KVZ_CSP_420` by default.
- Kvazaar's internal state expects `config->chroma_format` to be updated (or `kvz_config_parse(config, "input-format", "P400")` to be invoked), which sets `chroma_format = KVZ_CSP_400` and `chroma_shift_w = 0, chroma_shift_h = 0`.
- Because `chroma_format` was never changed from `KVZ_CSP_420`, Kvazaar's encoder was created expecting 4:2:0 input.
- Later, in `kvazaar_encode_sequence_frame()`, the picture was allocated via `api->picture_alloc_csp(KVZ_CSP_400, ...)`. In Kvazaar's `videoframe.c`, `KVZ_CSP_400` does not allocate U and V chroma planes, setting `pic->u = NULL; pic->v = NULL;`.
- When `api->encoder_encode()` ran, Kvazaar's intra mode search (`search_intra.c: search_intra_chroma_rough`) checked `state->encoder_control->cfg.chroma_format != KVZ_CSP_400`. Because `cfg.chroma_format` was still `KVZ_CSP_420`, it attempted to access `pic->u` and `pic->v`, resulting in an immediate crash dereferencing NULL.

### Solution
1. In `kvazaar_start_sequence_encoding_intern()`, explicitly set `config->chroma_format` and corresponding chroma shift parameters.
2. Call `api->config_parse(config, "input-format", format_str)` to ensure Kvazaar's internal configuration state is fully synchronized.
3. Added a regression test `kvazaar encodes monochrome image` in `tests/encode.cc`.


farindk:
Thanks for the report and the analysis. The crash is real, but only against kvazaar git master since July 2026 (commit cbdcc31a moved chroma_format into kvz_config). The fields your patch writes do not exist in any released kvazaar, so the PR does not compile against 2.3.x. The portable fix is to configure the input format through config_parse("input-format", ...), which is what kvazaar's own CLI does and which works on every version. I've applied that variant together with your test in d3772c0f.

PR #1916, opened 2026-09-21
Fix signed 16-bit overlay offsets

0xk3r0:
## Summary

Fix incorrect sign conversion of 16-bit signed offsets in
`ImageItem_Overlay::readvec_signed()`.

The current implementation uses a fixed 32-bit mask (`0x7fffffff`)
when converting negative values. For 16-bit fields, valid negative
offsets such as `0xFFFF` are decoded incorrectly instead of as `-1`.

This can cause valid overlay content to be considered outside the
destination canvas and omitted during decoding.

## Changes

- Use a mask matching the actual signed field width.
- Add a regression test covering a negative 16-bit overlay offset.
- Add a 32-bit control case to preserve the existing behavior.

No memory-safety impact was observed during testing; the demonstrated
effect is incorrect image composition/decoding.

The detailed security report and PoC are intentionally not included
in this PR.

## Testing

The regression tests pass on the fixed build.


farindk:
Thanks for the report and the PoC. This was already fixed on master earlier
in f313930b, and your branch is based on top of that fix, so the patch
replaces the corrected code with an equivalent variant. The existing tests in
tests/overlay_offsets.cc cover negative offsets in both field widths, and
overlay_spec_parse.cc now checks the field decoding over the whole 16-bit range.
Closing as superseded.

